PR #1, opened 2015-05-17
Editor is not in the SDK

sethladd:
cc @ricowind


kevmoo:
LGTM. Landed as e57c13c

PR #23582, opened 2015-06-04
Tweak to markdown for link in service protocol spec

sethladd:


sethladd:
@turnidge is OOO, asking @johnmccutchan for a quick review

PR #23586, opened 2015-06-04
Add one-line doc for webGL

sethladd:


sethladd:
cc @alan-knight will this be overwritten, or is the preamble to this file stable?

sethladd:
Need to change a different file. See https://github.com/dart-lang/sdk/pull/23588#issuecomment-109086133

PR #23587, opened 2015-06-04
Add library level docs for SVG

sethladd:


sethladd:
Need to change a different file.

See https://github.com/dart-lang/sdk/pull/23588#issuecomment-109086133

PR #23588, opened 2015-06-04
Add one-line doc for web audio

sethladd:


vsmenon:
I think you'll need to add this to the template:

https://github.com/dart-lang/sdk/blob/master/tools/dom/templates/html/dartium/web_audio_dartium.darttemplate

and perhaps the dart2js version too (I think our online docs use dartium for now):

https://github.com/dart-lang/sdk/blob/master/tools/dom/templates/html/dart2js/web_gl_dart2js.darttemplate

sethladd:
Thanks! Happy to do that. I'll close these requests and open new ones.

Much appreciated.

PR #23589, opened 2015-06-04
Change link to reporting bugs for analyzer package

sethladd:


sethladd:
Thanks for taking a look!

bwilkerson:
LGTM. We should make a similar change for analysis_server's README.

sethladd:
where is the analysis_server's README? I don't see a repo for analysis_server (at least, by that name)

PR #23689, opened 2015-06-21
fix links in the sdk readme doc

devoncarew:
- fix the links for the `dart:core`, `dart:html`, and `dart:io` libraries
- various tweaks to the display of the sdk readme

@keertip @sethladd

### DIFF
--- a/sdk/api_readme.md
+++ b/sdk/api_readme.md
@@ -1,38 +1,31 @@
-Welcome to the Dart API reference documentation,
-covering the official Dart API libraries.
-
-Some of the most fundamental Dart libraries include:
+Welcome to the Dart API reference documentation, covering the official Dart API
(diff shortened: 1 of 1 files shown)


devoncarew:
My understanding is that this file is for use by dartdoc, when generating the sdk documentation.

The links are broken now when viewed in github (https://github.com/dart-lang/sdk/blob/master/sdk/api_readme.md); they're referring to index.html, which will only exist once gen'ed by dartdoc. They're also currently broken in the generated documentation, which is what this PR is addressing :)

I don't think there's a way to get them both happy. The directory structure is different - the names and the nesting level. We should either commit this PR as is or remove the links to the `dart:` libraries.

keertip:
lgtm!

PR #23730, opened 2015-06-26
Avoid multiline string literal in svgelement test.

ghost:
Due to the recently introduced crlf normalization, the svnelement_test no longer passes on windows. The issue is that the multiline string literal might contain crlf on windows if the local git client is configured to do the translation (which is the default). To avoid this, I have changed the test to use single line literals with explicit \n.


herhut-ggl:
I have created https://codereview.chromium.org/1219673002. Closing here.

PR #23867, opened 2015-07-19
make indentation configurable for codegen tools

guillermooo:
These changes should not affect the existing generators, but they may come in handy to generate API classes for Python and other languages if they prefer indentation sizes other than 2 spaces.


bwilkerson:
LGTM. I'll get it committed in the morning.

guillermooo:
ping

Let me know if any changes are needed.

PR #24024, opened 2015-08-08
add optional param to control whether to add newline

guillermooo:
@bwilkerson 

A new parameter enables client classes to decide whether they want
a newline char added at the end of a chunk of generated doc
comment text.

In Java, for example, this is not needed because the doc comment is
surrounded by `/** .. */`. In this case, the generator works fine as it is.
However, for Python I'm using `#`, so there's a gap between the
doc comment and the commented item, which looks odd.

The new parameter doesn't alter the default behavior; it must be
explicitly set to `true` if a client class doesn't want newlines at
the end of doc comments.


bwilkerson:
LGTM

PR #24159, opened 2015-08-20
Fix dart2js in the dev branch.

harryterkelsen:
I wasn't sure how to create a commit that only applied to the dev branch with 'git cl' so I just did it on my fork. Please merge this into dev or let me know the right way to create this commit so you can merge it into dev. Thanks!

### DIFF
--- a/pkg/compiler/lib/src/apiimpl.dart
+++ b/pkg/compiler/lib/src/apiimpl.dart
@@ -389,7 +389,8 @@ class Compiler extends leg.Compiler {
         packages =
             new MapPackages(pkgs.parse(packageConfigContents, packageConfig));
       }).catchError((error) {
-        reportError(NO_LOCATION_SPANNABLE, MessageKind.INVALID_PACKAGE_CONFIG,
+        reportError(leg.NO_LOCATION_SPANNABLE,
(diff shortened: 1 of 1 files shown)


sigmundch:
@hterkelsen I think one option is to do `git cl upload origin/dev` - that will tell rietveld to use `origin/dev` as the base CL.  `git cl land` might support that too, but I haven't tried that.

whesse:
Yes, you would make a branch with origin/dev, but you don't want to land it, because we do that with a merge commit to make the new version.  We would want a new branch, based on dev, in the main repo, with this commit on it.  Then I can cherry-pick it to dev in a new version.  But this is fine as is - I'll do it in the next merge.

hterkelsen:
My bad, I had not seen https://github.com/dart-lang/sdk/wiki/Cherry-picks-to-dev-channel
This could have been avoided by just including this patch in the original merge request. I'll follow those instructions in the future. Thanks Bill!

hterkelsen:
You can merge this instead: https://codereview.chromium.org/1308343003/

hterkelsen:
Closing this since the Rietveld patch was merged instead.

PR #24236, opened 2015-08-30
use const constructors of Utf8* at several places

a14n:


sethladd:
cc @lrhn @floitschG

lrhn:
LGTM

PR #24244, opened 2015-08-31
Bring in the latest stable version of pub.

nex3:
This fixes three issues with the `1.12.0` changes:
- Pub will now respect `.gitignore` when validating a package before it's
  published. For example, if a `LICENSE` file exists but is ignored, that is
  now an error.
- If the package is in a subdirectory of a Git repository and the entire
  subdirectory is ignored with `.gitignore`, pub will act as though nothing
  was ignored instead of uploading an empty package.
- The heuristics for determining when `pub get` needs to be run before various
  commands have been improved. There should no longer be false positives when
  non-dependency sections of the pubspec have been modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+## 1.12.1
+
+### Tool changes
+
+* Pub
+
+  * Pub will now respect `.gitignore` when validating a package before it's
+    published. For example, if a `LICENSE` file exists but is ignored, that is
+    now an error.
+
+  * If the package is in a subdirectory of a Git repository and the entire
+    subdirectory is ignored with `.gitignore`, pub will act as though nothing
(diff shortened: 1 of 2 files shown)


mit-mit:
nex3@, please clarify why this is needed now. 1.12 is only one day old. Can we not do this in 1.13?

mit-mit:
Discussed with Bill, this can piggy back if there is another, urgent request for a 1.12 MergeToStable.

nex3:
@mit-mit It's bug fixes for functionality in 1.12—isn't that what the stable channel is for?

munificent:
I have a matching patch to bring this into bleeding edge: https://codereview.chromium.org/1326953005/

whesse:
Merged in version 1.12.1.

PR #24414, opened 2015-09-23
Fix parameter type of MapMixin.containsValue

a14n:
Make this method compatible with [`Map.containsValue(Object value)`](https://github.com/dart-lang/sdk/blob/1.12.1/sdk/lib/core/map.dart#L147)


lrhn:
Thanks for the fix and finding the problem. 
The problem is actually a little more widespread, so I have landed https://codereview.chromium.org/1360053003/ to fix it "everywhere".

sethladd:
Thanks @a14n !

Issue #24623, opened 2015-10-19
Support rest params in package:js


ochafik:
A `@varargs` or `@rest` annotation would be useful in two use-cases:
- To implement ES6 JS helpers with rest params in DDC (see https://github.com/dart-lang/dev_compiler/issues/310):
  
  ``` dart
  mixin(base, @js.varargs mixins) {...}
  ```
  
  Would be compiled by DDC to:
  
  ``` javascript
  function mixin(base, ...mixins) {...}
  ```
- To call external (or internal) methods with rest params. The following call:
  
  ``` javascript
  foo(a, b, js.varargs(others)) // or js.rest(others)
  ```
  
  Would be compiled to ES6 code by DDC:
  
  ``` dart
  foo(a, b, ...others)
  ```
  
  And to something like the following ES5 code by dart2js (some devil in the `this` binding details):
  
  ``` javascript
  foo.apply(null, [a, b].concat(others))
  ```

mit-mit:
Clearing out 1.15 milestone as the last full push to dev has happened. If any changes are required before 1.15 is shipped, please file a merge request

jacob314:
Closing. We expect to instead leverage possible future core dart language features for rest parameters.

nex3:
@jacob314 What issue should we follow to get updates on those core Dart features and how they relate to rest parameters?

jacob314:
Here is a related language issue to follow. https://github.com/dart-lang/sdk/issues/16253

PR #24699, opened 2015-10-22
[package:js] Declarations for rest params / spread operator (issue #24623)

ochafik:


ochafik:
Thanks @jacob314, fixed the comments + separated as `library js.varargs`.

PR #24791, opened 2015-11-01
fix #24780 Error in dartium on literal JS object without named parameters

### LINKED ISSUE #24780: Error in dartium on literal object without named parameters
In dartium 1.13.0-dev.7.4 I get an error but it works in dart2js

```
import 'dart:html';
import 'package:js/js.dart';

@JS()
class B {
  external factory B();
}

main() {
  window.console.dir(new B());
}
```

Dartium does not seems to handle correctly an `external factory` without named parameters. The patch generated is :

```dart
patch class B {
  factory B({}) {
    var ret = new js_library.JsObject.jsify({});
    return ret;
  }
}
```

As you can see `factory B({})` is not a valid syntax.

a14n:


a14n:
cc @jacob314

jacob314:
Upstreamed a version of this containing a test.

PR #24904, opened 2015-11-12
Fix comment removal regex

diamondo25:
While this would also remove comments inside strings, this fix fixes generation of JS_CONST's to JavaScript.

Lines that created issues in the compiled result:

https://github.com/dart-lang/sdk/blob/7b48eb2b7fbcf6b7ee40372e31addec64d4c8d8f/sdk/lib/_internal/js_runtime/lib/native_helper.dart#L601

```
  C.JS_CONST_EyN = function(hooks) {  var getTag = hooks.getTag;  var prototypeForTag = hooks.prototypeForTag;  function getTagFixed(o) {    var tag = getTag(o);    if (tag == "Document") {      // "Document", so we check for the xmlVersion property, which is the empty      if (!!o.xmlVersion) return "!Document";      return "!HTMLDocument";    }    return tag;  }  function prototypeForTagFixed(tag) {    if (tag == "Document") return null;    return prototypeForTag(tag);  }  hooks.getTag = getTagFixed;  hooks.prototypeForTag = prototypeForTagFixed;};
```

https://github.com/dart-lang/sdk/blob/7b48eb2b7fbcf6b7ee40372e31addec64d4c8d8f/sdk/lib/_internal/js_runtime/lib/native_helper.dart#L515
https://github.com/dart-lang/sdk/blob/7b48eb2b7fbcf6b7ee40372e31addec64d4c8d8f/sdk/lib/_internal/js_runtime/lib/native_helper.dart#L519

```
  C.JS_CONST_jzj = function getTagFallback(o) {  var constructor = o.constructor;  if (typeof constructor == "function") {    var name = constructor.name;    if (typeof name == "string" &&        // constructor name does not 'stick'.  The shortest real DOM object        name.length > 2 &&        // On Firefox we often get "Object" as the constructor name, even for        name !== "Object" &&        name !== "Function.prototype") {      return name;    }  }  var s = Object.prototype.toString.call(o);  return s.substring(8, s.length - 1);};
```

Another workaround would be changing the comments to _not_ include quotes, but as the regex is only used (as far as I have seen) in blocks of code that do not include slashes inside strings, this fix should be fine.

### DIFF
--- a/pkg/compiler/lib/src/js_backend/constant_emitter.dart
+++ b/pkg/compiler/lib/src/js_backend/constant_emitter.dart
@@ -18,10 +18,10 @@ typedef jsAst.Expression _ConstantListGenerator(jsAst.Expression array);
 class ConstantEmitter
     implements ConstantValueVisitor<jsAst.Expression, Null> {
 
-  // Matches blank lines, comment lines and trailing comments that can't be part
-  // of a string.
+  // Matches blank lines, comment lines and trailing comments.
+  // Note: this will include slashes in strings!
   static final RegExp COMMENT_RE =
-      new RegExp(r'''^ *(//.*)?\n|  *//[^''"\n]*$''' , multiLine: true);
+      new RegExp(r'''^ *(//.*)?\n|  *//[^\n]*$''' , multiLine: true);
 
   final Compiler compiler;
   final Namer namer;


diamondo25:
I signed it!

sigmundch:
/cc @rakudrama who originally wrote this.

Thanks for reaching out and sending us a patch!

I was not able to reproduce the issue you noticed in native_helper, though. I believe that's because the first half of the regexp is still in use: we strip any line that only contains spaces and then a comment. By any chance are those files different on your local copy?

If we don't delete a comment, my understanding is that the code will be a bit bigger, but still correct. Is the generate code broken in your case? Or are you using another tool that strips new lines, which then breaks the code?

Because I believe we don't generate invalid code, I prefer that we fix the comments if we need to. If we strip contents of strings, I fear that it will have a semantic effect in the output program. It would then be hard to debug when someone adds code containing such string and he doesn't know about this regexp.

diamondo25:
Its a combination of things.
The line endings of parsed js_const blocks are different (\r only), while the rest is \n. When uploading this file from windows with filezilla on automated transfer mode, itll upload it as ASCII and actually break the line endings to \n only (\r will be deleted). This causes the js_const blocks to be one line (as shown above). The comments that arent stripped give the final blow: itll create commented out code.

sigmundch:
Does this mean that in your set up you'll always be broken if any of our generated code contains //-style comments?

One option here is to match for `\r` as well, but I am worried that there are many places where the SDK assumes it has linux-style line endings.

Is there a way to fix the line-ending issues in your filezilla setup?

diamondo25:
You can configure Filezilla to transfer in binary mode. However, this is unexpected behavior for new users that use FTP (pretty sure it's an FTP implementation issue).
Maybe there's a way to make Rust parse these multi-line strings with newlines they contained (the sourcecode has \n, but it seems to be converted to \r)

sigmundch:
@diamondo25 - just curious if you had any updates about this - Were able to resolve the issue directly under Filezilla or FTP? If not, let us know so we can add support for `\r` -- in that case it would be helpful if you can validate first if something like this works for you:
```dart
    new RegExp(r'''^ *(//.*)?[\n\r]|  *//[^''"\n\r]*$''' , multiLine: true);
```

diamondo25:
@sigmundch Uploading as binary instead of ASCII fixed it, but as it was uploaded using Filezilla with its default settings. I am unable to test that regex because I was helping someone with this issue and made a pull request for it (now its almost a year later).

Issue #25089, opened 2015-12-02
Wrong completion suggestions in annotation


kasperpeulen:
![image](https://cloud.githubusercontent.com/assets/1035299/11530997/afcb54ca-98f9-11e5-9f7f-9a2c81b64b97.png)

Not any of those suggestions are relevant. And the right suggestions are not in the list of suggestions.

This makes libraries that heavily use annotations much harder to explore, than other dart libraries.

kasperpeulen:
I saw this test in the analyzer_server source, and this test is passing.

```dart
  test_Annotation_local_constructor_named_param() {
    addTestSource('''
class A { A({int one, String two: 'defaultValue'}) { } }
@A(^) main() { }''');
    computeFast();
    return computeFull((bool result) {
      assertSuggestArguments(namedArguments: ['one', 'two']);
    });
  }
```

So that seems like there is already logic for this in place?

kasperpeulen:
haha I actually opened this same issue two months ago :P
https://github.com/dart-lang/sdk/issues/24296
and was then fixed by @danrubel 
but it doesn't seem to work for some reason in the editor
not in dartpad, atom or webstorm

kasperpeulen:
maybe interesting, I get different results if I put the annotation class in the same file, that could maybe explain why that test pass

![image](https://cloud.githubusercontent.com/assets/1035299/11549104/5a0f40e0-9963-11e5-8127-4b046d33ba91.png)

kasperpeulen:
this test would also fail:

```
  test_Annotation_imported_constructor_named_param() {
    addSource('/libA.dart', 'library libA; class A{A({int one}){}}');
    addTestSource('import "/libA.dart"; @A() main() { }');
    computeFast();
    return computeFull((bool result) {
      assertSuggestArguments(namedArguments: ['one', 'two']);
    });
  }
```

kasperpeulen:
I think this can be closed, the remaining part of this issue, is the junk that is provided, but that is essentially this issue:
https://github.com/dart-lang/sdk/issues/23992

@danrubel

PR #25098, opened 2015-12-03
suggest argument name completions for annotations that are imported

kasperpeulen:
suggest argument name completions for annotations that are imported

addresses #25089, but doesn't fix it entirely

@danrubel


kasperpeulen:
closing in favor of #25099

PR #25099, opened 2015-12-03
suggest argument name completions for annotations that are imported

kasperpeulen:
suggest argument name completions for annotations that are imported

addresses #25089, but doesn't fix it entirely

changed a typo in the test compared to #25098


kasperpeulen:
I'm not totally sure if this works. I'm not sure how to test this in webstorm for example.

danrubel:
Thank you! This looks good to me. I've recast it as https://codereview.chromium.org/1497773002/ following our standard commit process.

danrubel:
This change has now landed. Thanks!

kasperpeulen:
@danrubel awesome :+1:

PR #25191, opened 2015-12-09
add `not` getter to bool class

kasperpeulen:
When switching from Python to Dart and back, I notice that Python's negations keyword `not` is often easier to read/scan than dart's `!` operator. I often miss the `!` in a long expression. 

I understand that Dart is not going to adopt the Python syntax. But I think adding a `not` getter to the bool class, would be easy to implement and fits the dart syntax very well (imo).

In expressions like:

``` dart
if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync())
```

It is easy to oversee the `!` operator:

``` dart
if (!new File(expandPath('myFancyDir/myFancyFile.txt').existsSync())
```

Therefore, I propose the following syntax:

``` dart
if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync().not)
```

It is not only expressive, but also, easy to type. I rather type an autocompleted `not`, then moving my cursor to the beginning and back.

### DIFF
--- a/sdk/lib/core/bool.dart
+++ b/sdk/lib/core/bool.dart
@@ -40,7 +40,14 @@ class bool {
    */
   external const factory bool.fromEnvironment(String name,
                                               {bool defaultValue: false});
-
+  
+  /// The negated value. For example:
+  ///
+  ///     if (new File('hello.txt').existsSync().not) {
+  ///       print("hello.txt doesn't exist");
+  ///     }
+  bool get not => !this;
+  
   /**
    * Returns [:"true":] if the receiver is [:true:], or [:"false":] if the
    * receiver is [:false:].


Scorpiion:
I never use the `!` syntax like in your example, just for the same reason that I think it harder to read, or I guess more specifically easier to miss and therefor it can create bugs.

In this participial case, I would instead use `== false` since it's less risk of missing it or creating a bug.

However, in the topic of suggested getter syntax, if we would have a getter like that may I suggest using a similar syntax as the `isEmpty` and `isNotEmpty`, that is `isFalse` and `isTrue`.

```
if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync().isFalse)
```

I would prefer that syntax as I find that even easier to understand and also because it aligns nicely with the `isEmpty`/`isNotEmpty` getters.

lrhn:
That would mean that we now have two ways to do the same thing, with most
(existing) code doing one, and some new code doing the other. That's not an
improvement

I agree that the "!" operator is hard to read in many fonts, and it's
occasionally annoying that it's a prefix operator.
That's a curse of using a C legacy syntax (the boon being familiarity), but
I think that's something we'll live with.

In many cases, you can rewrite your if-statements to "if (positiveTest) s2
else s1" instead of "if (!positiveTest) s1 else s2". In others, you can use
de Morgan's laws to move the negation onto a comparison operator, or you
can make an intermediate variable with the negated name "var hasNoFloo =
!biff.hasFoo;" and use that in the test expression.

All in all, I don't think adding a .not getter (nor an .isFalse getter) is
going to be worth it.

(And I *will* reject your CL if you write "x == false" instead of "!x". Any
== or != comparison to a bool, as well as any conditional expression with
bool branches, can be simplified).
/L


On Wed, Dec 9, 2015 at 5:54 PM, Kasper Peulen <notifications@github.com>
wrote:

> When switching from Python to Dart and back, I notice that Python's
> negations keyword not is often easier to read/scan than dart's !
> operator. I often miss the ! in a long expression.
>
> I understand that Dart is not going to adopt the Python syntax. But I
> think adding a not getter to the bool class, would be easy to implement
> and fits the dart syntax very well (imo).
>
> In expressions like:
>
> if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync())
>
> It is easy to oversee the ! operator:
>
> if (!new File(expandPath('myFancyDir/myFancyFile.txt').existsSync())
>
> Therefore, I propose the following syntax:
>
> if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync().not)
>
> It is not only expressive, but also, easy to type. I rather type an
> autocompleted not, then moving my cursor to the beginning and back.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/dart-lang/sdk/pull/25191
> Commit Summary
>
>    - add `not` getter to bool class
>
> File Changes
>
>    - *M* sdk/lib/core/bool.dart
>    <https://github.com/dart-lang/sdk/pull/25191/files#diff-0> (9)
>
> Patch Links:
>
>    - https://github.com/dart-lang/sdk/pull/25191.patch
>    - https://github.com/dart-lang/sdk/pull/25191.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dart-lang/sdk/pull/25191>.
>



-- 
Lasse R.H. Nielsen - lrn@google.com
'Faith without judgement merely degrades the spirit divine'
Google Denmark ApS - Frederiksborggade 20B, 1 sal - 1360 København K
- Denmark - CVR nr. 28 86 69 84

kasperpeulen:
@lrhn I feel you make this PR bigger than it is. This is not a dart language feature request. This is just an innocent little getter added the API of a class.

> That would mean that we now have two ways to do the same thing, with most (existing) code doing one, and some new code doing the other.

I think getters are often used for doing something that could already be done in another way. And often people add new getters to a class, as they find out, as time passes by, that they use some construct so often that a getter is convenient. 

I guess that is the reason someone once added the following getter to `Iterable`

`bool get isEmpty => !iterator.moveNext();`

This is another way of doing something, that could already be done. But it is shorter, more readable, and easier to type. 

Now, you could argue that `!myBool` is shorter than `myBool.not`, but I think more important than shortness, is that it easy to read, and also, very easy to type in a modern editor with autocompletion. I think easier, to type than adding the `!` keyword, because you often need to put your cursor back to the beginning.

```
if (!myList.where((i) => i > 10).isEmpty)
```

I think for similar reason (easier to read, easier to type), many times libraries actually do also add the negated getter. 

```
bool get isNotEmpty => !isEmpty;
```

In this example, we could also write this as:

```
if (myList.where((i) => i > 10).isNotEmpty)
```

Now this is exactly the same kind of getter that I propose :) But this getter is much more worthwhile, as now a library doesn't have to ship with a negated variant. We could just write:

```
if (myList.where((i) => i > 10).isEmpty.not)
```

If there is one thing inconsistent, then it is that some API's do ship with this negated variant, and some libraries only add the positive test (`.isVisible`/ `.isReady`/ etc.) With the `bool.not` getter libraries only have to provide one variant. When Dart 2.0 comes in, all the `isNotEmpty` kind of variants could be removed, to make things consistent.

> In many cases, you can rewrite your if-statements to `if (positiveTest) s2 else s1` instead of `if (!positiveTest) s1 else s2`. 

But this is also another way of doing the same thing! But I think this is even worse, because there is no clear winner, sometimes `!` doesn't hurt, sometimes it would be really more readable if you refactor it a bit.

With the `.not` getter, you could make a linter rule, that always prefers `.not` above `!`. Then at least on a project basis, code is consistent. And I think that is where the biggest productivity win lays (of having only one option).

lrhn:
>
> @lrhn <https://github.com/lrhn> I feel you make this PR bigger than it
> is. This is not a dart language feature request. This is just an innocent
> little getter added the API of a class.
>
Adding something to the core API is never innocent. It shows intent, it
suggests that you should use it.
But if you use it, you change the style of the code to be different from
other Dart code - you are introducing a dialect, the `x.not` dialect which
is different from the `!x` dialect.
We could do that if we actually though using `x.not` was better, and we
would then encourage everybody to switch to that.
I personally don't find it better, and wouldn't encourage it. That leaves
it at "user it or not, whatever you feel like", and we will have two
competing dialects which makes reading code harder for everybody. There is
no longer an idiomatic way of writing things, and you have to mentally
switch back and forth between the two different ways to write the same
thing.

That would mean that we now have two ways to do the same thing, with most
> (existing) code doing one, and some new code doing the other.
>
> I think getters are often used for doing something that could already be
> done in another way. And often people add new getters to a class, as they
> find out, as time passes by, that they use some construct so often that a
> getter is convenient.
>
> I guess that is the reason someone once added the following getter to
> Iterable
>
> bool get isEmpty => !iterator.moveNext();
>
> This is another way of doing something, that could already be done. But it
> is shorter, more readable, and easier to type.
>
Absolutely correct. It's way shorter, and it reads like what you are
actually checking: Is the iterable empty.
Reading `!iterator.moveNext()` doesn't say that, you have to parse it and
understand that if the first call to `moveNext` returns false, then there
was no elements. It's an abstraction - we give a short name to a more
complex computation.

An `x.not` doesn't say anything different from `!x`. It expresses exactly
the same thing, there is no abstraction, only an alias.


> Now, you could argue that !myBool is shorter than myBool.not, but I think
> more important than shortness, is that it easy to read, and also, very easy
> to type in a modern editor with autocompletion. I think easier, to type
> than adding the ! keyword, because you often need to put your cursor back
> to the beginning.
>
> if (!myList.where((i) => i > 10).isEmpty)
>
> I think for similar reason (easier to read, easier to type), many times
> libraries actually do also add the negated getter.
>
> bool get isNotEmpty => !isEmpty;
>
> In this example, we could also write this as:
>
> if (myList.where((i) => i > 10).isNotEmpty)
>
> Now this is exactly the same kind of getter that I propose :)
>

No, this is still an abstraction. We just found that isEmpty and isNotEmpty
were equally valid tests to abstract. It's not trying to remove the `!`
(even if doing that does make the expression more readable), it is
abstracting over `iterator.moveNext() == true`.


> But this getter is much more worthwhile, as now a library doesn't have to
> ship with a negated variant. We could just write:
>
> if (myList.where((i) => i > 10).isEmpty.not)
>
>
Ok, I have to say that I don't think this reads very well. I'm not
particularly fond of "That's funny, NOT!" in natural language either :).

You read da...da...da..isEmpty (ok, I got it) ... NOT! ... argh, now I have
to revise the thing I thought I understood.
Reading the "not" in front sets context of the expression you are reading.
That is, I don't buy that it is actually easier to read due to the
ordering.
I will accept that `!` is a character that is easy to overlook, but putting
a "not" at the end is not how I would fix that.


> If there is one thing inconsistent, then it is that some API's do ship with this negated variant, and some libraries only add the positive test (.isVisible/ .isReady/ etc.) With the bool.not getter libraries only have to provide one variant. When Dart 2.0 comes in, all the isNotEmpty kind of variants could be removed, to make things consistent.
>
> In many cases, you can rewrite your if-statements to if (positiveTest) s2
> else s1 instead of if (!positiveTest) s1 else s2.
>
> But this is also another way of doing the same thing! But I think this is
> even worse, because there is no clear winner, sometimes ! doesn't hurt,
> sometimes it would be really more readable if you refactor it a bit.
>
That's why we have code reviews - a second set of eyes can often tell you
if you have written something in a way that isn't very readable. Removing a
negation from a test is usually a good suggestion, but in the end, what is
readable to humans can only be decided by humans.


> With the .not getter, you could make a linter rule, that always prefers
> .not above !.
>
We could, and I (personally) don't want that. And if we don't do that, I
don't want to have `.not` at all because I don't want half the code written
one way and the other half written the other way.

I might be unique, but I doubt it. Using `!` in front is the habit of every
C, C++, Java, C# and JavaScript programmer, the languages users we have
chosen to make Dart be (somewhat) familiar to.


> Then at least on a project basis, code is consistent. And I think that is
> where the biggest productivity win lays (of having only one option).
>
It's still two dialects for a somewhat small win (or loss, depending on
your preferences).

Scorpiion:
While I understand the resistance towards adding a new way of doing things, I would not agree with the statement:

> Using `!` in front is the habit of every C, C++, Java, C# and JavaScript programmer

Personally I never use it. I have no statistics but if I were to guess I would say that it's way more common to use either `== false` or `!= true`, these two combined I think is way more common than expressions using `!`. The same way as `if () {}` statements are way more common than usage of the `?:` ternary operator. Some programmers might of course always use `!` when they can, but I think most don't.

I do agree that `.not` is not that much better for readability either, if an easier to use getter would be added I would prefer `isFalse` and `isTrue`, similar to `isEmpty`/`isNotEmpty`.

This example:
```
if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync().isFalse)
```

Does read quite well and as @kasperpeulen suggested with an auto completing IDE it might be quite nice to use instead of either of the three different alternatives; `!`, `== false`, `!= true`.

eernstg:
There are two separate issues here:

(1) The prefix `!` may simply be so small (and maybe so similar to the
neighboring parenthesis) that it's overlooked entirely; that should be
fixable by using a bigger token, maybe a prefix keyword 'not' or something
like that ... but we don't want to pay for that in terms of breaking every
Dart program in the universe (that we know of), making programs more
verbose, and alienating everyone who is used to a prefix `!` and frankly
couldn't care much less. (Surely some people don't care so much about
syntax. Surely. I'm absolutely sure).  We could also simply put a rule like
"always ensure that there is a space before `!`" into the style guide and
`dartfmt`, no I did not write that.  ;-)

The other issue is that (2) the negation could be at the other end (as a
suffix operator or a method invocation).

I generally don't think the advice from Knuth (associated with literate
programming) that code should "read like literature" should be followed
slavishly, but it still seems to be a reasonable sanity check to try to
read code snippets aloud and see if they make sense as a kind of natural
language.

In English natural language it is common to embed the negation in the
middle of the verb phrase ("I do not think..", "It won't matter", etc.),
which tends to support variant method calls like `isNotEmpty` and operators
like `!=` (rather than using the prefix `!`, far far away from the "verb").

That doesn't do much for a suffix `.not`, though, especially because you
might think differently about a pure test (if you start with `!` then you
really wouldn't expect side-effects) and an operation producing an entity
with a built-in success reporter (

    if (new File(..).writeAsStringSync().okInSomeSense(..)) {
      .. /* proceed, trusting that the file was written */ ..
    }

), and the terminating `not` might, as Lasse mentioned, force you to
reinterpret a non-trivial sequence of steps. Certainly, your line of
thinking will flow smoothly into the brain of every reader if you negate at
the end, NOT!  ;^)

But there is a fine line here: `<big_thing>.isEmpty.not` might read pretty
much as smoothly as `<big_thing>.isNotEmpty`, although that might as well
be a hint that `<big_thing>.isEmpty` is also ill shaped in the first place
and something like `if (isEmpty: <big_thing>) ..` would read better. "State
your main goals up front" would probably be a good motto when making these
micro-organizational choices, if possible.


On Fri, Dec 11, 2015 at 11:03 AM, Robert Åkerblom-Andersson <
notifications@github.com> wrote:

> While I understand the resistance towards adding a new way of doing
> things, I would not agree with the statement:
>
> Using ! in front is the habit of every C, C++, Java, C# and JavaScript
> programmer
>
> Personally I never use it. I have no statistics but if I were to guess I
> would say that it's way more common to use either == false or != true,
> these two combined I think is way more common than expressions using !.
> The same way as if () {} statements are way more common than usage of the
> ?: ternary operator. Some programmers might of course always use ! when
> they can, but I think most don't.
>
> I do agree that .not is not that much better for readability either, if
> an easier to use getter would be added I would prefer isFalse and isTrue,
> similar to isEmpty/isNotEmpty.
>
> This example:
>
> if (new File(expandPath('myFancyDir/myFancyFile.txt').existsSync().isFalse)
>
> Does read quite well and as @kasperpeulen
> <https://github.com/kasperpeulen> suggested with an auto completing IDE
> it might be quite nice to use instead of either of the three different
> alternatives; !, == false, != true.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dart-lang/sdk/pull/25191#issuecomment-163896384>.
>



-- 
Erik Ernst  -  Google Danmark ApS
Skt Petri Passage 5, 2 sal, 1165 København K, Denmark
CVR no. 28866984

lrhn:
On Fri, Dec 11, 2015 at 12:50 PM, Erik Ernst <notifications@github.com>
wrote:

> In English natural language it is common to embed the negation in the
> middle of the verb phrase ("I do not think..", "It won't matter", etc.),
> which tends to support variant method calls like `isNotEmpty` and operators
> like `!=` (rather than using the prefix `!`, far far away from the "verb").
>

I did have a thought (but decided to not post it to muddy waters even more)
that you could insert `!` in the middle of expressions as `x.!isEmpty`.
It would be equivalent to (!(x.isEmpty)), but without the need for
parentheses.

Again the operator's appearance makes that pretty unreadable, so it's
probably not a good idea.
It doesn't generalize to anything but unary prefix operators (the other
ones being ~, and potentially ++ and --).

floitschG:
I think Lasse's latest comment captures the real problem:
"!" is a prefix operator for the whole expression. This means, that readers have to (mentally) push the not-operator on the stack, while parsing the rest of the expression, before finally popping it once the expression has been parsed. This adds overhead when reading code.
Unfortunately, this is the standard way of writing not-expressions.

lrhn:
We don't plan to land this change. 
For the reasons given above, I don't think this is a clear an improvement, nor would it be even if changed to `.isTrue` or `.isFalse`.

Issue #25328, opened 2016-01-02
Do not hardcode #!/bin/bash in shell scripts


mulander:
Related to the BSD porting efforts ongoing in https://github.com/dart-lang/sdk/issues/10260.

BSD systems don't place bash in /bin and a lot of the dart tools hardcode a #!/bin/bash shebang that fails the 'all' target build (not able to execute dart2js since the interpreter is not found).

I would like to replace those paths with `#!/usr/bin/env bash` which should work on Linux, Mac OS X & all BSD's.

Places with a hardcoded shebang are as follows:

```
$ git grep /bin/bash
pkg/analysis_server/tool/spec/generate_files:#!/bin/bash
pkg/analyzer/tool/generate_files:#!/bin/bash
runtime/tools/bmu_benchmark_gallery.sh:#!/bin/bash
sdk/bin/dart:#!/bin/bash
sdk/bin/dart2js:#!/bin/bash
sdk/bin/dart2js_developer:#!/bin/bash
sdk/bin/dart2js_sdk:#!/bin/bash
sdk/bin/dartanalyzer:#!/bin/bash
sdk/bin/dartanalyzer_developer:#!/bin/bash
sdk/bin/dartanalyzer_sdk:#!/bin/bash
sdk/bin/dartdoc:#!/bin/bash
sdk/bin/dartfmt:#!/bin/bash
sdk/bin/dartfmt_sdk:#!/bin/bash
sdk/bin/pub:#!/bin/bash
sdk/bin/pub_sdk:#!/bin/bash
site/try/create_manifest.sh:#!/bin/bash
tests/standalone/io/create_sample_certificates.sh:#!/bin/bash
tests/standalone/io/dart_std_io_pipe_test.sh:#! /bin/bash
tools/apps/update_homebrew/bin/ssh_with_key:#!/bin/bash
tools/apps/update_homebrew/bin/update_homebrew.dart:      #!/bin/bash
tools/bots/run_android_tests.sh:#!/bin/bash
tools/create_debian_chroot.sh:#!/bin/bash
tools/dartium/download_shellscript_template.sh:#!/bin/bash
tools/dartium/start_dartium_roll.sh:#!/bin/bash -e
tools/dom/scripts/go.sh:#!/bin/bash -x
tools/precompilation/test_linux.sh:#!/bin/bash -ex
tools/precompilation/test_linux_simarm.sh:#!/bin/bash -ex
tools/precompilation/test_macos.sh:#!/bin/bash -ex
[...]

PR #25330, opened 2016-01-02
Refs #25328 Use a portable shebang

mulander:
BSD systems don't place bash in /bin and a lot of the dart tools
hardcode a #!/bin/bash shebang that fails the 'all' target build
(not able to execute dart2js since the interpreter is not found).

Solve the issue by using #!/usr/bin/env as the shebang. For scripts
that need to pass arguments to bash modify the script to use the set
command as the first executed line of shell.

### DIFF
--- a/pkg/analysis_server/tool/spec/generate_files
+++ b/pkg/analysis_server/tool/spec/generate_files
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
 # for details. All rights reserved. Use of this source code is governed by a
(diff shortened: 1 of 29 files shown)


kaendfinger:
LGTM :+1:

mulander:
Related CL https://codereview.chromium.org/1552313002

kevmoo:
We landed this https://github.com/dart-lang/sdk/commit/634e5a1d02f64cbb0922161433366fde3fd19b4d, right @whesse ?

whesse:
Yes, we landed it.

kevmoo:
https://github.com/dart-lang/sdk/commit/634e5a1d02f64cbb0922161433366fde3fd19b4d

PR #25343, opened 2016-01-05
allow for selectively writing begin/end doc comments markers

guillermooo:
If a marker is null, it won't be written. For all other strings,
they'll be written as they are, including the empty string.

This enables nicer output for languages that don't have begin/end
markers for doc comments, such as Python.


bwilkerson:
LGTM

bwilkerson:
Let me know if you want/need me to merge it for you.

guillermooo:
@bwilkerson I don't think I can merge this myself, so please do!

Issue #12921, opened 2013-08-30
Allow users to specify type arguments in ClassMirror.newInstance


rmacnak-google:
Maybe?

Or maybe specify that the same arguments from the ClassMirror will be used if it is a mirror on an instantiated generic and all dynamic otherwise?

pulyaevskiy:
Is this something that requires DEP or can be implemented and submitted in a PR?

eernstg:
I believe this is subsumed by the feature requested in https://github.com/dart-lang/sdk/pull/26012.

natebosch:
I think we can close this since it's covered by type arguments in `reflectType`

PR #26012, opened 2016-03-16
reflectType() dynamic type arguments support

pulyaevskiy:
Hey team,

I decided to post this here as a proof-of-concept implementation for #12921.
This is by no means finished yet, but I'd like to get some general feedback on the approach.

This implementation differs from what @gbracha suggested in the original issue. For some reason I wasn't able to get this working for `reflectClass()`, I'm still not sure if I get semantics of this function right (looks like it or `ClassMirror` does not handle typeArguments at all).

However for `reflectType()` it did work. There is many open questions still and I'm willing to continue this work and get it to finished state. If proposed approach makes sense to the team, of course.

Also, even though this doesn't enable #12921 use case directly, it would still be possible with something like:

``` dart
class A<T> { }
class B { }
var typeMirror = reflectType(A, [B]); // TypeMirror on 'A<B>'
var classMirror = reflectClass(typeMirror.reflectedType);
var instance = classMirror.newInstance(new Symbol(""), []).reflectee;  // Instance 'A<B>'
```

However this doesn't seem to work as expected and type argument information is lost for `instance` variable, even on stable version (1.15). So it doesn't seem like regression.


pulyaevskiy:
I signed it!

gbracha:
Thanks. We'll take a look. Give us a few days.

pulyaevskiy:
Sure, thanks!

In the meantime I pushed basic support for js_mirrors, which turned out to be a lot easier to achieve than I thought (the test case is passing at least in chrome).

I'd like to emphasize that this PR is by no means finished yet:

* I haven't updated the docs
* Several edge cases are not covered, like checking number of passed type arguments against actual number of arguments defined in type being reflected.
* There is caching issue in the VM where we basically can't cache calls with `typeArguments` parameter  provided.
* Should the new parameter be optional or named?
* Add more test cases

That being said, basic use case is implemented and there is test case which passes in VM and in Chrome.

Also, according to documentation in code, looks like `ClassMirror` is not supposed to handle type arguments at all and always refer to "declaration" of a particular class. I tested it with following snippet which confirms my assumption:

```dart
class A<T> {}
class B {}
var type = new A<B>().runtimeType; // A<B>
var m = reflectClass(type);
var o = m.newInstance(new Symbol(""), []).reflectee;
print(o); // Instance of 'A'
print(reflect(o).type.typeArguments); // [TypeMirror on 'dynamic']
```

Dartpad: https://dartpad.dartlang.org/b491835a3df19dd98885

Just to be clear: for my personal needs I don't really need this functionally and current implementation in this PR is sufficient enough.

mit-mit:
@gbracha any update?

mit-mit:
CC @floitschG

pulyaevskiy:
Just another friendly reminder here.
Still very much interested in working on this.

floitschG:
@pulyaevskiy: I will work with you on this feature.

Note that we generally discourage the use of mirrors because of their detrimental impact on performance and code size. In the long run it could be that we deprecate the mirrors-library entirely.
Until then, this feature looks useful, and I will work with you on getting it into the system.

I will start by reviewing your PR. I hope this won't take too much time. In any case, with exception for some vacations in July, I should be responsive. In general, don't hesitate to ping much more often.

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
+      var typeMirror = _instantiationCache[key];
+      if (typeMirror == null) {
+        typeMirror = makeLocalTypeMirror(key, arguments);
+        _instantiationCache[key] = typeMirror;
+        if (typeMirror is ClassMirror && !typeMirror._isGeneric) {
+          _declarationCache[key] = typeMirror;
+        }
       }
+      return typeMirror;
+    } else {
```
floitschG: No need for the "else" since the 'then' finishes with a 'return'.

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
+      // We can't cache type provided in `key` since actual type will depend
+      // on resolving provided `key` and type arguments.
+      typeArguments ??= [];
```
floitschG: typeArgumenst can't be 'null'. You just checked this in line 1668.

runtime/lib/mirrors.cc
```diff
@@ -840,9 +840,29 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+    TypeArguments& type_args_obj = TypeArguments::Handle();
+    type_args_obj ^= TypeArguments::New(num_expected_type_arguments);
+    AbstractType& type_arg = AbstractType::Handle();
+    for (intptr_t i = 0; i < args.Length(); i++) {
+      type_arg ^= args.At(i);
+      type_args_obj.SetTypeAt(i, type_arg);
+    }
+    const Class& clz = Class::Handle(type.type_class());
+    Type& instantiated_type = Type::Handle(
+        Type::New(clz, type_args_obj, TokenPosition::kNoSource));
```
floitschG: I guess we need to check that the number of generic arguments is actually correct.

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
+      if (typeMirror == null) {
+        typeMirror = makeLocalTypeMirror(key, arguments);
+        _instantiationCache[key] = typeMirror;
+        if (typeMirror is ClassMirror && !typeMirror._isGeneric) {
+          _declarationCache[key] = typeMirror;
+        }
       }
+      return typeMirror;
+    } else {
+      // We can't cache type provided in `key` since actual type will depend
```
floitschG: I'm not sure I understand the comment. There is no caching in the line that follows.
Also it actually *does* cache the type eventually (line 1689)

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
-    if (typeMirror == null) {
-      typeMirror = makeLocalTypeMirror(key);
-      _instantiationCache[key] = typeMirror;
-      if (typeMirror is ClassMirror && !typeMirror._isGeneric) {
-        _declarationCache[key] = typeMirror;
+  static TypeMirror reflectType(Type key, [Iterable<Type> typeArguments]) {
+    if (typeArguments == null || typeArguments.isEmpty) {
+      var typeMirror = _instantiationCache[key];
+      if (typeMirror == null) {
+        typeMirror = makeLocalTypeMirror(key, arguments);
```
floitschG: Where does the "arguments" come from?
Should this simply be `const []`?

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
+      return typeMirror;
+    } else {
+      // We can't cache type provided in `key` since actual type will depend
+      // on resolving provided `key` and type arguments.
+      typeArguments ??= [];
+      List arguments = typeArguments.toList(growable: false);
+      var resolvedTypeMirror = makeLocalTypeMirror(key, arguments);
+      var actualType = resolvedTypeMirror.reflectedType;
+
+      // We check if actual type has been cached already and return cached copy.
```
floitschG: Nit. ... "if the actual type" ...

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
+      var cachedTypeMirror = _instantiationCache[actualType];
+      if (cachedTypeMirror == null) {
+        _instantiationCache[actualType] = resolvedTypeMirror;
+        if (resolvedTypeMirror is ClassMirror && !resolvedTypeMirror._isGeneric) {
```
floitschG: Can this ever be true?
In theory, the `resolvedTypeMirror` must be a generic class, since could provide generic arguments.

floitschG:
/cc @rmacnak-google: could you have a look at the c++ code and make sure I didn't miss anything? Also have a look at the questions below.

There is not much missing.
As you stated:
- update the docs.
- we need to decide if we want to cache the types. In theory there is an unbound number, contrary to the current cache. Either we don't cache, or we use a size-limited cache. Ryan, what do you think?
- a few more tests.

I think making it an optional argument is fine.

Thanks, and sorry this took so long.

runtime/lib/mirrors_impl.dart
```diff
@@ -1646,7 +1646,7 @@ class _Mirrors {
 
   static ClassMirror makeLocalClassMirror(Type key)
       native "Mirrors_makeLocalClassMirror";
-  static TypeMirror makeLocalTypeMirror(Type key)
+  static TypeMirror makeLocalTypeMirror(Type key, typeArguments)
```
rmacnak-google: There should be a separate native that does the instantiation step.

runtime/lib/mirrors_impl.dart
```diff
@@ -1664,15 +1664,35 @@ class _Mirrors {
...
   }
 
-  static TypeMirror reflectType(Type key) {
-    var typeMirror = _instantiationCache[key];
-    if (typeMirror == null) {
-      typeMirror = makeLocalTypeMirror(key);
-      _instantiationCache[key] = typeMirror;
-      if (typeMirror is ClassMirror && !typeMirror._isGeneric) {
-        _declarationCache[key] = typeMirror;
+  static TypeMirror reflectType(Type key, [Iterable<Type> typeArguments]) {
```
rmacnak-google: Iterable -> List

rmacnak-google:
```
reflectType(Type key, [List<Type> typeArguments]) {
  if (typeArguments != null) {
    key = _instantiateType(key, typeArguments);
  }
  <old code>
}
```

sdk/lib/mirrors/mirrors.dart
```diff
@@ -178,7 +178,7 @@ external ClassMirror reflectClass(Type key);
  * function can only be used to obtain type mirrors on types of the current
  * isolate.
  */
-external TypeMirror reflectType(Type key);
+external TypeMirror reflectType(Type key, [Iterable<Type> typeArguments]);
```
rmacnak-google: List

tests/lib/mirrors/reflected_type_generics_test.dart
```diff
@@ -0,0 +1,18 @@
...
+
+import 'package:expect/expect.dart';
+
+import 'reflected_type_helper.dart';
+
+class A<T> {}
+class G {}
+
+main() {
+  expectReflectedType(reflectType(A, [G]), new A<G>().runtimeType);
```
rmacnak-google: Add tests for
 - providing empty list of type arguments for a non-generic type.
 - providing some type arguments for a non-generic type
 - providing too many / too few type argument for a generic type
 - a non-type in the type arguments list

And cases where the VM's internal representation of the type have a different number of type argument that the language model:

```
class Super<T> { }
class Sub extends Super<int> {}
reflectType(Sub, [String]) // error
```

```
class Super<T> {}
class Sub<S> extends Super<S> {}
reflectType(Sub, [String]) // okay
reflectType(Sub, [String, String]) // error
```

pulyaevskiy:
@floitschG @rmacnak-google 

Great, thanks! Sorry for late response, I'm on vacation right now. Will take a look when I get back later this week.

mit-mit:
Hi @pulyaevskiy, any updated on this one?

pulyaevskiy:
@mit-mit Thanks for pinging me! I apologize for the delay.  
I got a bit overwhelmed after vacation, but I'm almost done with the CR fixes. Will push changes later today!

pulyaevskiy:
I just pushed some changes. This addresses most of the comments from @floitschG and @rmacnak-google . However this work is not final yet.

I forgot about JS part of mirrors. Will need to spend some more time there, but I'm pretty close, just want to make sure it behaves consistently in different runtimes. 

One thing I couldn't figure yet is how to properly do `instanceof` checks in C code to verify that list of type arguments actually contains only instances of `Type`, but will get there.

Thanks again for helping me out with this!

eernstg:
For the record, a `ClassMirror` does include type arguments, but `reflectClass` was specifically intended to provide a mirror for "the class declaration as such" which is interpreted to mean `A<dynamic>` for the generic instantiation `A<B>`. So we just need to omit the step that erases the type argument. For example:

```dart
import 'dart:mirrors';

class A<T> {}
class B {}

main() {
  Type type = new A<B>().runtimeType; // The `Type` for `A<B>`.
  TypeMirror typeMirror = reflectType(type);
  if (typeMirror is ClassMirror) { // Yes, it is.
    Object o = typeMirror.newInstance(const Symbol(""), []).reflectee;
    print(o); // Prints "Instance of 'A<B>'".
    print(reflect(o).type.typeArguments); // Prints "[ClassMirror on 'B']".
  }
}
```

So this feature should suffice to solve the problem described in https://github.com/dart-lang/sdk/issues/12921.

pulyaevskiy:
@eernstg I think you are right (and I've just tested it, actually). This feature solves #12921 indirectly.
And it removes reliance on constructors as in your example:

```
  Type type = new A<B>().runtimeType; // The `Type` for `A<B>`.
  TypeMirror typeMirror = reflectType(type);
```

Which is now possible with this:

```
  TypeMirror typeMirror = reflectType(A, [B]);
```

pulyaevskiy:
@floitschG @rmacnak-google Sorry it took so long, but I think I addressed all of your comments.

I've discovered one interesting edge case which I haven't fixed yet and wanted to bring up here:

```dart
class Super<K, V> {}
class Sub<K> extends Super<K, int> {}

// This is failing currently:
expectReflectedType(reflectType(Sub, [String]), new Sub<String>().runtimeType);
// throws RuntimeError with Invalid number of type arguments
```

I need to figure out how to check if there are pre-existing resolved type arguments in the class and use those when instantiating it.

But this seems to me like the only thing left here. Please let me know if I missed something!

pulyaevskiy:
Also, there are some changes in the diff caused by `dartfmt`. Please let me know if this is undesirable, I can try to revert.

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,49 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+DEFINE_NATIVE_ENTRY(Mirrors_makeLocalTypeMirrorWithTypeArguments, 2) {
+  GET_NON_NULL_NATIVE_ARGUMENT(AbstractType, type, arguments->NativeArgAt(0));
+  GET_NON_NULL_NATIVE_ARGUMENT(Array, args, arguments->NativeArgAt(1));
+
+  const Class& clz = Class::Handle(type.type_class());
+  if (!clz.IsGeneric()) {
+    const Array& errorArgs = Array::Handle(Array::New(3));
+    errorArgs.SetAt(0, type);
+    errorArgs.SetAt(1, String::Handle(String::New("key")));
+    errorArgs.SetAt(2, String::Handle(String::New("Type must be a generic class")));
```
floitschG: Nit, but I think I would terminate the message with a ".".

runtime/lib/mirrors_impl.dart
```diff
@@ -1676,4 +1681,18 @@ class _Mirrors {
     }
     return typeMirror;
   }
+
+  static Type _resolveType(Type key, List<Type> typeArguments) {
+    if (!typeArguments.every((_) => _ is Type)) {
```
floitschG: I think you would would want to make sure that this is actually a '_Type' and not a user-defined subclass of `Type`.

Not sure if you have access to `_Type`, though.

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
```diff
@@ -589,8 +587,18 @@ InstanceMirror reflect(Object reflectee) {
   }
 }
 
-TypeMirror reflectType(Type key) {
-  return reflectClassByMangledName(getMangledTypeName(key));
+TypeMirror reflectType(Type key, [List<Type> typeArguments]) {
+  String mangledName = getMangledTypeName(key);
+  if (typeArguments != null) {
+    if (!typeArguments.every((_) => _ is Type)) {
```
floitschG: Test that it's actually a `TypeImpl` and not just a user-implemented `Type` class.

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
```diff
@@ -589,8 +587,18 @@ InstanceMirror reflect(Object reflectee) {
...
+      throw new ArgumentError.value(typeArguments, 'typeArguments');
+    }
+    var mangledTypeArguments =
+        typeArguments.map((_) => getMangledTypeName(_));
```
floitschG: `typeArguments.map(getMangedTypeName)`

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
```diff
@@ -612,11 +620,27 @@ TypeMirror reflectClassByName(Symbol symbol, String mangledName) {
...
       throw new UnimplementedError();
     }
-    mirror = new JsTypeBoundClassMirror(originalDeclaration,
+    if (originalDeclaration.typeVariables.isEmpty) {
+      throw new ArgumentError.value(originalDeclaration.reflectedType);
+    }
+    List<String> typeArguments = mangledName
+        .substring(typeArgIndex + 1, mangledName.length - 1)
+        .split(', ')
+        .where((_) => _.isNotEmpty)
```
floitschG: Why is this necessary?

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
```diff
@@ -1562,14 +1569,12 @@ class JsSyntheticAccessor implements MethodMirror {
...
-  final _target;  /// The field or type that introduces the synthetic accessor.
+  final _target;
 
-  JsSyntheticAccessor(this.owner,
-                      this.simpleName,
-                      this.isGetter,
-                      this.isStatic,
-                      this.isTopLevel,
-                      this._target);
+  /// The field or type that introduces the synthetic accessor.
```
floitschG: Please fix this comment. It should be before the `_target` field.

floitschG:
@rmacnak-google: could you again have a look at the C++ code please?
@rmacnak-google: the failing test only fails on the VM. Could you help us find the code to fill in the generic types for the super?

wrt dartfmt: would probably be a good idea to just run the formatter on the corresponding files first, but I already did the review, so it doesn't really matter too much anymore.

Otherwise LGTM.

runtime/lib/mirrors.cc
pulyaevskiy: Thanks! This is interesting, I've been trying to follow examples in existing code and everywhere I could find there was no terminating `.`.
Examples: [1](https://github.com/dart-lang/sdk/blob/master/runtime/lib/mirrors_impl.dart#L1074), [2](https://github.com/dart-lang/sdk/blob/master/runtime/lib/mirrors_impl.dart#L1151), [3](https://github.com/dart-lang/sdk/blob/master/runtime/lib/mirrors_impl.dart#L1560).

Curious if this is reflected in a style guide somewhere or just a personal preference thing?
(I'd personally prefer to terminate with a `.` as well, but staying consistent is a higher priority).

floitschG: Hmmm.
Grepping through our sources, I can unfortunately see both. Even worse, the errors themselves sometimes add a dot to the message and sometimes they don't.
ArgumentError doesn't add a dot, so it should be safe to add one.
There is really no consistency right now. Do as you please.
We will (hopefully) eventually go through them and clean them up.

runtime/lib/mirrors_impl.dart
pulyaevskiy: Looks like you're right and I don't have access to `_Type` (also not sure I entirely understand a concept of `patch` used in some places in the SDK and `_Type` is a patch). I might need to figure out how to do this on the C side.

pulyaevskiy:
Thanks @floitschG ! The remaining C++ part will require some more code learning from me. I will look at it over the weekend and push updates/fixes as soon as I can.

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
pulyaevskiy: Done: 5fdc5b9

runtime/lib/mirrors.cc
pulyaevskiy: Added dots: 5fdc5b9

runtime/lib/mirrors_impl.dart
pulyaevskiy: Figured out how to do this in C: 5fdc5b9 ( [actual line(s)](https://github.com/dart-lang/sdk/pull/26012/files#diff-627364e40d12c87611f09ab984c5c521R872) ).
Sanity check from @rmacnak-google would be great though!

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
pulyaevskiy: Thanks! Done: 5fdc5b9.

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
pulyaevskiy: Must write more Dart to get used to it.
5fdc5b9.

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
pulyaevskiy: I agree it's awkward. Unfortunately it's the way `split` works on empty strings (results in `[""]`).

I moved `isEmpty` check on one level up the chain, since this would only be possible with a class name like `Foo<>`, which is not entirely valid. So preventing such behavior seems like a better solution to me.
5fdc5b9.

pulyaevskiy:
Small update, I pushed some changes to address @floitschG comments.
Figured out how to check whether an object is a `Type` in C++.

The only thing left is the edge case with `class Super<K> extends Sub<K, int>`.

pulyaevskiy:
@floitschG @rmacnak-google 

I think I resolved the last issue with `class Sub<K> extends Super<K, int>` use case.
I probably I got confused by terminology but now I think I got it right:

TypeArgument (in C++) == TypeArgument (in Dart)
TypeParameter (in C++) == TypeVariable (in Dart)

Looking forward for any remaining comments!

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+    errorArgs.SetAt(1, String::Handle(String::New("key")));
+    errorArgs.SetAt(2, String::Handle(String::New("Type must be a generic class.")));
+    Exceptions::ThrowByType(Exceptions::kArgumentValue, errorArgs);
+    UNREACHABLE();
+  }
+  if (clz.NumTypeParameters() != args.Length()) {
+    const Array& errorArgs = Array::Handle(Array::New(3));
+    errorArgs.SetAt(0, args);
+    errorArgs.SetAt(1, String::Handle(String::New("typeArguments")));
+    errorArgs.SetAt(2, String::Handle(String::New("Number of type arguments does not match.")));
```
floitschG: 80 chars.

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+  type_args_obj ^= TypeArguments::New(num_expected_type_arguments);
+  AbstractType& type_arg = AbstractType::Handle();
+  Instance& instance = Instance::Handle();
+  for (intptr_t i = 0; i < args.Length(); i++) {
+    instance ^= args.At(i);
+    if (!instance.IsType()) {
+      const Array& errorArgs = Array::Handle(Array::New(3));
+      errorArgs.SetAt(0, args);
+      errorArgs.SetAt(1, String::Handle(String::New("typeArguments")));
+      errorArgs.SetAt(2, String::Handle(String::New("Type arguments must be instances of Type.")));
```
floitschG: 80 chars.

runtime/lib/mirrors_impl.dart
```diff
@@ -1676,4 +1681,20 @@ class _Mirrors {
     }
     return typeMirror;
   }
+
+  static Type _resolveType(Type key, List<Type> typeArguments) {
+    if (typeArguments.isEmpty) {
+      throw new ArgumentError.value(
+        typeArguments, 'typeArguments', 'Type arguments list can not be empty.'
```
floitschG: This doesn't look like normal indentation. The ");" should not be on a new line.

runtime/lib/mirrors_impl.dart
```diff
@@ -1665,7 +1667,10 @@ class _Mirrors {
     return classMirror;
   }
 
-  static TypeMirror reflectType(Type key) {
+  static TypeMirror reflectType(Type key, [List<Type> typeArguments]) {
+    if (typeArguments != null) {
+      key = _resolveType(key, typeArguments);
```
floitschG: Maybe just `return _reflectGenericType(key, typeArguments)` and change the name of `_resolveType (and its return value).
The current code should work, though.

sdk/lib/_internal/js_runtime/lib/js_mirrors.dart
```diff
@@ -475,13 +472,12 @@ class JsLibraryMirror extends JsDeclarationMirror with JsObjectMirror
       }
       bool isConstructor = unmangledName.startsWith('new ');
       bool isStatic = !isConstructor; // Top-level functions are static, but
-                                      // constructors are not.
+      // constructors are not.
```
floitschG: Could you please fix this comment as well.
I would just put it above the `bool isStatic = !isConstructor;` line.

sdk/lib/mirrors/mirrors.dart
```diff
@@ -174,11 +174,17 @@ external ClassMirror reflectClass(Type key);
  * If [key] is not an instance of [Type], then this function throws an
  * [ArgumentError].
  *
+ * One can optionally pass a list of [typeArguments] for a generic class.
```
floitschG: "Optionally takes a list of [typeArguments] for generic classes. If the list is provided, then the [key] must be a generic class type, and the number of the provided type arguments must be equal to the number of type parameters declared by the class."

As a general advice, don't use future tense in dartdocs unless something happens after the function has returned.

Don't declare which error class is used when errors happen. Errors (contrary to exceptions) are bugs and should only happen during development. There should never be the need to catch them specifically. As such it is more flexible to give the implementations the room to throw different errors.

tests/lib/mirrors/reflected_type_generics_test.dart
```diff
@@ -0,0 +1,63 @@
...
+  expectReflectedType(reflectType(A, [P]), new A<P>().runtimeType);
+  expectReflectedType(reflectType(C, [B, P]), new C<B, P>().runtimeType);
+  expectReflectedType(reflectType(D, [P]), new D<P>().runtimeType);
+  expectReflectedType(reflectType(E, [P]), new E<P>().runtimeType);
+
+  // Edge cases:
+  Expect.throws(
+    () => reflectType(P, []),
+    (e) => e is ArgumentError && e.invalidValue is List,
+    "Should throw an ArgumentError if reflecting not a generic class with empty list of type arguments");
```
floitschG: As written above: I wouldn't test that it really returns an argument error.

Since you already wrote the tests, keep them.

Long lines: 31, and 35.

floitschG:
LGTM to me.
I will merge this PR as soon as @rmacnak-google oks the C++ code, and the final comments are addressed.
Thanks.

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
                            Object::null_instance());
 }
 
-
 DEFINE_NATIVE_ENTRY(Mirrors_makeLocalTypeMirror, 1) {
   GET_NON_NULL_NATIVE_ARGUMENT(AbstractType, type, arguments->NativeArgAt(0));
   return CreateTypeMirror(type);
 }
```
rmacnak-google: Two blank lines between functions.

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
 
+DEFINE_NATIVE_ENTRY(Mirrors_makeLocalTypeMirrorWithTypeArguments, 2) {
+  GET_NON_NULL_NATIVE_ARGUMENT(AbstractType, type, arguments->NativeArgAt(0));
+  GET_NON_NULL_NATIVE_ARGUMENT(Array, args, arguments->NativeArgAt(1));
+
+  const Class& clz = Class::Handle(type.type_class());
```
rmacnak-google: I don't think this will for function types, type parameters or ref types.

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+  if (!clz.IsGeneric()) {
+    const Array& errorArgs = Array::Handle(Array::New(3));
```
rmacnak-google: error_args

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+    errorArgs.SetAt(0, type);
+    errorArgs.SetAt(1, String::Handle(String::New("key")));
+    errorArgs.SetAt(2, String::Handle(String::New("Type must be a generic class.")));
+    Exceptions::ThrowByType(Exceptions::kArgumentValue, errorArgs);
+    UNREACHABLE();
+  }
+  if (clz.NumTypeParameters() != args.Length()) {
+    const Array& errorArgs = Array::Handle(Array::New(3));
```
rmacnak-google: error_args

runtime/lib/mirrors.cc
```diff
@@ -835,12 +835,59 @@ DEFINE_NATIVE_ENTRY(Mirrors_makeLocalClassMirror, 1) {
...
+    }
+    type_arg ^= args.At(i);
+    type_args_obj.SetTypeAt(i, type_arg);
+  }
+
+  Type& instantiated_type = Type::Handle(
+      Type::New(clz, type_args_obj, TokenPosition::kNoSource));
+  instantiated_type ^= ClassFinalizer::FinalizeType(
+      clz, instantiated_type, ClassFinalizer::kCanonicalize);
+  return CreateTypeMirror(instantiated_type);
```
rmacnak-google: Why not just return the type without creating a mirror? See comment on _resolveType.

runtime/lib/mirrors_impl.dart
```diff
@@ -1676,4 +1681,20 @@ class _Mirrors {
     }
     return typeMirror;
   }
+
+  static Type _resolveType(Type key, List<Type> typeArguments) {
```
rmacnak-google: _resolveType -> _instantiateType

runtime/lib/mirrors_impl.dart
```diff
@@ -1676,4 +1681,20 @@ class _Mirrors {
...
+    if (typeArguments.isEmpty) {
+      throw new ArgumentError.value(
+        typeArguments, 'typeArguments', 'Type arguments list can not be empty.'
+      );
+    }
+    var resolvedTypeMirror = makeLocalTypeMirrorWithTypeArguments(
+      key, typeArguments.toList(growable: false));
+    var actualType = resolvedTypeMirror.reflectedType;
+    // We check if the actual type has been cached already and return cached copy.
+    if (_instantiationCache[actualType] == null) {
```
rmacnak-google: This caching will happen anyway in reflectType, so why just return the type directly from the native?

tests/lib/mirrors/reflected_type_generics_test.dart
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS file
```
rmacnak-google: 2016

runtime/lib/mirrors.cc
rmacnak-google: Consider

class T<G> {}
typedef bool Predicate<T>(T arg);
class FBounded<S extends FBounded> {}

main() {
   reflectType(reflectType(T).typeVariables[0].reflectedType, [int]);
   reflectType(Predicate, [int]);
   reflectType(reflectType(FBounded).typeVariables[0].upperBound.reflectedType, [int]);
}

runtime/lib/mirrors_impl.dart
rmacnak-google: can not -> cannot

sdk/lib/mirrors/mirrors.dart
pulyaevskiy: Thanks, this is very helpful! One question: would it be OK if I replace `type parameters` with `type variables` since that what's being used in Dart code?

I might need to update it to include generic typedefs as well (per @rmacnak-google  comments)

Unrelated, but is it supposed to be like this having "type parameters" in C and "type variables" in Dart referring to the same thing? It seems like unnecessary ambiguity to me, but maybe I'm not seeing the whole picture...

runtime/lib/mirrors_impl.dart
pulyaevskiy: This might conflict with @rmacnak-google comments on reflectType.
I think I like his idea by just instantiating generic type in this if statement and leaving everything else as it was before (sorry @rmacnak-google I didn't realize this in your very first comment about `_instantiateType`).

@floitschG would it be ok to go with `_instantiateType` approach for you?

tests/lib/mirrors/reflected_type_generics_test.dart
pulyaevskiy: @floitschG @rmacnak-google  I've been trying to wrap my head around error handling in such cases.

Could you help me in understanding "best practices" here?

Basically, I'm currently using pretty verbose way to handle argument errors, e.g.:

```cpp
  if (!clz.IsGeneric()) {
    const Array& error_args = Array::Handle(Array::New(3));
    error_args.SetAt(0, type);
    error_args.SetAt(1, String::Handle(String::New("key")));
    error_args.SetAt(2, String::Handle(
      String::New("Type must be a generic class.")));
    Exceptions::ThrowByType(Exceptions::kArgumentValue, error_args);
    UNREACHABLE();
  }
```

That's why I put these strict assertions here so that I can actually verify my code.
However as @floitschG pointed out this is undesirable and we only need to know that an `Error` happened.

Now, I realize that I can replace those verbose `if` statements with a call to `ASSERT`, so that it would become:

```cpp
ASSERT(clz.IsGeneric());
// or:
ASSERT(clz.NumTypeParameters() != args.Length());
```

This would throw a `RuntimeError`. I see that this approach is used a lot throughout the codebase, while constructing exceptions like I do currently is fairly rare.

Would you say that `ASSERT` is preferable in such cases?

I feel like providing `ArgumentError` with detailed context is more user-friendly, but on the other hand it's really verbose and maybe provided value is not justified in terms of maintenance cost.

runtime/lib/mirrors.cc
pulyaevskiy: I'm assuming 1st and 3rd lines should throw an error and the 2nd line should return mirror of `Predicate<int>`?

I'm working on it and it works almost as expected, with couple issues:

* I can't really construct `Predicate<int>` as I do with normal classes using `new A().runtimeType`. So I haven't figured yet how to write this assertion.
* Looks like bounded types are not being checked atm. I'm figuring out how to fix this.

runtime/lib/mirrors_impl.dart
floitschG: Fine with me.
I wanted to suggest something similar, but in the end it looked more difficult than necessary. Maybe I missed something.

runtime/lib/mirrors.cc
rmacnak-google: Two ways to construct `Predicate<int>`:

```
class C {
  Predicate<int> field;
}
reflectClass(C).declarations[#field].type.reflectedType;
```

```
class C<T> {
  get t => T;
}
new C<Predicate<int>>().t
```

tests/lib/mirrors/reflected_type_generics_test.dart
rmacnak-google: ASSERTs only run if the VM is compiled in debug mode, and if they fail they crash the VM.
GET_NON_NULL_NATIVE_ARGUMENT will throw an exception if the argument has the wrong type, and Exceptions::ThrowByType can be used to throw exceptions for other cases.

If think a specific exception is fine, Florian is only saying that the documentation should not promise you'll get any particular exception.

runtime/lib/mirrors.cc
pulyaevskiy: @rmacnak-google Thanks! I found the second option myself, but didn't realize there is one more. :)

I've also spent some time with bounded types and I couldn't really get the test case to fail. I can actually get `FBounded<int>` from `reflectType(FBounded, [int])`. 
I see that `ClassFinalizer::FinalizeType()` performs `CheckBounds` but not sure why it doesn't throw an error for me yet. 

Will look into it more tonight, but if you have any quick hints it would help a lot!

pulyaevskiy:
@floitschG @rmacnak-google 

I just pushed an update which addresses your latest comments. 
I've also added a couple test cases for generic mixins.
Everything looks good on the VM side now, but JS tests are failing at this point:

* Looks like reflection of generic typedefs is not implemented in JS mirrors.
* Some strange behavior with mixins, it seems to me like a bug in JS implementation which complains about:

  ```
  class Composite<K, V> extends Object with Mixin<K> {}
                                            ^^^^^^^^
  tests/lib/mirrors/reflected_type_generics_test.dart:
  ```
  
  However this is perfectly valid statement.

I'm not sure if there is a way to write "platform" specific tests for the SDK, seems like a bad idea.

What would be the best course of action to address this issue?

pulyaevskiy:
Also, just wanted to throw this out there.
I've been thinking that I really like this new `_instantiateType` function and I feel like it almost deserves it's own place in public API.

It doesn't even have to be associated with Mirrors library at all, since it just allows you to dynamically construct a generic type with specific arguments. And there are quite a few use cases which would benefit from such a feature.
Moreover, it would allow to avoid any changes to `reflectType` itself, e.g.:

```dart
class Repository<T> {}
class User {}
var mirror = reflectType(instantiateType(Repository, [User]));
```

Though this would not solve the issues with JS tests and they would still have to be addressed somehow.

Anyways, just thought it deserves mentioning here.

tests/lib/mirrors/reflected_type_generics_test.dart
```diff
@@ -0,0 +1,98 @@
...
+
+class Composite<K, V> extends Object with Mixin<K> {}
+
+main() {
+  // "Happy" paths:
+  expectReflectedType(reflectType(A, [P]), new A<P>().runtimeType);
+  expectReflectedType(reflectType(C, [B, P]), new C<B, P>().runtimeType);
+  expectReflectedType(reflectType(D, [P]), new D<P>().runtimeType);
+  expectReflectedType(reflectType(E, [P]), new E<P>().runtimeType);
+  // expectReflectedType(
```
rmacnak-google: Uncomment.

pulyaevskiy: Sorry, fixed: ff17c80

rmacnak-google:
VM changes LGTM.

Note that you can separate out cases that don't pass on all implementations into and mark them as failing in the status file, which for mirrors is in tests/lib/lib.status. You'll notice that dart2js still fails a significant fraction of the test suite.

pulyaevskiy:
@rmacnak-google Thanks, that's what I was looking for. I've updated my tests according to the documentation.

The build is failing though. It seems like it's failing on `master` as well.
When I run my tests locally I get success for both vm and dart2js.

One more thing I forgot to mention: the `Type::IsMalbounded()` would only report an issue if `--checked` mode is enabled, so in production mode doing `reflectType(FBounded, [int])` would not throw an error.

I suppose such errors should be caught during development and in tests. However it should be possible to make it more strict. Let me know your thoughts!

floitschG:
Thanks.

I tried the patch, and from what I can see, the test is still failing when running with the VM in unchecked mode.
If the test is only supposed to fail in checked mode, guard it in the test:
``` dart
bool isChecked = false;
assert((isChecked = true));
if (isChecked) {
  // do test.
}
```

If it should always work, but we still have an acceptable bug, then add it to the status lines under a `$checked` section.
@whesse: what would be the correct header for this?
`[ $compiler == none && $unchecked ]` ?

Looks like we are getting really close, though.

whesse:
I think $checked == false.  I don't think ! Works for not, only !=.

floitschG:
I have seen `$unchecked` in status files, so that should be fine. It's more about the 'runs on the VM' where I'm not sure.

pulyaevskiy:
Thanks @floitschG, @whesse !

I used `[ $compiler == none && $runtime == vm && $unchecked ]`.
Just pushed this update. Looks like Travis build failed due to a network issue (couldn't install Dart SDK). Not sure if there is a way to restart it.

Tests are now passing for me locally in both checked and unchecked mode.

pulyaevskiy:
@floitschG @rmacnak-google 
Wondering if someone can help and hit "Restart build" button in Travis for the latest build?
(I know I can do this for my own repos, but don't see this button for the SDK repo)

The build failed on trying to download the SDK:

```
unzip:  cannot find zipfile directory in one of /home/travis/dartsdk.zip or
        /home/travis/dartsdk.zip.zip, and cannot find /home/travis/dartsdk.zip.ZIP, period.
The command "unzip $HOME/dartsdk.zip -d $HOME > /dev/null" failed and exited with 9 during .
```

floitschG:
This CL seems to have broken `typearguments_mirror_test` for dart2js:
```
python tools/test.py -cdart2js -rd8 -mrelease -t60 --packages /mnt/ssd/floitsch/code/dart/sdk/.packages lib/mirrors/typearguments_mirror_test

python tools/test.py -cdart2js --checked -rd8 -mrelease -t60 --packages /mnt/ssd/floitsch/code/dart/sdk/.packages lib/mirrors/typearguments_mirror_test
```

I think just testing for `vm` is a bad idea since it won't trigger for drt or dartium which are also running the VM. I would go with ` $compiler == none `
We generally don't use `SkipByDesign`, unless the test takes too long. Maybe just write `Fail, Ok`.

After that we should be good to go :)

pulyaevskiy:
Thanks @floitschG !

I was able to reproduce errors in `typearguments_mirror_test` and working on a fix.
Will also try to update `lib.status` with just `$compiler == none` and see if it works.

floitschG:
Any updates?
Do you manage, or should I have a look?

pulyaevskiy:
Sorry for being silent.
I think I can submit an update later today. 
I've been trying to run the whole test suite in multiple runtimes and compilers to double check that everything is working as expected and found a bug in JS implementation. Fix is on the way.

mit-mit:
Hi @pulyaevskiy, any updates?

mit-mit:
Hi @pulyaevskiy, any news? If not, we will unfortunately have to close this pull request.

pulyaevskiy:
Hi, sorry for the delay! And Happy New Year to everyone!

I've just pushed latest changes, I was able to find a bug in handling nested generic types in parsing of type arguments. There was a piece of code already doing it properly so I had to extract it into it's own method.

I've tried to run all the tests but it's giving me inconsistent results, I blame my laptop is not that powerful so it just fails on random test cases.

floitschG:
Looks great. Thanks.
(And we wouldn't have discarded it. I would have taken the necessary steps to finish it)

mit-mit:
Thanks for contributing @pulyaevskiy !

asgerf:
I had to revert the commit because some tests still fail:
dart2js/mirrors_used_test
pkg/front_end/test/src/async_dependency_walker_test

asgerf:
@floitschG could you try to reland this when possible?

floitschG:
Already have a CL up.
Siggi is ooo, so this might only land at the end of the week.

floitschG:
Here is the CL:
https://codereview.chromium.org/2615943004/

Basically, I just remove a check that is likely to fail, when `MirrorsUsed` are used.
It's not perfect, but other solutions would require *much* more work.

pulyaevskiy:
@floitschG thanks a lot for taking care of this!

pulyaevskiy:
Just noticed this was released as part of 1.22! Thanks everyone for you help, especially @floitschG and @rmacnak-google for reviewing my code and working to get this finalized.

marcotas:
This is not working for me. This is what I'm trying to do:
```dart
class ClassA<T> {...}

class ClassB {...}
```
How can I make this code work?
```dart
// this doesn't work. It keeps returning ClassA<dynamic> instead.
ClassA<ClassB> instance = reflectClass(ClassA<ClassB>).newInstance(...).reflectee; 
```
I'm using `Dart SDK version: 2.17.1 (stable) (Tue May 17 17:58:21 2022 +0000) on "macos_arm64"`

marcotas:
It works perfectly! Thank you so much @mraleph! 🚀

PR #26193, opened 2016-04-05
Minor change to fix some strong-mode errors

stevemessick:
@bwilkerson @scheglov 
Subclasses that override `element` generated an error in strong-mode because the field could not be overridden. Changing it o a getter eliminates the error. All 3540 tests pass.


bwilkerson:
lgtm

scheglov:
LGTM

PR #26268, opened 2016-04-14
JS: Use direct function invocation when possible for calls with 4 or 5 args

greglittlefield-wf:
## Ultimate Problem

A library I'm working with ([react-dart](https://github.com/cleandart/react-dart)) involves frequent calls to Dart functions from JS code via `allowInteropCaptureThis`, and many of them take 4 arguments (for a total arity of 5 when you count the JS `this`).

Since these calls take too many arguments, they fall back to the generic function calling, which makes use of the JS `Function.prototype.apply`, which isn't as fast.
## Changes
- Directly invoke functions, when possible, for arities up to 5, as opposed to up to 3.
  - I think this threshold is more reasonable, but please say otherwise if it's not.
- Add test for Function.apply called with 0-5 arguments.


greglittlefield-wf:
Is signed it! (My company, Workiva, signed it!)

jacob314:
LGTM

PR #26294, opened 2016-04-19
Actualize .status files according to the latest co19 tests

sgrekhov:
Tests that were fixed/removed removed from .status files. Tests that fail added


whesse:
It would be good if this pull request also updated the co19_rev hash in DEPS and tools/deps/dartium.deps/DEPS to the correct revision.  I used
 "co19_rev": "@3ed795ea02e022ef19c77cf1b6095b7c8f5584d0",
which is a revision from Apr 20, the same date as this push request.

I think that tests/co19/co19_dart2analyzer.status needs to be updated too.  I ran
tools/test.py -mrelease -cdart2analyzer -rnone co19, and got 319 failures.  A log of the required updates is attached:
[dart2analyzerUpdates.txt](https://github.com/dart-lang/sdk/files/242424/dart2analyzerUpdates.txt)

sgrekhov:
@whesse it was a good idea to check analyzer as well. Thank you. I fixed tests (see https://github.com/dart-lang/co19/commit/f027b7ab540b287f444859d20638801afaeb7ff9 ) to fix analyzer failures. Updated co19_dart2analyzer.status added to this PR and tools/deps/dartium.deps/DEPS updated up to the latest revision of co19 tests. Please review

whesse:
It would be good if this had both the DEPS file at the root, and the DEPS file in tools/deps/dartium.deps updated with the co19 revision the status patch applies to.

Some tests that fail with this patch have been fixed in co19 since then.  My current update of this patch, that I will try to reland (I landed, and reverted), updates the DEPS to co19 revision 3f0a4bc9a080a792cdf5f093147a900f99ea301f

Since my CL has diverged from this push request, don't make any more changes to this one.

mit-mit:
@whesse is it correctly understood that you will address this issue, and that we an close this bug?

whesse:
Yes, I just can't do it during the day.  This should happen within a few days.

PR #26403, opened 2016-05-04
Removing overriden field from _HttpInboundMessage.

alexeieleusis:
Since the field is accessible from the derived class, there
is no need to override it.
Found via dart-lang/linter


alexeieleusis:
cc: @bwilkerson @pq

floitschG:
I have this change already pending in one of my CLs: https://chromiumcodereview.appspot.com/1904553006/diff/100001/sdk/lib/io/http_impl.dart

Will close this issue (since I will land my CL soon), but thanks for the PR.

alexeieleusis:
:)

PR #26985, opened 2016-07-29
Added js import to sample code.

DanTup:
This import is required for the sample code to pass the analyzer, otherwise you'll get an error on the first line.

```
[error] Annotation can be only constant variable or constant constructor invocation
```


mit-mit:
@jacob314 can you review, please?

mit-mit:
@jacob314 any update on this?

PR #27093, opened 2016-08-17
Updated directory_linux.cc to conform with the deprecation of readdir_r

starfys:
Since glibc 2.24, readdir_r is deprecated. According to [the man page](http://man7.org/linux/man-pages/man3/readdir.3.html):

> In the current POSIX.1 specification (POSIX.1-2008), readdir(3) is not required to be thread-safe.  However, in modern implementations (including the glibc implementation), concurrent calls to readdir(3) that specify different directory streams are thread-safe.  Therefore, the use of readdir_r() is generally unnecessary in multithreaded programs.  In cases where multiple threads must read from the same directory stream, using readdir(3) with external synchronization is still preferable to the use of readdir_r(), for the reasons given in the points above.

### DIFF
--- a/runtime/bin/directory_linux.cc
+++ b/runtime/bin/directory_linux.cc
@@ -120,7 +120,7 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
   dirent* result;
   errno = 0;
   result = readdir(reinterpret_cast<DIR*>(lister_));
-  if (result != NULL) {
+  if ((errno == 0) && (result != NULL)) {
     if (!listing->path_buffer().Add(result->d_name)) {
       done_ = true;
       return kListError;
--- a/runtime/bin/directory_linux.cc
+++ b/runtime/bin/directory_linux.cc
@@ -117,20 +117,19 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
 
   // Iterate the directory and post the directories and files to the
   // ports.
-  int status = 0;
-  dirent entry;
   dirent* result;
-  status = NO_RETRY_EXPECTED(readdir_r(
-      reinterpret_cast<DIR*>(lister_), &entry, &result));
-  if ((status == 0) && (result != NULL)) {
-    if (!listing->path_buffer().Add(entry.d_name)) {
+  errno = 0;
+  result = readdir(
+      reinterpret_cast<DIR*>(lister_));
+  if (result != NULL) {
+    if (!listing->path_buffer().Add(result->d_name)) {
       done_ = true;
       return kListError;
     }
-    switch (entry.d_type) {
+    switch (result->d_type) {
       case DT_DIR:
-        if ((strcmp(entry.d_name, ".") == 0) ||
-            (strcmp(entry.d_name, "..") == 0)) {
+        if ((strcmp(result->d_name, ".") == 0) ||
+            (strcmp(result->d_name, "..") == 0)) {
           return Next(listing);
         }
         return kListDirectory;
@@ -178,16 +177,16 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
             // Recurse into the subdirectory with current_link added to the
             // linked list of seen file system links.
             link_ = new LinkList(current_link);
-            if ((strcmp(entry.d_name, ".") == 0) ||
-                (strcmp(entry.d_name, "..") == 0)) {
+            if ((strcmp(result->d_name, ".") == 0) ||
+                (strcmp(result->d_name, "..") == 0)) {
               return Next(listing);
             }
             return kListDirectory;
           }
         }
         if (S_ISDIR(entry_info.st_mode)) {
-          if ((strcmp(entry.d_name, ".") == 0) ||
-              (strcmp(entry.d_name, "..") == 0)) {
+          if ((strcmp(result->d_name, ".") == 0) ||
+              (strcmp(result->d_name, "..") == 0)) {
             return Next(listing);
           }
           return kListDirectory;
@@ -204,8 +203,7 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
   }
   done_ = true;
 
-  if (status != 0) {
-    errno = status;
+  if (errno != 0) {
     return kListError;
   }
 
@@ -277,28 +275,27 @@ static bool DeleteRecursively(PathBuffer* path) {
 
   // Iterate the directory and delete all files and directories.
   int path_length = path->length();
-  dirent entry;
   dirent* result;
-  while (NO_RETRY_EXPECTED(readdir_r(dir_pointer, &entry, &result)) == 0) {
+  while ((result = readdir(dir_pointer)) != NULL) {
     if (result == NULL) {
       // End of directory.
       return (NO_RETRY_EXPECTED(closedir(dir_pointer)) == 0) &&
              (NO_RETRY_EXPECTED(remove(path->AsString())) == 0);
     }
     bool ok = false;
-    switch (entry.d_type) {
+    switch (result->d_type) {
       case DT_DIR:
-        ok = DeleteDir(entry.d_name, path);
+        ok = DeleteDir(result->d_name, path);
         break;
       case DT_REG:
       case DT_LNK:
         // Treat all links as files. This will delete the link which
         // is what we want no matter if the link target is a file or a
         // directory.
-        ok = DeleteFile(entry.d_name, path);
+        ok = DeleteFile(result->d_name, path);
         break;
       case DT_UNKNOWN: {
-        if (!path->Add(entry.d_name)) {
+        if (!path->Add(result->d_name)) {
           break;
         }
         // On some file systems the entry type is not determined by
@@ -310,12 +307,12 @@ static bool DeleteRecursively(PathBuffer* path) {
         }
         path->Reset(path_length);
         if (S_ISDIR(entry_info.st_mode)) {
-          ok = DeleteDir(entry.d_name, path);
+          ok = DeleteDir(result->d_name, path);
         } else if (S_ISREG(entry_info.st_mode) || S_ISLNK(entry_info.st_mode)) {
           // Treat links as files. This will delete the link which is
           // what we want no matter if the link target is a file or a
           // directory.
-          ok = DeleteFile(entry.d_name, path);
+          ok = DeleteFile(result->d_name, path);
         }
         break;
       }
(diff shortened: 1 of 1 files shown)


starfys:
I signed it!

runtime/bin/directory_linux.cc
```diff
@@ -117,20 +117,19 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
...
-  int status = 0;
-  dirent entry;
   dirent* result;
-  status = NO_RETRY_EXPECTED(readdir_r(
-      reinterpret_cast<DIR*>(lister_), &entry, &result));
-  if ((status == 0) && (result != NULL)) {
-    if (!listing->path_buffer().Add(entry.d_name)) {
+  errno = 0;
+  result = readdir(
+      reinterpret_cast<DIR*>(lister_));
```
floitschG: This should now fit on one line.

runtime/bin/directory_linux.cc
```diff
@@ -277,28 +275,27 @@ static bool DeleteRecursively(PathBuffer* path) {
 
   // Iterate the directory and delete all files and directories.
   int path_length = path->length();
-  dirent entry;
   dirent* result;
-  while (NO_RETRY_EXPECTED(readdir_r(dir_pointer, &entry, &result)) == 0) {
+  while ((result = readdir(dir_pointer)) != NULL) {
```
floitschG: This doesn't do the same thing anymore: it doesn't closedir and remove since it will never enter line 282.

What about:
```
while (true) {
  dirent* result = readdir(dir_pointer);
  if (result == NULL) { ...
```

floitschG:
Thanks. I added a few comments.
@mkustermann : I'm not that familiar with Linux APIs. Could you also have a short look?

starfys:
Made the requested changes. Considering the original loop uses the return value of readdir_r, and readdir writes to errno instead of returning a status code, should I organize the loop around this?

johnmccutchan:
@zanderso our dart:io maintainer

zanderso:
Is readdir thread safe going as far back as glibc 2.11?

starfys:
[readdir.c from 2.24](https://sourceware.org/git/gitweb.cgi?p=glibc.git;a=blob;f=sysdeps/posix/readdir.c;h=e1e63f6157f057fef5d4556413c4f86efa4321c0;hb=HEAD)
[readdir.c from 2.11](https://sourceware.org/git/gitweb.cgi?p=glibc.git;a=blob;f=sysdeps/unix/readdir.c;hb=7f9cd6ed20b12435e68153870cee5adbe09c4bca)
Ran diff on the two and got
```diff
diff --git a/readdir-master.c b/readdir-2.11.c
index e1e63f6..13e5e9a 100644
--- a/readdir-master.c
+++ b/readdir-2.11.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 1991-2016 Free Software Foundation, Inc.
+/* Copyright (C) 1991-1997,1999,2000,2002,2007 Free Software Foundation, Inc.
    This file is part of the GNU C Library.
 
    The GNU C Library is free software; you can redistribute it and/or
@@ -12,8 +12,9 @@
    Lesser General Public License for more details.
 
    You should have received a copy of the GNU Lesser General Public
-   License along with the GNU C Library; if not, see
-   <http://www.gnu.org/licenses/>.  */
+   License along with the GNU C Library; if not, write to the Free
+   Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
+   02111-1307 USA.  */
 
 #include <errno.h>
 #include <limits.h>
@@ -40,7 +41,7 @@ __READDIR (DIR *dirp)
   DIRENT_TYPE *dp;
   int saved_errno = errno;
 
-#if IS_IN (libc)
+#ifndef NOT_IN_libc
   __libc_lock_lock (dirp->lock);
 #endif
 
@@ -110,7 +111,7 @@ __READDIR (DIR *dirp)
       /* Skip deleted files.  */
     } while (dp->d_ino == 0);
 
-#if IS_IN (libc)
+#ifndef NOT_IN_libc
   __libc_lock_unlock (dirp->lock);
 #endif
```
There seems to be a bugfix and some compiler flag changes between them, but there seems to be no differences in terms of thread safety.

runtime/bin/directory_linux.cc
```diff
@@ -117,20 +117,18 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
...
-  int status = 0;
-  dirent entry;
   dirent* result;
-  status = NO_RETRY_EXPECTED(readdir_r(
-      reinterpret_cast<DIR*>(lister_), &entry, &result));
-  if ((status == 0) && (result != NULL)) {
-    if (!listing->path_buffer().Add(entry.d_name)) {
+  errno = 0;
+  result = readdir(reinterpret_cast<DIR*>(lister_));
+  if ((errno == 0) && (result != NULL)) {
```
mkustermann: `result != NULL` implies `errno == 0` according to the man page.

Only if `result == NULL` one needs to distinguish between error (in which case `errno != 0`) and end of stream (in which case `errno == 0`).

runtime/bin/directory_linux.cc
```diff
@@ -117,20 +117,18 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
...
   // ports.
-  int status = 0;
-  dirent entry;
   dirent* result;
-  status = NO_RETRY_EXPECTED(readdir_r(
-      reinterpret_cast<DIR*>(lister_), &entry, &result));
-  if ((status == 0) && (result != NULL)) {
-    if (!listing->path_buffer().Add(entry.d_name)) {
+  errno = 0;
+  result = readdir(reinterpret_cast<DIR*>(lister_));
```
mkustermann: Maybe keep the old name `entry` or `dentry` instead of just `result`?

runtime/bin/directory_linux.cc
```diff
@@ -277,28 +274,27 @@ static bool DeleteRecursively(PathBuffer* path) {
 
   // Iterate the directory and delete all files and directories.
   int path_length = path->length();
-  dirent entry;
-  dirent* result;
-  while (NO_RETRY_EXPECTED(readdir_r(dir_pointer, &entry, &result)) == 0) {
+  while (true) {
+    dirent* result = readdir(dir_pointer);
     if (result == NULL) {
```
mkustermann: If `readdir` returns `NULL` it might be end-of-stream **or** error. Why not implement the same error handling as above?

starfys: Would something like this work?
```c
errno=0
dirent* entry = readdir(dir_pointer);
while(errno == 0) {
    if (entry == NULL) {
      // End of directory.
      return (NO_RETRY_EXPECTED(closedir(dir_pointer)) == 0) &&
             (NO_RETRY_EXPECTED(remove(path->AsString())) == 0);
    }
//Insert rest of loop here
errno = 0
entry = readdir(dir_pointer);
}
```

runtime/bin/directory_linux.cc
```diff
@@ -117,20 +117,18 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
...
   // Iterate the directory and post the directories and files to the
   // ports.
-  int status = 0;
-  dirent entry;
-  dirent* result;
-  status = NO_RETRY_EXPECTED(readdir_r(
-      reinterpret_cast<DIR*>(lister_), &entry, &result));
-  if ((status == 0) && (result != NULL)) {
-    if (!listing->path_buffer().Add(entry.d_name)) {
+  dirent* entry;
```
mkustermann: nit: Just move the variable declaration down:

```c++
dirent* entry = readdir(...);
```

mkustermann:
@zanderso Regarding thread safety: Do you agree that our code should ensure we don't call `readdir/readdir_r` on two different threads at the same point in time **on the same directory stream** (`DIR*`)?

There is also a comment on lwn about this: http://lwn.net/Articles/696475/

zanderso:
Yes, we should be able to ensure that.

runtime/bin/directory_linux.cc
```diff
@@ -277,28 +273,32 @@ static bool DeleteRecursively(PathBuffer* path) {
...
-  dirent* result;
-  while (NO_RETRY_EXPECTED(readdir_r(dir_pointer, &entry, &result)) == 0) {
-    if (result == NULL) {
+  while (true){
+    // In case `readdir()` returns `NULL` we distinguish between end-of-stream and error
+    // by looking if `errno` was updated.
+    errno = 0;
+    dirent* entry = readdir(dir_pointer);
+    if (entry == NULL) {
+      // Failed to read next directory entry
```
floitschG: Nit: finish comments with ".".

runtime/bin/directory_linux.cc
```diff
@@ -277,28 +273,32 @@ static bool DeleteRecursively(PathBuffer* path) {
...
   int path_length = path->length();
-  dirent entry;
-  dirent* result;
-  while (NO_RETRY_EXPECTED(readdir_r(dir_pointer, &entry, &result)) == 0) {
-    if (result == NULL) {
+  while (true){
+    // In case `readdir()` returns `NULL` we distinguish between end-of-stream and error
+    // by looking if `errno` was updated.
+    errno = 0;
+    dirent* entry = readdir(dir_pointer);
```
zanderso: Some summary of the discussion in the pull request should probably go in a comment here.

runtime/bin/directory_linux.cc
```diff
@@ -277,28 +273,32 @@ static bool DeleteRecursively(PathBuffer* path) {
...
+    if (entry == NULL) {
+      // Failed to read next directory entry.
+      if (errno != 0) break;
```
zanderso: Please use curly braces:
if (errno != 0) {
  break;
}

zanderso:
lgtm with nits

runtime/bin/directory_linux.cc
```diff
@@ -277,28 +273,43 @@ static bool DeleteRecursively(PathBuffer* path) {
...
   int path_length = path->length();
-  dirent entry;
-  dirent* result;
-  while (NO_RETRY_EXPECTED(readdir_r(dir_pointer, &entry, &result)) == 0) {
-    if (result == NULL) {
+  while (true){
+    // In case `readdir()` returns `NULL` we distinguish between end-of-stream and error
+    // by looking if `errno` was updated.
+    errno = 0;
+    //In glibc 2.24+, readdir_r is deprecated.
```
zanderso: Please add spaces after //

zanderso:
Thanks!

PR #27171, opened 2016-08-26
Remove a duplicated angle bracket.

DanTup:
This doesn't look valid, seems like a typo to me.

### DIFF
--- a/pkg/analysis_server/tool/spec/spec_input.html
+++ b/pkg/analysis_server/tool/spec/spec_input.html
@@ -422,7 +422,7 @@
           <field name="sources">
             <map>
               <key><ref>String</ref></key>
-              <value><list>><ref>String</ref></list></value>
+              <value><list><ref>String</ref></list></value>
             </map>
             <p>
               A mapping from source URIs to directly reachable source URIs. For example,


bwilkerson:
lgtm

Did you try re-generating the spec afterward to ensure that there are no unexpected changes?

DanTup:
Sorry, I did not - I thought it was only used to generate `api.html` and that didn't seem to have any issue in this area.

If you can confirm what I need to do, I'm happy to do it :-)

bwilkerson:
The script that needs to be run is in analysis_server/tool/spec/generate_files. (Not sure whether it will run as-is on Windows.)

DanTup:
Heh, it's a bash script! ;(

Windows 10 apparently can run Bash now, but I haven't set it up yet. I can give it a go over the weekend if you'd like, but if it's more convenient you could just abort this PR and change/test it when suits you. Sorry! ;(

bwilkerson:
I ran the script locally, and there are no changes (I was fairly sure that would be the case). So I'm merging this request.

DanTup:
Thanks!

DanTup:
FWIW, I did try to run the script on Bash for Windows... First time it failed because the script contained `\r`'s (I used a clone I did on Windows, whoops!) but when I re-cloned from within Bash and tried again this happened:

`No valid dart configuration found in /mnt/m/Coding/TestStuff/sdk/sdk/bin/../../out/`

Probably not a big deal since you're already sorted this and it's unlikely I'll need to run it in future but I thought it was worth mentioning in case this was expected to work on a clean clone without other setup. If other setup is expected, then this might be by design!

PR #27565, opened 2016-10-11
Fix analysis errors in analysis_server and analyzer packages tests

ddrone:
Apparently `path` package started exporting `equals` function at some point, which makes `equals` ambiguous in a bunch of test files. This CL fixes it.

### DIFF
--- a/pkg/analysis_server/test/context_manager_test.dart
+++ b/pkg/analysis_server/test/context_manager_test.dart
@@ -24,7 +24,7 @@ import 'package:analyzer/src/task/options.dart'
 import 'package:analyzer/src/util/glob.dart';
 import 'package:linter/src/plugin/linter_plugin.dart';
 import 'package:linter/src/rules/avoid_as.dart';
-import 'package:path/path.dart';
+import 'package:path/path.dart' hide equals;
 import 'package:plugin/manager.dart';
 import 'package:plugin/plugin.dart';
 import 'package:test/test.dart';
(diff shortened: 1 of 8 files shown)


kevmoo:
cc @bwilkerson @scheglov

bwilkerson:
Sorry, but I'm not understanding the need for this change. The `equals` function we're using is the one defined in `core_matchers.dart`, which is the one that `test` exports. Where is the ambiguity that you're seeing?

ddrone:
`path` package exposes `equals` function (https://www.dartdocs.org/documentation/path/1.4.0/path/equals.html), which gets imported here, which results to `equals` being reported as ambiguous by analyzer.

ddrone:
Basically, I've opened `analyzer` and `analysis_server` in IntelliJ as Dart projects and seen a bunch of errors reported in "Dart Analysis" window. This CL makes these errors disappear

bwilkerson:
@ddrone Thanks, I understand now. I wasn't seeing problems yet because we haven't pulled the latest version of `path` into the SDK yet.

@nex3 I expect that we will not be the only team with breakages because of the conflicting definitions of `equals` in `test` and `equals` in `path`. Is it possible that we could fix the problem rather than forcing everyone to add hide combinators or prefixes?

ddrone:
@bwilkerson so what's the right setup for browsing SDK packages code? I like to do it in IntelliJ, so all the navigation stuff works, but it runs `pub get` and apparently picks wrong package versions.

bwilkerson:
There is (now) a `.packages` file in the root of the SDK, so there is no need to run pub. If you remove the `.packages` file (and `packages` directory if it exists) that pub created, then everything should work fine.

nex3:
The `path` package has been clear that it was intended to be imported with a prefix—I don't think it makes sense to change the API to accommodate the very few users who don't do that.

bwilkerson:
I suppose then that the better change is to add a prefix to all imports of the path package in order to guard against future breakage. The other disadvantage of this PR is that it would introduce hints for anyone not yet on 1.4.0. For those reasons I think I'd rather not have this committed.

I do appreciate the contribution, though! Thanks!

PR #27690, opened 2016-10-29
Make Element.accept() generic.

alorenzen:


kevmoo:
CC @scheglov @bwilkerson

alorenzen:
Ping?

bwilkerson:
Sorry about the delay, I missed the first e-mail notification.

I really like the change, but we need to update all of the methods that override the accept method to have the same signature, otherwise we get strong mode errors.

alorenzen:
Ok, I've updated all of the subclasses that I could find (based on analyzer warnings in intellij) for the analyzer package.
Please let me know if there are other packages that would need to be updated in parallel.

bwilkerson:
Unfortunately, there are also a couple of places in the 'analysis_server' package, which is in the same repository as the 'analyzer' package and hence must be updated at the same time.

alorenzen:
I looked, but was unable to find any instances in the 'analysis_server' package. There were plenty of places that called accept(), but none that overrode it.

mit-mit:
@bwilkerson can you elaborate?

bwilkerson:
Sorry about the delay. I often miss e-mails like this, so if I haven't responded within a day, please feel free to send me reminder.

I apologize. I don't know what made me think there were implementations in analysis server; I certainly can't find them now. I can only guess that I selected the wrong method when I looked for other implementations of `accept`.

It definitely lgtm.

mit-mit:
Merging.

mit-mit:
@alorenzen thanks so much for your contribution!

PR #27701, opened 2016-10-31
Add a note on the README of js package

a14n:


a14n:
Without _factory constructor_ you face issue with _dart2js_ and _dartium_ .

```
Js-interop method '' has named arguments but is not a factory constructor of an @anonymous @JS class.
```

pkg/js/README.md
```diff
@@ -90,6 +90,8 @@ class Options {
 }
 ```
 
+NB: This _anonymous_ class must have an unnamed _factory constructor_.
```
a14n: _NB_ means [_Nota Bene_](https://en.wikipedia.org/wiki/Nota_bene). I can remove it if it is confusing.

PR #27897, opened 2016-11-25
Remove scary warning from dart:convert#Codec

matanlurey:
Seems unnecessarily scary to have these types of messages in the stable SDK.


kevmoo (approved):
🤘 🐑

PR #27932, opened 2016-11-29
Add a performance benchmark for the angular analyzer plugin

MichaelRFairhurst:
It doesn't look like this is very easy to add to our own repo. Also noticed that flutters benchmark is in here even though they have a separate codebase for their analyzer plugins. That said, if there should indeed be a way to commit this to the angular analyzer repo and that would be preferred, I can definitely look into that.

### DIFF
--- a/pkg/analysis_server/benchmark/perf/benchmark_angular.dart
+++ b/pkg/analysis_server/benchmark/perf/benchmark_angular.dart
@@ -1,0 +1,117 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library server.performance.local;
+
+import 'dart:async';
+
+import 'package:analysis_server/plugin/protocol/protocol.dart';
+
+import 'benchmark_scenario.dart';
+import 'memory_tests.dart';
+
+main(List<String> args) async {
+  int length = args.length;
+  if (length < 1) {
+    print('Usage: dart benchmark_local.dart path_to_np8080 (an example ngdart project)'
+        ' [benchmark_id]');
+    return;
+  }
+  paths = new PathHolder(projectPath: args[0]);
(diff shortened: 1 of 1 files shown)


scheglov:
It seems that I didn't explain it in right words initially. The existing performance tests are not tests for Flutter, they are tests for Analysis Server performance on the Flutter codebase.

scheglov:
The benchmark looks OK, although the one for changing HTML is probably not very useful.

Whether or not to put the benchmark into Analysis Server is another question.
I would prefer not to.
@bwilkerson might also want to see this.

MichaelRFairhurst:
Ah! Thought the arrangement was a little strange.

Is there a way for me to use this infrastructure though from my own project? Since I'm using files in `/test` and `/benchmark` instead of `lib`. I tried using hardcoded relative paths, but that also seemed to have some issues.

bwilkerson:
I might be missing something, but I think this code would run the main in analysis_server's bin, and that main doesn't install the angular2dart plugin, so I don't think you'd be benchmarking your plugin.

> Is there a way for me to use this infrastructure though from my own project?

Unfortunately, no. A package can only import from the 'lib' directory of another package. And then there's the problem that we don't publish the 'analysis_server' package.

MichaelRFairhurst:
> I might be missing something, but I think this code would run the main in analysis_server's bin, and that main doesn't install the angular2dart plugin, so I don't think you'd be benchmarking your plugin.

Yes, it runs the bin. If you have our plugin installed, it will replace the analysis server bin, so it will run our analysis code.

As an extra nicety, this allows be to run these benchmarks with and without my plugin by switching out the bin.

MichaelRFairhurst:
Want to check on what direction we should take with this.

If merging isn't the right option, I can leave it as a branch in my fork for us to use, or maybe we could extract the benchmarking stuff into a library so that I can use it from the angular analysis repo.

Or alternative suggestions welcomed!

bwilkerson (approved):
Given that the benchmark doesn't introduce a dependency on Angular, I have no problems merging it in.

Issue #27992, opened 2016-12-05
Migrate strong mode version of tests from pkg/dev_compiler to tests


vsmenon:
We should move these files out so that kernel and other tools can use them for testing strong mode.  On the DDC side, we'll need to patch:
- test/codegen_test.dart
- test/not_yet_strong_tests.dart (should this be moved / modified?)
- test/browser/language_tests.js (if the names are changed from "corelib" to "corelib_strong")

munificent:
Done: https://github.com/dart-lang/sdk/commit/6c757957db8c31969c8a4c7b6e1693e1d93555bf

PR #28044, opened 2016-12-09
Move the dev_compiler strong mode tests into sdk/tests/.

munificent:
Sending this out as a pull request because Reitveld apparently can't handle a patch this big...

DDC's codegen test copies those files to a local "gen" directory so
that it can do stuff like splitting out the multitests before it
compiles them to JS.

I left all of that alone, so the rest of DDC's test infrastructure is
unchanged. The very first step that builds the "gen" directory just
copies from sdk/tests/..._strong/... instead and the rest is good to go.

I did not move not_yet_strong_tests.dart somewhere more accessible yet
because I'm not sure if kernel needs it or where it should go.

I did not create any status files because DDC doesn't need them and
there are no test suites for the new directories for the other
platforms.


munificent:
Fixes #27992.

vsmenon:
lgtm - seems travis is unhappy though on a couple tests for some reason.

asgerf:
It seems the line ending encoding was changed for some tests that rely on a specific encoding (I believe it affects multiline string literals).

Otherwise LGTM, and thanks!

munificent:
> It seems the line ending encoding was changed for some tests that rely on a specific encoding (I believe it affects multiline string literals).

Oops, yeah, I think Git renormalized them since it thinks they are new files. I'll fix that.

I think that's what's breaking Travis too.

PR #28062, opened 2016-12-10
Use the new path to the linter package

yyoon:
@zanderso @bwilkerson

### DIFF
--- a/pkg/analysis_server/BUILD.gn
+++ b/pkg/analysis_server/BUILD.gn
@@ -11,10 +11,10 @@ dart_package("analysis_server") {
 
   deps = [
     "//dart/pkg/analyzer",
+    "//dart/third_party/pkg/linter",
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

kevmoo:
@yyoon can you merge this? If not, @bwilkerson would you do the honors?

yyoon:
@kevmoo I don't have write access here.

zanderso:
I think this was probably already discussed, but I just wanted to note again that in general the version of the package pulled into //dart/third_party/pkg may not agree with the versions of packages pulled into //third_party/dart-pkg.

However, I suppose that you've tried this change out, and it will unblock you until we get a full SDK built in the Fuchsia tree.

yyoon:
Yes. There aren't too many packages that depend on linter right now, and I've tried all these changes locally to make sure the dart analysis works as expected.

PR #28076, opened 2016-12-12
Remove 'Getting Kernel' readme section

lukechurch:
This looks like a legacy from the old location of kernel.

@kmillikin


lukechurch:
I was clearly being too enthusiastic here.

Issue #27486, opened 2016-10-04
Support covariant method overrides


munificent:
[**Informal spec**](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/covariant-overrides.md)

Dart 1.0 allows any overridden method to tighten the type of a parameter, which is not statically sound without introducing runtime checks.

Strong mode currently uses `@checked`. This must be changed to `covariant`.

Dart2js and the VM should ignore the marker (for now).

For 1.22:
- [x] informal specification: (Merged PR #28176) [current version](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/covariant-overrides.md)
- [x] dart2js #28165
- [x] vm #28166
- [x] analyzer #28167
- [x] ddc
  * [x] support `covariant` syntax. #28168
  * [x] change runtime type of tear-offs. This change is backwards-compatible (since it would allow more). #27484.
- [x] dartfmt https://github.com/dart-lang/dart_style/issues/572
- [x] dartdoc https://github.com/dart-lang/dartdoc/issues/1322
- [x] changelog updated (pending https://codereview.chromium.org/2648203003/)

At some later point:
- full specification #28164

leafpetersen:
@sethladd @Hixie @abarth

leafpetersen:
status: Deciding between two proposals for syntax.  We aim to have a decision soon, but possibly not this week. @munificent needs to review the latest proposal from @floitschG .

sethladd:
Curious, any updates?

leafpetersen:
Basic proposal is accepted and decided on.  Just trying to choose a keyword that everyone hates equally.... :)

@eernstg is incorporating our discussions into the original proposal that @munificent put together.

sethladd:
Thanks for the update! When the original proposal is updated, can you link to it here?

munificent:
> When the original proposal is updated, can you link to it here?

Yup. We've got some other rationalizing to do with it so that it lines up with what @jmesserly already implemented, but as soon as we can, we'll put it somewhere public.

leafpetersen:
A brief summary of the current proposal.  A more detailed development will be forthcoming shortly.  We don't yet have a choice for the modifier name that we're all happy with, but we will close on this soon.  

[...]

PR #28176, opened 2016-12-21
Informal proposal for covariant overrides.

munificent:
This is the informal proposal for the in-progress covariant overrides feature (#27486).

### DIFF
--- a/docs/language/informal/covariant-overrides.md
+++ b/docs/language/informal/covariant-overrides.md
@@ -41,6 +41,19 @@ simpleFormalParameter: // CHANGED
 
 declaredIdentifier: // CHANGED
   metadata 'covariant'? finalConstVarOrType identifier
+
+declaration: // CHANGED: last alternative
+  constantConstructorSignature (redirection | initializers)? |
+  constructorSignature (redirection | initializers)? |
+  'external' constantConstructorSignature |
+  'external' constructorSignature |
+  ('external' 'static' ?)? getterSignature |
+  ('external' 'static'?)? setterSignature |
+  'external'? operatorSignature |
+  ('external' 'static'?)? functionSignature |
+  'static' ('final' | 'const') type? staticFinalDeclarationList |
+  'final' type? initializedIdentifierList |
+  'static'? ('var' | 'covariant'? type) initializedIdentifierList
 ```
 
 ### Standard mode
--- a/docs/language/informal/covariant-overrides.md
+++ b/docs/language/informal/covariant-overrides.md
@@ -11,8 +11,7 @@ many UI frameworks.
 
 Note that this feature is relevant in strong mode where parameter types
 cannot otherwise be tightened, but it affects standard mode Dart in the
-sense that the syntax should be accepted and ignored. We specify the
-feature for strong mode first, and then for standard mode.
+sense that the syntax should be accepted and ignored.
 
 ## Informal specification
 
@@ -98,8 +97,13 @@ An invocation of an instance method with one or more covariant parameters
 is checked as if no `covariant` modifiers had been present on any of the
 involved declarations.
 
-*Intuitively, covariant overrides are irrelevant for clients, it is a
-feature which is encapsulated in the invoked method.*
+*From one point of view, covariant overrides are irrelevant for clients,
+it is a feature which is encapsulated in the invoked method. This is
+reflected in the typing. From another point of view, clients may need to
+provide arguments with a proper subtype of the one required in the static
+type, because there may be a dynamic check for that subtype. This is
+handled by developers using ad-hoc reasoning and suitable programming
+idioms. The whole point of this mechanism is to allow this.*
 
 ##### Overriding
 
--- a/docs/language/informal/covariant-overrides.md
+++ b/docs/language/informal/covariant-overrides.md
@@ -183,9 +183,7 @@ The return type of *f* the is the static return type of *D*. For each
 parameter `p` declared in *D* which is not covariant, the part in the
 dynamic type of *f* which corresponds to `p` is the static type of `p` in
 *D*. For each covariant parameter `q`, the part in the dynamic type of *f*
-which corresponds to `q` is the least upper bound of all the types declared
-for parameters corresponding to `q` in directly and indirectly overridden
-declarations of `m`.
+which corresponds to `q` is `Object`.
 
 # Background Material
 
--- a/docs/language/informal/covariant-overrides.md
+++ b/docs/language/informal/covariant-overrides.md
@@ -1,6 +1,6 @@
 # Covariant Overrides
 
-Owner: rnystrom@, eernst@.
+Owner: rnystrom@, eernstg@.
 
 ## Summary
 
@@ -14,10 +14,12 @@ cannot otherwise be tightened, but it affects standard mode Dart in the
 sense that the syntax should be accepted and ignored. We specify the
 feature for strong mode first, and then for standard mode.
 
-## Informal specification (strong mode)
+## Informal specification
 
-We set out by giving the informal specification for this
-feature in strong mode.
+We set out by giving the informal specification for the syntax of this
+feature (which is shared among standard mode and strong mode). Following
+that, we specify the other aspects of the feature in standard mode,
+followed by those other aspects in strong mode.
 
 ### Syntax
 
@@ -42,11 +44,26 @@ declaredIdentifier: // CHANGED
   metadata 'covariant'? finalConstVarOrType identifier
 ```
 
-### Static checking
+### Standard mode
 
-The static checks are affected in several ways. In this section we discuss
-a few general issues; several larger subtopics are discussed in the
-following sections.
+The static analysis in standard mode ignores `covariant` modifiers, and so
+does the dynamic semantics.
+
+*This means that covariant overrides are essentially ignored in standard
+mode. The feature is useless because covariant parameter types are always
+allowed, but we wish to enable source code to be used in both standard and
+strong mode. So standard mode needs to include support for accepting and
+ignoring the syntax.*
+
+### Strong mode
+
+In strong mode, the covariant overrides feature affects the static analysis
+and dynamic semantics in several ways.
+
+#### Static checking
+
+In this section we discuss a few general issues; several larger subtopics
+are discussed in the following subsections.
 
 It is a compile-time error if the `covariant` modifier occurs on a
 parameter of a function which is not an instance method (which includes
@@ -59,7 +76,6 @@ operator) `m`, `p` is considered to be a **covariant parameter** if there
 is a direct or indirect supertype containing an overridden declaration of
 `m` where the parameter corresponding to `p` has the modifier `covariant`.
 
-<-- Commentary: Shown in italics as in the other informal specs -->
 *In short, the property of being covariant is inherited, for each
 parameter. There is no conflict if only some overridden declarations have
 the `covariant` modifier, and others do not. The parameter is covariant iff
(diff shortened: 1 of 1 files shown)


munificent:
cc @eernstg

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+
+`covariant` modifiers are ignored when deciding subtype relationships among
+classes. So this:
+
+```dart
+var group = new RadioGroup();
+Widget widget = group;
+```
+
+... is perfectly fine.
```
lrhn: I'm not sure what this shows - subtyping never had anything to do with member signatures. Either elaborate what the concern could be, or remove the section, because I can't what point it's trying to make.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+```
+
+... reports an error on the marked line. But this:
+
+```dart
+Widget widget = new RadioGroup();
+widget.addChild(new Widget());
+```
+
+... does not. Both will fail at run time, but the whole point of allowing
```
lrhn: The one that reports an error at compile time won't fail at runtime - since it can never run.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+
+#### The runtime type of a tear-off
+
+For any given declaration *D* of a method `m`, the reified type
+obtained by tearing off `m` from a receiver whose dynamic type is a
+class `C` wherein *D* is the most specific declaration of `m` (that
+is, *D* has not been overridden by any other method declaration in
+`C`) can be computed as follows: Collect the set of all declarations
+of `m` in direct and indirect supertypes of `C`. For the reified
+return type, choose the greatest lower bound of all declared return
```
lrhn: Just use the actual return type. It must be a subtype of all other return types, and there is nothing special about return types that makes you need to look at super declarations.
It's only because the "original type" is different from the "desired type" and "original type" isn't written in the function declaration, that we need to look at super-declarations to find it for covariant parameters.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+As a consequence, the reified type of a torn off method is at least as
+specific as every possible statically known type for that method, and
+that means that it is safe to assign the torn off method to a
+variable whose declared type is the statically known type of that
+method, no matter which supertype of the dynamic type of the receiver
+is the statically known receiver type.
+
+(Note that the computation of a "least upper bound" in Dart does not
+in fact yield a *least* upper bound, but it does yield some *upper
+bound*, which is sufficient to ensure that this safety property holds.)
```
lrhn: The LUB/MAX we are discussing elsewhere has the property that it sometimes has to give up. In that case, we should probably use `Object` here (which is always an upper bound). For inference, we want something else that defaulting to `Object`, so it's not part of the "LUB" computation.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+Instead of "covariant", we could use "checked" to indicate that invocations
+passing actual arguments to this parameter will be subject to a runtime
+check. We decided not to do this, because it puts the focus on an action
+which is required by the type system to enforce a certain discipline on the
+program behavior at runtime, but the developers should be allowed to focus
+on what the program will do and how they may achieve that, not how it might
+fail.
+
+The word "covariant" may be esoteric, but if a developer knows the word or
+looks it up the meaning of this word directly addresses the functionality that
```
lrhn: Long sentence. Could there be a comma before "the meaning"?

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+    Even though the desired type is ensconced in the body of the method in the
+    manual pattern, it really is part of the method's signature. If you are
+    calling `addChild()` *on an object that you statically know is a
+    RadioGroup*, you want the errors as if its declared type was the tighter
+    desired type, not the original one. This is something the manual pattern
+    can't express.
+
+So we need to understand the original type, the desired type, and when
+this feature comes into play. To enable it on a method parameter, you
+mark it with the contextual keyword `covariant`:
```
bwilkerson: Does "contextual" keyword mean that `covariant` is neither a keyword nor a built-in identifier in the sense defined in the spec?

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+
+```
+normalFormalParameter:
+  "covariant"? functionSignature |
+  fieldFormalParameter |
+  simpleFormalParameter
+  ;
+
+simpleFormalParameter:
+  metadata "covariant"? finalConstVarOrType? identifier
```
bwilkerson: If `covariant` is neither a keyword nor a built-in identifier, then it can also be used as a type name, so it makes sense that we would require a type to follow it in order to prevent ambiguity.
```
class covariant {
  void m(covariant covariant c) {}
}
```
but what about the case of a function signature? Consider
```
class covariant {}
class Timer {
   int time(covariant f()) {}
}
```
How should we interpret the `covariant` before the parameter `f`?

Also, as a minor nit, the way the production `normalFormalParameter` is written the `covariant` keyword comes before the metadata for the parameter. I assume that isn't the intent. (Technically, the spec makes the same mistake for `static` and/or `external` functions, but analyzer ignores that and expects the metadata to precede any keywords. I expect we should do the same here.)

docs/language/informal/covariant-overrides.md
lrhn: If possible, yes. That's the least breaking way to introduce it.

The alternative is to make it a built-in identifier like `get` and `set`.

docs/language/informal/covariant-overrides.md
lrhn: That was the reason that I want to require the type.
The function type is a problem, though.

bwilkerson: Looking more closely at Lasse's suggestion, it isn't clear to me why we'd disallow the use of either `const` or `var` with `covariant`, or why `final` would require a type. If those aren't desired restrictions, how about:
```
simpleFormalParameter:
  metadata ("covariant" finalConstVarOrType | finalConstVarOrType?) identifier
```

lrhn: I think that could work too. The only problem with omitting the type (and why people will probably not do it even if they can) is that it's not clear which type the parameter is covariant wrt. Still, it should work.

Also, the function type example above isn't a problem because that syntax isn't handler by `simpleFormalParameter`. It means that you can't write `covariant` in front of a function signature (or if you do, it's a type - but that's silly, so maybe we should just make it a built-in identifier as soon as possible, it's not going to break anybody anyway).

mit-mit:
Hey @munificent, can we get this landed?

eernstg:
@mit-mit I just returned to work on this, I hope to have an update which is usable as an informal spec for implementors later today.

docs/language/informal/covariant-overrides.md
eernstg: I've introduced a new `functionFormalParameter` to work as the variant of `functionSignature` which is intended to declare a formal parameter. That one needs to admit `covariant` on the parameter declaration as a whole, but other usages of `functionSignature` (e.g., for declaring a top-level function) should not. We can't just add `covariant` in front of a `functionSignature` in the rule for `normalFormalParameter`, because it would then put the metadata in an inconsistent location.

Next, I have made `covariant` a **built-in identifier**, which I believe is accepted at this point. So `covariant` cannot be a type, and hence we cannot have the function header `void foo(covariant covariant covariant);`.

This means that we do not need to disambiguate unconditionally when seeing `covariant` in the parser, which means that we _do_ allow omitting the type (`void foo(covariant x);`) and using `covariant` as the name of the parameter (`void foo(covariant);` and `void foo(bool covariant);`).

As a consequence, we interpret `int time(covariant f())` as having a covariant parameter named `f` whose type is `dynamic Function()`.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+```
+
+<!--
+  Note that it is possible to mark a parameter as covariant and leave out the
+  type, e.g., `covariant var x`. This is probably useless, but benign.
+-->
+
+The `covariant` modifier can *only* be applied to a parameter declaration
+on an instance method, instance setter, or instance operator, and it can be
+applied to a mutable field declaration. Anywhere else is a static error.
```
eernstg: Now using `non-final instance variable`.

eernstg: I'm now using the usual spec style 'It is a compile-time error if ..'.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+
+### Static checking
+
+The static checks are affected in several ways when adding covariant overrides
+to the language.
+
+#### Overriding
+
+With this feature, type checking of an overriding instance method or operator
```
eernstg: This is now completely different.

docs/language/informal/covariant-overrides.md
```diff
@@ -0,0 +1,742 @@
...
+We could have chosen to require that every supertype chain must make the
+parameter covariant, rather than just requiring that there exists such a
+supertype chain. This is a software engineering trade off: both choices
+can be implemented, and both choices provide a certain amount of protection
+against tightening types by accident. We have chosen the permissive variant,
+and it is always possible to make a linter require a stricter one.
+
+For a regular (non-covariant) parameter, the override rule is is unchanged:
+its type must be a supertype of the type declared for the same parameter in
+each directly overridden declaration. This includes the typical case where the
```
eernstg: The spec defines 'overrides' to mean 'overrides directly' (cf. \ref{inheritanceAndOverriding}: 'Then $m$ overrides $m^\prime$ if $m^\prime$ is not already overridden by a member of at least one of $S_1 \ldots S_{j-1}$'). This means that we can talk about 'overrides' and 'indirectly overrides'. But I thought that would be too subtle, so I made it a bit more explicit and used 'directly overrides' and 'indirectly overrides'.

Of courses, the spec never needs to talk about 'indirectly overrides', because it suffices to talk about the relationship between a declaration and the declaration that it directly overrides for all current purposes. But that won't suffice for covariant parameter type checking.

docs/language/informal/covariant-overrides.md
eernstg: Removed from the informal spec part, which is what I'm currently working on. It is still present at the end of the document (I'll return to that soon).

docs/language/informal/covariant-overrides.md
eernstg: Right, I'll just use the declared return type!

docs/language/informal/covariant-overrides.md
eernstg: We could also simply reject the covariant parameter which creates a situation where there is no "reasonable" dynamic type. (Or we could use guarded functions ;-)

docs/language/informal/covariant-overrides.md
eernstg: I've had widespread support for making it a built-in identifier today. We can take another fight or two about it, but my current work is based on that assumption.

eernstg:
I just pushed a version of covariant-override.md which includes an informal specification at the beginning. This involved extracting a substantial amount of material from the document as a whole, which means that there is a lot of near-redundant material (same topic is addressed in the informal spec and, more verbosely, in the rest of the document). I'll look into that tomorrow.

docs/language/informal/covariant-overrides.md
eernstg: Ah, @bwilkerson, I overlooked that you already mentioned the metadata/"keyword" ordering issue. As mentioned, I do handle the issue for `covariant` in the proposal, but I created a [new issue](https://github.com/dart-lang/sdk/issues/28283) about the similar issues that arise with `static` and `external`.

Issue #27499, opened 2016-10-05
Specify strong mode "top level" inference


munificent:
We've spent a lot of time discussing strategies around how we handle type inference of "top level" declarations—top level variables and static fields. There is what strong mode currently implements, and the revised rules we've considered informally.

The next step is to write up the set of rules we want to commit to.

Edit by @eernstg Feb 2018: Note that this is now essentially one of the subtasks of #31228.

leafpetersen:
Proposal is now here for consideration: https://github.com/dart-lang/sdk/pull/28218 .

leafpetersen:
Tracking bug for implementation here:  https://github.com/dart-lang/sdk/issues/28219

anders-sandholm:
Can we close this one?

eernstg:
No Dart 2 milestone: This is now considered a subtask of #31228, which has a Dart 2 milestone.

leafpetersen:
I have a pull request out with a specification for this.

PR #28218, opened 2016-12-30
Informal proposal for strong mode toplevel inference

leafpetersen:
This is an informal specification of proposed changes to strong mode top level inference (https://github.com/dart-lang/sdk/issues/27499).

### DIFF
--- a/docs/language/informal/toplevel-inference.md
+++ b/docs/language/informal/toplevel-inference.md
@@ -10,9 +10,9 @@ Status: Proposal
 Top level inference has two phases:
 
 1. **Method override inference**
-    * If you omit a return type or parameter type from an overridden method,
-    inference will try to fill in the missing type using the signature of the
-    method you are overriding.
+    * If you omit a return type or parameter type from an overridden or
+    implemented method, inference will try to fill in the missing type using the
+    signature of the method you are overriding.
 2. **Static variable and field inference** 
     * If you omit the type of a field, setter, or getter, which overrides a
    corresponding member of a superclass, then inference will try to fill in the
@@ -98,12 +98,12 @@ and it does not override or implement anything from a super type, then the
 omitted types are treated as dynamic.
 
 
-Otherwise, the missing types are filled in with the type from the overridden
-method.  If there are multiple overridden/implemented methods and the type to be
-filled in does not agree, it is an error.  If there is no corresponding
-parameter position in the overridden method to infer from, it is treated as
-dynamic (e.g. overriding a one parameter method with a method that takes a
-second optional parameter).
+Otherwise, the missing types are filled in with the type from the overridden or
+implemented method.  If there are multiple overridden/implemented methods and
+the type to be filled in does not agree, it is an error.  If there is no
+corresponding parameter position in the overridden method to infer from, it is
+treated as dynamic (e.g. overriding a one parameter method with a method that
+takes a second optional parameter).
 
 
 ### Setter return types
@@ -236,6 +236,9 @@ which is not an IE has no inferred type.  For every immediately evident
 expression, we also define the set of variable dependencies upon which its
 inference relies.
 
+The intention is that toplevel inference should always return the same result as
+local inference, or else produce an error.
+
 
 * null, boolean, numeric, string, type, and symbol literals have their
    corresponding type.
@@ -253,8 +256,23 @@ evident expression with type `T` or `Future<T>`.
   two returned sub-expressions.
   * Inference dependencies are the union of the inference dependencies of the
   two returned sub-expressions.
-* Logical boolean expressions have inferred type `boolean`.
+* Logical boolean expressions and equality expressions have inferred type
+  `boolean`.
   * No inference dependencies
+* Relational expressions, bitwise expressions, and shift expressions are treated
+as the appropriate method call as defined in the spec.
+  * Inference dependencies are those of the left hand operand.
+* A multiplicative binary expression is treated as a method call as defined in
+  the spec, with the same exceptions for when the left hand operand is of type `int`.
+  * Inference dependencies are those of the left hand operand.
+* Applications of the prefix operator `!` have type `boolean`
+  * No inference dependencies  
+* Applications of the prefix operators `++` and `--` have the type of the operand.
+  * Inference dependencies are those of the operand.
+* Applications of other prefix operators are treated as method calls.
+  * Inference dependencies are those of the operand.
+* Application of a postfix operator to an expression `e` has the inferred type of `e`.
+  * Inference dependencies are those of the operand.
 * A list literal or a map literal with explicit type arguments has the type
    `List<T>` or `Map<K, V>` respectively, where `<T>` or `<K, V>` are the
    provided type arguments.
@@ -293,9 +311,6 @@ evident expression with type `T` or `Future<T>`.
    * If the function is marked `async`, then the return type of the function is
      inferred as `Future<flatten<T>>` where `T` is the inferred type of the
      return expression and flatten is as described in the spec.
-* An instance creation expression with no omitted type arguments has the
-   obvious type.
-   * No inference dependencies
 * An instance creation expression with no omitted type arguments has the
    obvious type.
    * No inference dependencies
--- a/docs/language/informal/toplevel-inference.md
+++ b/docs/language/informal/toplevel-inference.md
@@ -70,17 +70,17 @@ var b = a.x;
 
 
 To ensure stability with respect to cycles and different orderings, current
-strong mode does inference in two passes: first on the static variables, and
-secondly on fields (method override based inference is mostly irrelevant for
-this discussion). This ensures that inference is stable, but it has several
-unfortunate consequences, most notably the following.  Inference may result in
-the same variable being implicitly seen at two different types. In the example
-above, if all declarations are in the same library, then b will be inferred
-before x has been inferred, and so b will be inferred to have type dynamic.  The
-same declaration in a separate library will see the type of a.x after inference,
-and hence will be inferred to have type `int`.  This violates one of the
-principles above, and also means that changes to the library structure that
-introduce a cycle may change inference results.
+strong mode does inference in two passes: firstly on the static and top-level
+variables, and secondly on fields (method override based inference is mostly
+irrelevant for this discussion). This ensures that inference is stable, but it
+has several unfortunate consequences, most notably the following.  Inference may
+result in the same variable being implicitly seen at two different types. In the
+example above, if all declarations are in the same library, then b will be
+inferred before x has been inferred, and so b will be inferred to have type
+dynamic.  The same declaration in a separate library will see the type of a.x
+after inference, and hence will be inferred to have type `int`.  This violates
+one of the principles above, and also means that changes to the library
+structure that introduce a cycle may change inference results.
 
 
 ## Method inference
@@ -116,11 +116,11 @@ If the return type is left off a setter, the return type is inferred to be
 ## Static variable and field inference
 
 
-A “candidate” for inference is any of the following:
+A “candidate” for inference is one of any of the following kinds of definitions,
+for which the type annotation has been omitted:
 * A top level variable definition
 * A static class level variable definition
-* An instance variable, getter or setter definition with the type annotation
-  omitted.
+* An instance variable, getter or setter definition.
 
 
 The *inference dependencies of a candidate* which is inferred via override
@@ -168,10 +168,10 @@ initializer.
 #### Instance field, getter, and setter inference
 
 
-The inferred type of a getter, setter, or field is as follows.  Note that I say
+The inferred type of a getter, setter, or field is as follows.  Note that we say
 that a setter overrides a getter if there is a getter of the same name in some
-superclass or interface, and similarly for setters overriding getters, fields,
-etc.
+superclass or interface (explicitly declared or induced by an instance variable
+declaration), and similarly for setters overriding getters, fields, etc.
 
 
 A getter, setter or field which overrides/implements only a getter is inferred
@@ -227,7 +227,7 @@ result.  Implementations are free to treat this as a result of dynamic after
 emitting the error for the purposes of emitting subsequent error messages, but
 the behavior of programs for which initializer inference fails is unspecified.
 
-#### Immediately evident expressions
+#### Expressions with immediately-evident type
 
 We define a set of expressions with immediately-evident types (called
 immediately-evident expressions, abbreviated IE) as follows.  Immediately
@@ -237,9 +237,24 @@ expression, we also define the set of variable dependencies upon which its
 inference relies.
 
 
-* null, boolean, numeric, string, and symbol literals have their corresponding
-   type.
+* null, boolean, numeric, string, type, and symbol literals have their
+   corresponding type.
    * No inference dependencies
+* An `await` of an expression has type `T` if the expression is an immediately
+evident expression with type `T` or `Future<T>`.
+   * Inference dependencies are the inference dependencies of the awaited
+     expression
+* A `throw` expression has type `bottom`.
+   * No inference dependencies
+* A parenthesized expression has the inferred type of the sub-expression
+   * Inference dependencies are the dependencies of the sub-expression
+* A conditional expression where both values are immediately evident expressions
+  has type `T` where `T` is the least upper bound of the inferred type of the
+  two returned sub-expressions.
+  * Inference dependencies are the union of the inference dependencies of the
+  two returned sub-expressions.
+* Logical boolean expressions have inferred type `boolean`.
+  * No inference dependencies
 * A list literal or a map literal with explicit type arguments has the type
    `List<T>` or `Map<K, V>` respectively, where `<T>` or `<K, V>` are the
    provided type arguments.
@@ -254,7 +269,7 @@ inference relies.
   `List<dynamic>`.
    * No inference dependencies.
 * A const empty list literal with no type arguments has the type
-  `List<bottom>`.
+  `List<Null>`.
    * No inference dependencies.
 * A map literal with no type arguments, and for which all of the keys and
    values are IEs, has the type `Map<K, V>` where `K` is the least upper bound of
@@ -266,29 +281,39 @@ inference relies.
 * An non-const empty map literal with no type arguments has the type
   `Map<dynamic, dynamic>`.
    * No inference dependencies.
-* A const empty map literal with no type arguments has the type `Map<bottom,
-  bottom>`.
+* A const empty map literal with no type arguments has the type `Map<Null,
+  Null>`.
    * No inference dependencies.
-* A type literal has type `Type`
-   * No inference dependencies
 * A function literal with an expression body for which all parameter types are
    type annotated and for which the return expression is an IE and has an
    inferred type, has the corresponding function type.
    * If the return expression has no inferred type it is an error.
    * The inference dependencies are the inference dependencies of the return
-      expression.
+   expression.
+   * If the function is marked `async`, then the return type of the function is
+     inferred as `Future<flatten<T>>` where `T` is the inferred type of the
+     return expression and flatten is as described in the spec.
+* An instance creation expression with no omitted type arguments has the
+   obvious type.
+   * No inference dependencies
 * An instance creation expression with no omitted type arguments has the
    obvious type.
    * No inference dependencies
 * An `as` expression has the cast to type
    * No inference dependencies
+* An `is` expression has type boolean
+   * No inference dependencies
 * A simple or qualified identifier referring to a top level function, static
    variable, field, getter; or a static class variable, static getter or method;
-   has the inferred type of the identifier.
-   * If the identifier has no inferred type it is an error.
+   or an instance method; has the inferred type of the identifier.
+   * Otherwise, if the identifier has no inferred or annotated type then it is
+   an error.
+   * Note: specifically, references to instance fields and instance getters are
+     disallowed here.
    * The inference dependency of the identifier is itself if the identifier is
       a candidate for inference.  Otherwise there are no inference dependencies.
-* A simple identifier denoting a formal parameter has the type with which it was
+* A simple identifier denoting a formal parameter which has an annotated type is
+  an immediately evident expression and has the type with which it was
   annotated.
    * No inference dependencies.
 * A function (or function expression) invocation with no omitted generic
@@ -309,3 +334,5 @@ inference relies.
    * The inference dependencies are the inference dependencies of the cascade
       target.
 
+Note: Assignment expressions are not immediately evident expressions, and
+if-null expressions are not immediately evident expressions.
--- a/docs/language/informal/toplevel-inference.md
+++ b/docs/language/informal/toplevel-inference.md
@@ -1,0 +1,293 @@
+﻿# Toplevel inference
+
+Owner: leafp@google.com
+
+Status: Proposal
+
+
+## Top level inference overview
+
+Top level inference has two phases:
+1. Method inference
+2. Static variable and field inference
+
+
+All results from the first phase are available in the second phase.
+
+
+As a general principle, when inference fails it is an error.  Tools are free to
+behave in implementation specific ways to provide a graceful user experience,
+but with respect to the language and its semantics, programs for which inference
+fails are unspecified.
+
+
+The intuitive idea behind this proposal is that top level inference works by
+first inferring methods, and then doing inference for "field like things"
+(static variables, fields, setters, and getters), where we don't allow inference
+for "field like things" to depend on the results of inference for "field like
+things" except in cases where the dependencies are statically visible.
+
+
+Some broad principles for type inference that the language team has agreed to,
+and which this proposal is designed to satisfy:
+* Type inference should be free to fail with an error, rather than being
+  required to always produce some answer
+* It should not be possible for a programmer to observe a declaration as having
+  two different types (e.g. dynamic for recursive uses, but subsequently int).
+  Some consistent answer (or an error) should always be produced.
+* The inference for local variables, toplevel variables, and fields, should
+  either agree or error out.  The same expression should not be inferred
+  differently at different syntactic positions.  It’s ok for an expression to be
+  inferrable at one level but not at another.
+* Obvious types should be inferred
+* Inferred and annotated types should be treated the same
+* Simple intuition for users
+* Efficient to implement
+* Type arguments should be mostly inferred
+
+
+A motivating example for the current strong mode implementation (that also
+serves as a demonstration of its shortcomings) is as follows:
+
+
+```dart
+class A {
+  var x = 3;
+}
+var a = new A();
+var b = a.x;
+```
+
+
+To ensure stability with respect to cycles and different orderings, current
+strong mode does inference in two passes: first on the static variables, and
+secondly on fields (method override based inference is mostly irrelevant for
+this discussion). The implementation ensures that inference is stable, but it
+has several unfortunate consequences, most notably the following.  Inference may
+result in the same variable being implicitly seen at two different types. In the
+example above, if all declarations are in the same library, then b will be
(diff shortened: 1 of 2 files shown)


docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+If a method leaves a type off of its signature (parameter type or return type)
+and it does not override or implement anything from a super type, it is an
+error.
+
+
+If a method leaves a type off of its signature (parameter type or return type)
+and it does not override or implement anything from a super type, then the
+omitted types are treated as dynamic.
```
bwilkerson: Am I missing something, or is the sentence at line 91 contradicting the one at line 86?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+   * If the elements do not all have an inferred type, it is an error
+   * The inference dependencies of the list literal are the collected
+      dependencies of the elements.
+* A map literal with no type arguments, and for which all of the keys and
+   values are IEs, has the type `Map<K, V>` where `K` is the least upper bound of
+   the inferred types of the provided keys, and `V` is the least upper bound of
+   the provided values.
+   * If the keys and values do not all have an inferred type, it is an error.
+   * The inference dependencies of the map literal are the collected
+      dependencies of the keys and values.
```
munificent: What about empty maps and lists?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+* A type literal has type `Type`
+   * No inference dependencies
+* A function literal with an expression body for which all parameter types are
+   specified and for which the return expression is an IE and has an inferred
```
munificent: By "all parameter types are specified" to you mean explicitly annotated? If so, maybe "all parameters are type annotated"?

I'm guessing there are no cases where the parameter types can be inferred?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+   * If the return expression has no inferred type it is an error.
+   * The inference dependencies are the inference dependencies of the return
+      expression.
+* An instance creation expression with no omitted type arguments has the
+   obvious type.
+   * No inference dependencies
+* An `as` expression has the cast to type
+   * No inference dependencies
+* A simple or qualified identifier referring to a top level function, getter,
+   or a static class getter or method, has the inferred type of the identifier.
```
munificent: Does this include top level variables and static fields too? If so, might be good to call that even though technically they are synthesized getters.

munificent (approved):
Couple small things here and there, but otherwise this looks really nice.

It might be worth talking a little about how this compares to the current implementation. Which parts are different? Which currently-inferred won't be with this and vice versa?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+   * If the identifier has no inferred type it is an error.
+   * The inference dependency of the identifier is itself if the identifier is
+      a candidate for inference.  Otherwise there are no inference dependencies.
+* A simple identifier denoting a formal parameter has the annotated type of the
+   variable
```
scheglov: Which variable? Do you mean "of the parameter"?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+have the type taken from the overridden getter result type.
+
+
+A setter which overrides/implements both a setter and a getter is inferred to
+have the type taken from the overridden setter parameter type.
+
+
+A field which overrides/implements both a setter and a getter is inferred to
+have the type taken from the overridden setter parameter type if this type is
+the same as the return type of the overridden getter (if the types don’t match
```
scheglov: Does "match" mean "the same"?
If so, it would be to write so explicitly to prevent any doubts.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
+﻿# Toplevel inference
+
+Owner: leafp@google.com
+
+Status: Proposal
+
+
+## Top level inference overview
+
+Top level inference has two phases:
```
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+1. Method inference
```
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+behave in implementation specific ways to provide a graceful user experience,
+but with respect to the language and its semantics, programs for which inference
+fails are unspecified.
+
+
+The intuitive idea behind this proposal is that top level inference works by
+first inferring methods, and then doing inference for "field like things"
+(static variables, fields, setters, and getters), where we don't allow inference
+for "field like things" to depend on the results of inference for "field like
+things" except in cases where the dependencies are statically visible.
```
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+Some broad principles for type inference that the language team has agreed to,
+and which this proposal is designed to satisfy:
+* Type inference should be free to fail with an error, rather than being
+  required to always produce some answer
+* It should not be possible for a programmer to observe a declaration as having
+  two different types (e.g. dynamic for recursive uses, but subsequently int).
+  Some consistent answer (or an error) should always be produced.
```
leafpetersen: The e.g. is an example!  Added a forward pointer though.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+}
+var a = new A();
+var b = a.x;
+```
+
+
+To ensure stability with respect to cycles and different orderings, current
+strong mode does inference in two passes: first on the static variables, and
+secondly on fields (method override based inference is mostly irrelevant for
+this discussion). The implementation ensures that inference is stable, but it
```
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Yes.  Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.  This proposal is explicitly designed not to require downwards inference at the toplevel.  We could choose to allow this later, in which case they could be inferred.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: yes.  Clarified.

docs/language/informal/toplevel-inference.md
leafpetersen: Aargh.  Well... aargh.

Is it to weird if we make it `List<dynamic>` for non const, `List<bottom>` for const (and similar for `Map`)?  Unfortunately, I think this captures the most useful pattern.  const empty lists are only useful as lists of `bottom`.  Non const empty lists are a useful shorthand for `dynamic` lists.

The only alternative I see is to make them all `List<bottom>`, at least if we want to be able to use `bottom` to make empty const lists usable as a default value.  I think we'd have to do that in local code to be consistent as well.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+   inferred type, has the corresponding function type.
+   * If the return expression has no inferred type it is an error.
+   * The inference dependencies are the inference dependencies of the return
+      expression.
+* An instance creation expression with no omitted type arguments has the
+   obvious type.
+   * No inference dependencies
+* An `as` expression has the cast to type
+   * No inference dependencies
+* A simple or qualified identifier referring to a top level function, static
```
stereotype441: I had trouble parsing this section.  My understanding after our conversation today is that:
- References to a top level function, top level variable, top level getter, class static method, class static field, or class static getter are all inferred.
- References to a class instance method are inferred (this is possible because instance inference is always done before static variable and field inference).  Note that we're talking about method tear-offs here.
- References to a class instance field or a class instance getter are not inferred (and result in an error).  This is because we can't support this without losing the ability to compute the dependency graph statically.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+  var x = 3;
+}
+var a = new A();
+var b = a.x;
+```
+
+
+To ensure stability with respect to cycles and different orderings, current
+strong mode does inference in two passes: first on the static variables, and
+secondly on fields (method override based inference is mostly irrelevant for
```
eernstg: It makes sense to mention top-level variables because they are in the example: 'first[ly?] on the static [and top-level] variables, and', or maybe 'secondly' --> 'then'.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+The inferred type of a static class variable is the type inferred from its
+initializer.
+
+#### Instance field, getter, and setter inference
+
+
+The inferred type of a getter, setter, or field is as follows.  Note that I say
+that a setter overrides a getter if there is a getter of the same name in some
+superclass or interface, and similarly for setters overriding getters, fields,
+etc.
```
eernstg: Might as well use 'we'. Also, it seems natural to mention here that setters/getters induced by instance variable declarations are included: '.. superclass or interface (explicitly declared, or induced by an instance variable declaration), ..'

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+A meta-goal is to keep this inference relatively lightweight and efficient to
+implement.
+
+
+In all cases, if initializer inference fails, it is an error and produces no
+result.  Implementations are free to treat this as a result of dynamic after
+emitting the error for the purposes of emitting subsequent error messages, but
+the behavior of programs for which initializer inference fails is unspecified.
+
+#### Immediately evident expressions
```
eernstg: I'd prefer 'Expressions with an immediately evident type', such that the reader gets the most informative phrase as the canonical one, up front, and then the abbreviated forms can be used more freely later on.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+* null, boolean, numeric, string, and symbol literals have their corresponding
+   type.
+   * No inference dependencies
+* A list literal or a map literal with explicit type arguments has the type
+   `List<T>` or `Map<K, V>` respectively, where `<T>` or `<K, V>` are the
+   provided type arguments.
+   * No inference dependencies
+* A non-empty list literal with no type arguments, and for which all of the
+   elements are IEs, has the type `List<T>` where `T` is the least upper bound
+   of the inferred types of the elements.
```
eernstg: This is one more location where we discussed LUB (and it's a place where the inferred type argument easily jumps to `Object` now and then, because the list contains a completely unrelated element by accident), so I think we should remind ourselves about the simpler options that we might prefer:

We could require that all elements have the same immediately evident type (maybe too rigid).

We could require that there is a maximal type, and use that (more flexible, and avoids reifying a type argument which was never written by the developer, which has been mentioned several times as a source of confusion).

I tend to prefer a `max` rule. Developers can write the type argument explicitly in the cases where that doesn't suffice.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+   * If the elements do not all have an inferred type, it is an error
+   * The inference dependencies of the list literal are the collected
+      dependencies of the elements.
+* An non-const empty list literal with no type arguments has the type
+  `List<dynamic>`.
+   * No inference dependencies.
+* A const empty list literal with no type arguments has the type
+  `List<bottom>`.
```
eernstg: Unfortunately, we'll probably have to use the notation `List<Null>` for that, for some time.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+   the inferred types of the provided keys, and `V` is the least upper bound of
+   the provided values.
+   * If the keys and values do not all have an inferred type, it is an error.
+   * The inference dependencies of the map literal are the collected
+      dependencies of the keys and values.
+* An non-const empty map literal with no type arguments has the type
+  `Map<dynamic, dynamic>`.
+   * No inference dependencies.
+* A const empty map literal with no type arguments has the type `Map<bottom,
+  bottom>`.
```
eernstg: `Map<Null, Null>`, for now.

docs/language/informal/toplevel-inference.md
eernstg: It's a bit bureaucratic, but I agree that `List<dynamic>` and `List<Null>` as you describe it is the most useful approach.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+   * No inference dependencies.
+* A type literal has type `Type`
+   * No inference dependencies
+* A function literal with an expression body for which all parameter types are
+   type annotated and for which the return expression is an IE and has an
+   inferred type, has the corresponding function type.
```
eernstg: With 'is an IE and has an inferred type', it sounds like an IE without an inferred type can exist. I thought IE was a semantic concept (so we can say that they all have an inferred type), rather than a syntactic one (where we would specify a larger set of expressions and then filter out the ones that don't have an inferred type).

With the semantic approach, I think we shouldn't say 'it is an error' here and there, we should say 'the expression is not an IE'. With the syntactic approach, we'd need to say 'an IE which has an inferred type' in a lot of places.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+* An `as` expression has the cast to type
+   * No inference dependencies
+* A simple or qualified identifier referring to a top level function, static
+   variable, field, getter; or a static class variable, static getter or method;
+   has the inferred type of the identifier.
+   * If the identifier has no inferred type it is an error.
+   * The inference dependency of the identifier is itself if the identifier is
+      a candidate for inference.  Otherwise there are no inference dependencies.
+* A simple identifier denoting a formal parameter has the type with which it was
+  annotated.
```
eernstg: If the declaration of the identifier is a parameter that doesn't have an annotation then the identifier isn't an IE.

eernstg (commented):
Finally found a bit of time to finish the review. ;-)
Clearly, the rule that inferred and annotated types must be treated identically is the most controversial point.

docs/language/informal/toplevel-inference.md
stereotype441: We should clarify what happens if the function literal is declared with "async".  I'm assuming that the return type should be `Future<flatten(T)>` (where `T` is the inferred type of the return expression, and `flatten` is as described in the spec).

Also, I notice that `await ...` expressions are not considered IEs.  Is this intentional?

mit-mit:
@leafpetersen can we land this?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,311 @@
...
+1. **Method override inference**
+    * If you omit a return type or parameter type from an overridden method,
+    inference will try to fill in the missing type using the signature of the
+    method you are overriding.
+2. **Static variable and field inference** 
+    * If you omit the type of a field, setter, or getter, which overrides a
+   corresponding member of a superclass, then inference will try to fill in the
+   missing type using the type of the corresponding member of the superclass.
+    * Otherwise, declarations of static variables and fields that omit a type
+   will be inferred from their initializer if present.
```
leafpetersen: > Hence, I wanted to make that commitment visible---if my new setter in this new class is to get a type from somewhere out there, it ought to be stated as a type annotation. In contrast, relying on an initializer expression in an overridden instance variable declaration could mean that several libraries suddenly get different signatures on several methods, suddenly giving lots of call sites different typing properties,

The same is true of changing the annotated type on a field, for what it's worth.  I hear what you're saying.  I'm not super keen on override inference in general for this reason.

However, we did enunciate a principle when we discussed this in August that there should be no difference in the treatment of inferred types vs annotated types.  Splitting out override inference from initializer inference from fields seems to violate that, and seems like it would lead to odd errors.  For example, this code:

```
class A {
  var x = 3;
}

class B extends A {
  var x;
}
```

Would generate an override error for `x`, since `B.x` would get `dynamic` from override inference before `A.x` got inferred to `int`.  Even worse:

```
class A {
  final var x = 3;
}

class B extends A {
  final var x;
}
```
would generate no error, but would silently infer `B.x` as `dynamic`, while `A.x` would get inferred as `int`.  It feels weird to me.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+  required to always produce some answer
+* It should not be possible for a programmer to observe a declaration as having
+  two different types (e.g. dynamic for recursive uses, but subsequently int).
+  Some consistent answer (or an error) should always be produced.
+* The inference for local variables, toplevel variables, and fields, should
+  either agree or error out.  The same expression should not be inferred
+  differently at different syntactic positions.  It’s ok for an expression to be
+  inferrable at one level but not at another.
+* Obvious types should be inferred
+* Inferred and annotated types should be treated the same
```
leafpetersen: I like the principle because when I look at examples where we don't do this, we seem to get stranger behavior (see my example on your previous comment for example).

I would prefer to resolve this by doing less signature inference, but I'm still somewhat persuaded that insofar as we are going to do signature inference, the least surprising thing is to do it as uniformly as possible.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+## Static variable and field inference
+
+
+A “candidate” for inference is any of the following:
+* A top level variable definition
+* A static class level variable definition
+* An instance variable, getter or setter definition with the type annotation
+  omitted.
```
leafpetersen: > Maybe 'is a definition omitting a type annotation for any of the following:' and deleting 'definition..' in the bullets would work.

Done.

leafpetersen: > Also, I'd expect instance variables to potentially have their type inferred during method inference (based on the treatment of methods which would presumably include getters and setters), so they would only be candidates if they have an initializing expression, and they did not get a type during method inference.

No, getters and setters are not inferred during the main override inference phase.  This proposal treats getters and setters more like fields than like methods.  This allows them to get information from override inference, but prevents them from being used to infer initializer types.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+The *inference dependencies of a candidate* which is inferred via override
+inference are the instance fields, setters, and getters that the candidate
+overrides.
```
leafpetersen: The issue is that for fields/getters/setters, we allow initializer inference to feed into override inference, so this must be done separately from method override inference.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Acknowledged.  I think we need to revisit this globally for inference.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Is the intent clear enough for now?  I agree that this should be stated precisely, but maybe we can leave this for the formal spec?

leafpetersen: Added `await` and some other missing cases.

docs/language/informal/toplevel-inference.md
leafpetersen: Clarified.  Is that better?

docs/language/informal/toplevel-inference.md
leafpetersen: Done.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,338 @@
...
+   * The inference dependencies are the inference dependencies of the return
+   expression.
+   * If the function is marked `async`, then the return type of the function is
+     inferred as `Future<flatten<T>>` where `T` is the inferred type of the
+     return expression and flatten is as described in the spec.
+* An instance creation expression with no omitted type arguments has the
+   obvious type.
+   * No inference dependencies
+* An instance creation expression with no omitted type arguments has the
+   obvious type.
```
scheglov: Is this item different than the previous one?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,338 @@
...
+* A function (or function expression) invocation with no omitted generic
+   arguments where the applicand is an IE that has an inferred type with a
+   return type of `T`, has type `T` (with any generic arguments substituted in).
+   * If the applicand has no inferred type, it is an error.
+   * The inference dependencies are the inference dependencies of the
+      applicand.
+* A method invocation `o.m(...)` with no omitted type arguments where the
+   receiver is an IE with inferred type `T` such that `T` provides a signature
+   for `m` with return type `S` has type `S` (with any generic arguments
+   substituted in).
```
scheglov: Are binary expressions handled as method invocations?
Is `a + b` handled as `a.+(b)`?
Will you specify additional logic for `int + int`, `int + double`, `double + double`, etc?

Are unary expressions handled as method invocations?

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,338 @@
...
+
+Status: Proposal
+
+
+## Top level inference overview
+
+Top level inference has two phases:
+
+1. **Method override inference**
+    * If you omit a return type or parameter type from an overridden method,
```
scheglov: What does exactly `overridden method` mean?

docs/language/informal/toplevel-inference.md
leafpetersen: Duplicate.  Done.

docs/language/informal/toplevel-inference.md
leafpetersen: Yes.  In general, the intention is that the inferred toplevel type should be the same as the inferred type in a local context.  I added bullets for binary and unary expressions.

docs/language/informal/toplevel-inference.md
scheglov: There are fairly complex rules about methods overriding in the specification. I try to understand whether you indent that we follow all these rules, or whether we use simple 'find by name and infer formal parameter types one-by-one'.

leafpetersen: Can you be more concrete about cases where the interpretation differs?  Note I am not defining how inference behaves on erroneous cases such as defining a method with the same name as a field from a superclass, or defining a method with the same name as a superclass method with an incompatible signature.  Those are errors: the tool can choose how best to handle surfacing those errors in a way that doesn't cause noise from cascading inference errors.

scheglov: For example this code is not an error.
But what is the type of `a` in `C`?

```dart
class A {
  foo(num a) {}
}

class B {
  foo(double a) {}
}

class C extends A implements B {
  foo(a) {}
}
```

leafpetersen: That's intended to be addressed in the section below:  

```
If there are multiple overridden/implemented methods and
the type to be filled in does not agree, it is an error.  
```

So your example is an inference failure error.

scheglov: OK
The exact meaning of `does not agree` was not clear.
Now I understand what it means.
Thank you.

leafpetersen: Clarified.

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,360 @@
...
+* An instance creation expression with no omitted type arguments has the
+   obvious type.
+   * No inference dependencies
+* An `as` expression has the cast to type
+   * No inference dependencies
+* An `is` expression has type boolean
+   * No inference dependencies
+* A simple or qualified identifier referring to a top level function, static
+   variable, field, getter; or a static class variable, static getter or method;
+   or an instance method; has the inferred type of the identifier.
```
stereotype441: Consider changing this to `...has the inferred type of the referent.`

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,360 @@
...
+   * Otherwise, if the identifier has no inferred or annotated type then it is
+   an error.
+   * Note: specifically, references to instance fields and instance getters are
+     disallowed here.
+   * The inference dependency of the identifier is itself if the identifier is
```
stereotype441: Consider changing this to `The inference dependency of the identifier is the referent if the referent is a candidate for inference...`

docs/language/informal/toplevel-inference.md
```diff
@@ -0,0 +1,360 @@
...
+  two returned sub-expressions.
+* Logical boolean expressions and equality expressions have inferred type
+  `boolean`.
+  * No inference dependencies
+* Relational expressions, bitwise expressions, and shift expressions are treated
+as the appropriate method call as defined in the spec.
+  * Inference dependencies are those of the left hand operand.
+* A multiplicative binary expression is treated as a method call as defined in
+  the spec, with the same exceptions for when the left hand operand is of type `int`.
+  * Inference dependencies are those of the left hand operand.
```
stereotype441: This doesn't work for `*` and `%`, since the exceptions for when the left hand operand is of type `int` require us to know the type of the right hand operand; we can't know this type reliably unless the right hand operand is also an interface dependency.

A similar situation exists for `+` and `-`, which it looks like you forgot to cover (I don't see a bullet addressing additive expressions).

Suggestion: change this to:

* A multiplicative or additive binary expression is treated as a method call as defined in the spec, with the same exceptions for when the left hand operand is of type `int`.
  * Inference dependencies are those of the left hand and right hand operands if the operator is `*`, `%`, `+`, or `-`.
  * Otherwise, inference dependencies are those of the left hand operand.

pkg/dev_compiler/tool/global_compile.dart
```diff
@@ -60,7 +60,7 @@ void main(List<String> args) {
     '--modules=legacy', // TODO(vsm): Change this to use common format.
     '--single-out-file',
     '--inline-source-map',
-    '-p',
+    '--packages',
```
stereotype441: Was this an accidental change?  It doesn't seem like it belongs in this pull request

mit-mit:
@leafpetersen can we merge this?

leafpetersen:
We shouldn't merge it.  I'd rather move this to the language repo.  I'll sort it out next week and then close this.

PR #28310, opened 2017-01-10
tools/create_pkg_manifest.py: Sort manifest entries for reproducibility

joshtriplett:
tools/create_pkg_manifest.py currently outputs manifest entries in hash
order, which results in spurious reordering on updates, making it harder
to see the actual changes.  Sort the manifest entries, to make the
output reproducible.

This will result in a large diff for the next manifest update, but every
subsequent update should have a minimal diff showing only the actual
changes, with no spurious reordering.


joshtriplett:
@zanderso: Does this look reasonable to you?

zanderso:
lgtm and thanks!

PR #28376, opened 2017-01-13
added Range.toString() to dart:html

eukreign:
Added missing `toString()` method to Range.

https://developer.mozilla.org/en-US/docs/Web/API/Range/toString

### DIFF
--- a/sdk/lib/html/dart2js/html_dart2js.dart
+++ b/sdk/lib/html/dart2js/html_dart2js.dart
@@ -27863,6 +27863,10 @@ class Range extends Interceptor {
   @DocsEditable()
   void surroundContents(Node newParent) native;
 
+  @DomName('Range.toString')
+  @DocsEditable()
+  String toString() native;
+
 
   /**
    * Checks if createContextualFragment is supported.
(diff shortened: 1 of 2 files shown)


alan-knight:
@terrylucas 

LGTM.  It would be nice if we had better toStrings in general without needing to hard-code them in templates for specific cases.

alan-knight (commented):
LGTM

asgerf:
I reverted this due to tests failing on dart2js (the build bot turned completely red).

It's not completely clear how the failing tests are affected by this. It could just be bad luck, but we have to get back to a clean state. One of the tests that appear to fail on several browsers is
`co19/LayoutTests/fast/filesystem/file-writer-abort-continue_t01`

Note that the fix also caused [a bunch of tests](https://github.com/dart-lang/sdk/commit/c279a951f490cf4327931f02b5f2b6abf282f640) to start passing, which is great. I hope we can land the fix again soon.

@alan-knight will you try to land it again?
@eukreign if your project relies on this fix, you may want to use commit 226a7e71 until the fix lands again.

eukreign:
What is the latest status on this fix?

So far it looks like my changes where merged, then reverted, than @alan-knight re-submitted a fixed version, then he subsequently reverted that.

Is there anything I can do to help land this?

https://github.com/dart-lang/sdk/commits/master/sdk/lib/html/dart2js/html_dart2js.dart
image

alan-knight:
It caused a number of test problems. I think it needs exception handling for cases where calling the native toString throws, e.g. because the element has been removed, and we're expecting it not to. And I just haven't gotten back to it.

eukreign:
I created an issue to help track this as well. The "merged" status of this pull request is a bit misleading :-D

kevmoo:
Thanks, @eukreign

PR #28469, opened 2017-01-23
Fix "classs" typos

benasocj:


pkg/analyzer/test/src/dart/analysis/driver_test.dart
```diff
@@ -639,7 +639,7 @@ part 'foo.dart';
     provider.newFile(a, 'class A {}');
     provider.newFile(b, "import 'a.dart'; A a;");
     provider.newFile(c, "import 'a.dart'; var a = new A();");
-    provider.newFile(d, "classs A{} A a;");
```
ben4ever: Don't know if this change will have any side effects.

mit-mit:
@a-siva can you take a look please?

floitschG:
I think this is not for Siva but for @bwilkerson

a-siva:
Make sense I will reassign the bug to @bwilkerson as it seems to be in the analysis driver test.

bwilkerson (approved):

mit-mit:
Thanks for the contribution, @ben4ever !

PR #28501, opened 2017-01-24
Fix tools/create_pkg_manifest.py to generate valid XML.

### LINKED ISSUE #28500: tools/create_pkg_manifest.py generates invalid XML
It generates a file that looks like:
```xml
<!-- This file is generated by //dart/tools/create_pkg_manifest.py. DO NOT EDIT -->
<?xml version="1.0" encoding="UTF-8"?>
<manifest>
```
but the processing instruction must be the first thing in an XML file.

I'm working on a trivial CL.

ianloic:
Fixes #28500


ianloic:
cc @zanderso @joshtriplett

zanderso:
lgtm

PR #28554, opened 2017-01-29
StreamIterator docs update with pause/resume behavior

pulyaevskiy:
Hi team!

I just discovered `StreamIterator` class in the SDK and it turned out to be perfect for my use case, so thanks again for awesome work!

I wasn't sure how it behaves in between calls to `moveNext()` though, so I had to go and look how it's implemented in `_StreamIteratorImpl`.

I thought docs can do a better jobs with that so I added a few words in this PR.

### DIFF
--- a/sdk/lib/async/stream.dart
+++ b/sdk/lib/async/stream.dart
@@ -1761,6 +1761,8 @@ abstract class StreamTransformer<S, T> {
  * This wraps a [Stream] and a subscription on the stream. It listens
  * on the stream, and completes the future returned by [moveNext] when the
  * next value becomes available.
+ *
+ * Pauses the stream between calls to [moveNext].
  */
 abstract class StreamIterator<T> {
 
@@ -1776,10 +1778,13 @@ abstract class StreamIterator<T> {
    * Returns a future which will complete with either `true` or `false`.
    * Completing with `true` means that another event has been received and
    * can be read as [current].
-   * Completing with `false` means that the stream itearation is done and
+   * Completing with `false` means that the stream iteration is done and
    * no further events will ever be available.
    * The future may complete with an error, if the stream produces an error,
    * which also ends iteration.
+   *
+   * The stream is paused when the future completes with `true` and
+   * resumed on the next call to `moveNext`.
    *
    * The function must not be called again until the future returned by a
    * previous call is completed.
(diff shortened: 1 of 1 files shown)


lrhn:
I recommend also looking at `StreamQueue` in `package:async`, it's really a better version of `StreamIterator`.

sdk/lib/async/stream.dart
```diff
@@ -1776,11 +1778,14 @@ abstract class StreamIterator<T> {
...
    * can be read as [current].
-   * Completing with `false` means that the stream itearation is done and
+   * Completing with `false` means that the stream iteration is done and
    * no further events will ever be available.
    * The future may complete with an error, if the stream produces an error,
    * which also ends iteration.
    *
+   * The stream is paused when the future completes with `true` and
+   * resumed on the next call to `moveNext`.
+   *
```
lrhn: This is almost true - if you call `moveNext` immediately in the  callback of the previous `moveNext` calls's future, then it doesn't pause at all.
Otherwise it's correct. The current behavior is a recent change, previous to that, the iterator had a one-element cache, and it didn't pause as long as you asked for the next `moveNext` before the event occurred. Now it pauses unless you call `moveNext` immediately - but that happens in many cases anyway, and the code is much simpler.

pulyaevskiy:
`StreamQueue` looks very promising, thanks for the hint, @lrhn !
I'll update `moveNext()` documentation to reflect those details you mentioned.

lrhn:
Thinking about it more, I'm actually not sure giving too much detail a good thing. That locks the implementation into its current choices. If the previous implementation was documented in detail, the change in behavior would have been breaking. 
I think I would prefer to just say that the stream iterator *may* pause the source stream while it's waiting for the next call to `moveNext`, and let it be an implementation decision if and when to actually pause.

pulyaevskiy:
@lrhn I've pushed some changes according to your comments.

Here is a strange thing though.
If I use `StreamController` I can react on `onPause` and `onResume` events and that's great. With regular stream subscription I can control this flow.
However, there is no way to know if `StreamIterator` would ever pause the stream. It encapsulates stream subscription and does not provide access to pause/resume functionality.

lrhn:
True, it's an abstraction. You don't pause or resume manually, instead you ask for action or not, and the underlying implementation decides for you whether to pause or not.

If you stop asking for events, it will eventually pause the source stream. More precision that that is taking choices away from the implementation, and those are *implementation* choices that are not important to users of the abstraction.

Issue #22, opened 2011-10-10
Support non-nullable types.


DartBot:
_**Admin comment**_: This is activaly being worked on, see https://github.com/dart-lang/language/issues/110 and https://dart.dev/null-safety

_This issue was originally filed by cflew...&#064;gmail.com_

---

Short version: Null pointers are a really good way to mess up a program at runtime, and I'd like the Dart team to reevaluate whether they're absolutely required.

Slightly longer version: I would say the #­1 cause of issues in my programs (excluding logical errors/requirements errors) are NPEs. Having a language support NPE removal, be it via some clever compiler warning or simply removing null altogether, would be wonderful. I'm personally partial to Scala's method of null removal, but I'm sure PL gurus like yourselves have seen many others.

Dart has a stated goal of avoiding the creation of programs that &quot;are difficult to debug or maintain.&quot; NPEs are a huge pain point in this regard. I'd be really happy if the Dart team reevaluated whether they are absolutely required to achieve the other aims.

sethladd:
FYI a relevant DEP is https://github.com/dart-lang/dart_enhancement_proposals/issues/30

1and1get2:
I'd really love to see this get implemented, or better yet, kotlin/swift style optional variable support

larssn:
So the latest Google I/O talks about this finally being implemented: [Non-Nullable Types](https://youtu.be/J5DQRPRBiFI?t=1989).

Is this part of any milestone yet? If so can we have it added to this thread?

It's a feature I'm very excited about, so I'm hoping its not too far off still.

lrhn:
No milestone yet. It's a big feature, with many small parts and necessary side-features, not all of which are designed yet, so all we can say is that we are working on it, and you can look at the progress in the [language issue tracker](https://github.com/dart-lang/language/issues?q=is%3Aissue+label%3Annbd).

MihaMarkic:
I assume quite a huge effort, at least at looking at C# 8 which is introducing the same feature. Keeping fingers crossed.

speller:
Hi guys,
[...]

PR #28619, opened 2017-02-03
Proposal for non-nullable types

munificent:
This is the living proposal and tracking bug for adding **non-nullable types** (#22) to Dart. Two important caveats to keep in mind:

1. **We have not committed to shipping this.** We may discover that the overall
   usability impact is negative and Dart is better without them.

2. Because this involves **breaking changes,** it would not ship until Dart 2.0.

## Prototype

In order to assess the usability, we're going to do a prototype first to see how it feels to convert a decent chunk of code (include our core libraries) to use non-nullable types.

The prototype lives on this branch: https://github.com/dart-lang/sdk/tree/prototype-nnbd

### Implement static checking rules for non-nullable types

- [x] Make types default to non-nullable.
- [x] Report errors on state that is non-nullable and not initialized.
- [x] Use `?` to define a nullable type.
- [x] Do the right type relations between nullable and non-nullable types.
- [x] Report errors when downcasting from a nullable to a non-nullable type.
- [x] Report errors on non-nullable optional parameters without default values.
- [x] Test for errors when passing `null` or a non-nullable type to a non-nullable parameter.
- [x] Decide on and implement the covariance behavior we want between nullable
      types. (Is `int?` subtype of `num?`? Is `int` it a subtype of `num?`?)
- [x] Incorporate nullable types into least upper bound calculation.
- [x] Figure out how to handle nullable types in conditions (`if (c)`, `c ? ... : ...`, `c && ...`, etc.)
- [x] Define what methods are available on an object of nullable type.

### Implement analysis rules to work with nullable types

- [x] Allow nullable types as receivers for null-aware operators.
- [x] Support `== null` and `!= null` in type promotion.
- [x] Handle `is` on nullable and non-nullable types in type promotion.
- [ ] Support `??` in type promotion. If the LHS side is a local variable, it should promote to non-nullable in the RHS.
- [x] Remove the nullity of the LHS when determining the type of a `??` expression.
- [x] Make sure `as` handles non-nullable types correctly. (In the analyzer.
      The runtime implementation is a different thing.)
- [ ] Handle a nullable type on the LHS of `??=`.
- [ ] Warn on pointless check for null on non-nullable types (`nonNullable == null` or `nonNullable != null`).

### Convert core libraries to be null-safe

- [x] Generate report of static errors in core libraries.
- [ ] Figure out if we want `Map[]` to return a nullable value.
- [ ] Figure out how to handle `new List(size)`.
- [ ] Fix the code.

### Determine which extensions to the proposal we want to try

- [ ] Non-null assertion operator.
- [ ] Non-null top type above Object.
- [ ] Definite assignment analysis to allow uninitialized non-nullable
        local variables.

### Convert some framework and library code to be null-safe

- [ ] AngularDart
- [ ] Flutter
- [ ] Mockito?

### Convert an application or two to be null-safe

**TODO: Which applications?**

Once we get far enough through this, we should be able to confidently decide if
non-nullable types are right for Dart. Assuming they are, we move to:

## Real implementation

This is a big feature with a lot of moving parts. I don't have a detailed plan
of everything that needs to be done, but I will progressively refine this as we
figure that out. The current plan is:

- [ ] Parse `?` in type annotations. (#27231)
- [ ] Add support for disabling non-null by default on a per-library/class basis (See [G.2. Migration aids](https://github.com/chalin/DEP-non-null/blob/master/doc/dep-non-null-AUTOGENERATED-DO-NOT-EDIT.md#g2-migration-aids)).

**TODO: Fill in more here.**

### DIFF
--- a/docs/language/informal/non-nullable-types.md
+++ b/docs/language/informal/non-nullable-types.md
@@ -1,0 +1,1017 @@
+# Non-nullable types
+
+Owner: **Bob Nystrom** (rnystrom@google.com)
+
+This is a proposal to extend Dart's type system to express types that can or
+cannot allow `null`.
+
+We know it is *technically feasible* to implement non-nullable types—other
+languages have. The hard questions are around *usability*—can we design
+something that isn't too painful to use and helps users write correct code? Will
+it still feel like Dart? Usability can't survive in a vacuum, so we will start
+with this informal proposal and progressively refine it as we prototype and
+experiment.
+
+Before I go any farther, a shout out to Patrice's amazing [non-nullable by
+default DEP][dep]. Almost all of the ideas here can be found there, and we
+wouldn't even be considering non-nullability in Dart if he hadn't written that
+up. You can look at this document as a distillation of his DEP with some of the
+open questions answered.
+
+[dep]: https://github.com/dart-lang/dart_enhancement_proposals/issues/30
+
+## Motivation
+
+Normally, there'd be a bunch of verbiage motivating this. But, now that Flow,
+Hack, Swift, Kotlin, Ceylon, and TypeScript have non-nullable types, most people
+are pretty familiar with the pros and cons.
+
+**TODO: Fill in more if readers want.**
+
+## Summary for executives who like bullet points
+
+*   By default, types are *non*-nullable.
+
+*   To make a type nullable, append `?`.
+
+*   A nullable type behaves like a [union type][] of the base type and the
+    `Null` class.
+
+[union type]: https://en.wikipedia.org/wiki/Union_type
+
+*   You must explicitly cast a nullable type to its non-nullable base type.
+    Since it is explicit **you can see in your code every place a `null` can go
+    where it is not expected.**
+
+*   A nullable type does not support the methods of the base type. No
+    `string.length` if `string` is a nullable String. This means **you will
+    never get a NoSuchMethodError from inadvertantly calling a method on
+    `null`.**
+
+## Nullable types as union types
+
+We'll start from the underlying semantic model and work our way up to the user
+visible surface features. Nullable types are built on two concepts, the *Null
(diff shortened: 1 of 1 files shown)


whesse:
Can we put this document someplace other than on an unmerged branch?  Unless we have some sort of system for tracking which branches have important information, but shouldn't be merged, then we don't know which branches besides master, dev, and stable are important.

munificent:
Yeah, we can figure out some different process for tracking in-progress proposals. At the time that we decided on this, we were actually using GitHub. I guess we'll need to do something different now.

1and1get2:
any update on this? would love better nullability handling like swift/kotlin

munificent:
Sorry, no real update. We've been very focused on getting everything onto strong mode and getting Dart 2 shipped. That's a big migration with a lot of breakage, so it's consumed most of our time and most over our user's capacity to absorb breakage.

There is still a lot of hope we can get to non-nullable types before too long, but we'll need to rest a bit and give our customers a little breathing room after the migration we've dragged most of them through. For obvious reasons, adding non-nullable types will cause a *lot* of breakage and migration cost, so we want to be sensitive to user pain tolerance. :)

kirbyfan64:
Just break everyone's code every other month, that way they get used to the breakage!

munificent:
> Just break everyone's code every other month, that way they get used to the breakage!

Yes! The beatings will continue until morale improves.

mit-mit:
@munificent can we merge this proposal?

munificent:
Oh, wow, I'd completely forgotten that I wrote this up.

If you'd like to use this as a starting for non-nullable types, sure. I don't think we want it in this repo, though. Want me to move it over to the language repo?

munificent:
This has been superceded by [a new proposal](https://github.com/dart-lang/language/issues/110).

Issue #28707, opened 2017-02-09
standalone/io/process_stdin_transform_unsubscribe_test timeout on Windows


nex3:
See [this](https://build.chromium.org/p/client.dart/builders/vm-win-debug-x64-be/builds/3113/steps/checked%20vm%20tests/logs/stdio), [this](https://build.chromium.org/p/client.dart/builders/vm-win-release-ia32-be/builds/5402/steps/checked%20vm%20tests/logs/stdio), and [this](https://build.chromium.org/p/client.dart/builders/vm-win-release-x64-be/builds/5149/steps/checked%20vm%20tests/logs/stdio) log. Output looks like:

```
FAILED: none-vm-checked release_x64 standalone/io/process_stdin_transform_unsubscribe_test
Expected: Pass 
Actual: Timeout
CommandOutput[vm]:
diagnostics: Using E:\b\depot_tools\win_toolchain\vs_files\d5dc33b15d1b2c086f2f6632e2fd15882f80dbd3\win_sdk\Debuggers\x64\cdb.exe to print stack traces
    
    Microsoft (R) Windows Debugger Version 10.0.14321.1024 AMD64
    Copyright (c) Microsoft Corporation. All rights reserved.
    
    *** wait with pending attach
    Symbol search path is: srv*
    Executable search path is: 
    ModLoad: 00000001`3f670000 00000001`3fd91000   E:\b\build\slave\vm-win-release-x64-be\build\sdk\out\ReleaseX64\dart.exe
    ModLoad: 00000000`76f60000 00000000`77108000   C:\Windows\SYSTEM32\ntdll.dll
    ModLoad: 00000000`76d40000 00000000`76e5f000   C:\Windows\system32\kernel32.dll
    ModLoad: 000007fe`fcd80000 000007fe`fcdec000   C:\Windows\system32\KERNELBASE.dll
    ModLoad: 000007fe`fa580000 000007fe`fa5a7000   C:\Windows\system32\IPHLPAPI.DLL
    ModLoad: 000007fe`fded0000 000007fe`fdf6f000   C:\Windows\system32\msvcrt.dll
    ModLoad: 000007fe`fd8e0000 000007fe`fd8e8000   C:\Windows\system32\NSI.dll
    ModLoad: 000007fe`fa560000 000007fe`fa56b000   C:\Windows\system32\WINNSI.DLL
[...]

PR #28708, opened 2017-02-09
[Gardening] Mark a test as flaky.

nex3:
standalone/io/process_stdin_transform_unsubscribe_test was flakily
timing out on Windows.

See #28707

### DIFF
--- a/tests/standalone/standalone.status
+++ b/tests/standalone/standalone.status
@@ -300,6 +300,9 @@ verbose_gc_to_bmu_test: Skip
 io/platform_test: RuntimeError # Expects to be running from 'dart' instead of 'dart_precompiled_runtime'
 io/directory_list_sync_test: Timeout, Skip # Expects to find the test directory relative to the script.
 
+[ $runtime == vm && $system == windows ]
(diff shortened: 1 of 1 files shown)


whesse (approved):
LGTM

fsc8000:
Not lgtm.

kasperl:
I'm completely with @nex3 on this one. We need to get the bots consistently green and we need to aggressively mark tests as flaky in the status files to achieve that. I understand that it is valuable to you @fsc8000 to see the bots fail from time to time as you track down the issue, but we can't use the build bots on the main waterfall for tracking down issues if it hurts the velocity of the rest of the team. 

Could we fix the issue by letting the logs clearly show the outcome of running the test before it matches that up to the expectations? Something like this perhaps @whesse:

```Done dart2js-d8-checked release_ia32 lib/mirrors/constructor_kinds_test/01: timed out expectedly -> pass```

If this "discussion" is in any way related to this issue only occurring on Windows and that we don't have enough developer workstations that run that particular OS, then it might be worth pointing out that I'm very supportive of ordering additional workstations.

Let's try to update our gardening documentation to make it clear that gardeners are empowered and encouraged to update status files and revert changes to make the main waterfall build bots green.

Issue #28712, opened 2017-02-09
corelib/list_unmodifiable_test is flaky with dart2js


nex3:
See [this log](https://build.chromium.org/p/client.dart/builders/dart2js-linux-drt-csp-minified-be/builds/2258/steps/dart2js%20drt%20tests/logs/stdio) for example.

```
FAILED: dart2js-drt release_x64 corelib/list_unmodifiable_test
Expected: Pass 
Actual: RuntimeError
CommandOutput[dart2js]:

stdout:
Dart file (/b/build/slave/dart2js-linux-drt-csp-minified-be/build/sdk/tests/corelib/list_unmodifiable_test.dart) compiled to JavaScript: out/ReleaseX64/generated_compilations/dart2js-minified-csp-sdk/tests_corelib_list_unmodifiable_test/test.js

CommandOutput[content_shell]:

stdout:
#READY
CONSOLE MESSAGE: line 78: {
  "type": "debug",
  "value": "test_controller.js started",
  "timestamp": 0
}
CONSOLE MESSAGE: line 78: {
  "type": "print",
  "value": "dart-calling-main",
  "timestamp": 0.132
}
CONSOLE MESSAGE: line 78: {
  "type": "sync_exception",
[...]

PR #28713, opened 2017-02-09
[Gardening] Mark a test as flaky on Dart2js.

nex3:
corelib/list_unmodifiable_test fails flakily on dart2js.

See #28712


sigmundch:
lgtm

Issue #28714, opened 2017-02-09
kernel/test/baseline_type_propagation_test times out on Mac OS


nex3:
See [this log](https://build.chromium.org/p/client.dart/builders/pkg-mac10.11-release-be/builds/2693/steps/package%20unit%20tests/logs/stdio).

```
FAILED: none-vm-checked release_x64 pkg/kernel/test/baseline_type_propagation_test
Expected: Pass 
Actual: Timeout
CommandOutput[vm]:

stdout:
00:00 +0: argument.dart
00:12 +1: arithmetic.dart
00:14 +2: async_function.dart
00:17 +3: bad_store.dart
00:19 +4: call.dart
00:21 +5: closure.dart
00:23 +6: covariant_generic.dart
00:25 +7: DeltaBlue.dart
00:27 +8: escape.dart
00:28 +9: external.dart
00:31 +10: fallthrough.dart
00:33 +11: micro.dart
00:35 +12: named_parameters.dart
00:37 +13: null_aware.dart
00:39 +14: optional.dart
00:41 +15: override.dart
00:43 +16: prefer_baseclass.dart
00:45 +17: redirecting_factory.dart
00:47 +18: static_setter.dart
00:49 +19: store_load.dart
00:51 +20: stringliteral.dart
00:53 +21: uninitialized_fields.dart
00:56 +22: unused_methods.dart
00:58 +23: void-methods.dart
01:01 +24: All tests passed!


Command[vm]: DART_CONFIGURATION=ReleaseX64 xcodebuild/ReleaseX64/dart-sdk/bin/dart --enable_asserts --enable_type_checks --ignore-unrecognized-flags --packages=/b/build/slave/pmr-be/build/sdk/.packages /b/build/slave/pmr-be/build/sdk/pkg/kernel/test/baseline_type_propagation_test.dart
Took 0:01:02.154351

Short reproduction command (experimental):
    python tools/test.py -mrelease --checked -t60 --use-sdk pkg/kernel/test/baseline_type_propagation_test
```

nex3:
I'm just going to mark this as slow.

PR #28715, opened 2017-02-09
[Gardening] Mark a test as timing out on OS X.

nex3:
kernel/test/baseline_type_propagation_test was flakily timing out on the
OS X bots.

See #28714


whesse:
Replace this (or change it to) mark the test as slow.

nex3:
Done.

whesse (approved):

Issue #28722, opened 2017-02-10
co19/LayoutTests/fast/css/fontfaceset-download-error_t01 flakily times out


nex3:
I've seen this multiple times on `dart2js-mac10.11-chrome-be`. See [this log](https://build.chromium.org/p/client.dart/builders/dart2js-mac10.11-chrome-be/builds/2771/steps/dart2js%20chrome%20co19%20fast-startup%20tests/logs/stdio).

```
FAILED: dart2js-chrome release_x64 co19/LayoutTests/fast/css/fontfaceset-download-error_t01
Expected: RuntimeError 
Actual: Timeout
CommandOutput[dart2js]:

stdout:
Dart file (/b/build/slave/10.11-c-be/build/sdk/tests/co19/src/LayoutTests/fast/css/fontfaceset-download-error_t01.dart) compiled to JavaScript: xcodebuild/ReleaseX64/generated_compilations/dart2js-sdk/tests_co19_src_LayoutTests_fast_css_fontfaceset-download-error_t01/test.js

CommandOutput[chrome]:

stdout:
message:
[...]

PR #28723, opened 2017-02-10
[Gardening] Mark a co19 test as flaky on dart2js.

nex3:
LayoutTests/fast/css/fontfaceset-download-error_t01 is usually fast, but
sometimes times out.

See #28722


sigmundch:
lgtm

PR #28795, opened 2017-02-16
Enable generic method parsing by default

kevmoo:
Running tests locally now...

### DIFF
--- a/pkg/analyzer/pubspec.yaml
+++ b/pkg/analyzer/pubspec.yaml
@@ -1,5 +1,5 @@
 name: analyzer
-version: 0.29.6
+version: 0.29.7
 author: Dart Team <misc@dartlang.org>
 description: Static analyzer for Dart.
 homepage: https://github.com/dart-lang/sdk/tree/master/pkg/analyzer
@@ -12,7 +12,7 @@ dependencies:
   crypto: '>=1.1.1 <3.0.0'
   glob: ^1.0.3
   html: '>=0.12.0 <1.14.0'
-  isolate: ^0.2.2
+  isolate: '>=0.2.2 <2.0.0'
   meta: ^1.0.2
   package_config: '>=0.1.5 <2.0.0'
   path: '>=0.9.0 <2.0.0'
--- a/pkg/analyzer/CHANGELOG.md
+++ b/pkg/analyzer/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.29.7
+
+* Enabled support for generic method syntax by default in the parser.
+* Support the latest version of `pkg/isolate`.
+
 ## 0.29.2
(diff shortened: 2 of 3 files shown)


pkg/analyzer/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: analyzer
-version: 0.29.6
+version: 0.29.6+1
```
bwilkerson: I don't understand the "+1". I was planning on making it "0.29.7". Can you explain the difference?

bwilkerson (approved):
This looks fine, as far as it goes, but Phil also needs to make a small change before we actually publish (unless we can't wait and have to publish twice).

pkg/analyzer/pubspec.yaml
kevmoo: I guess it could/should be `.7`

There is no API change, but there is a behavior change.

kevmoo:
there are 7 failed tests (11693 passed)

All in the form

```
FAIL: analysis engine | generated tests | DirectoryBasedDartSdkTest | test_[fromFile_XYZ | getXYZ]
  Expected: not null
    Actual: <null>
```

pkg/analyzer/pubspec.yaml
kevmoo: Sent an update

bwilkerson:
Did the tests pass before your change?

kevmoo:
> Did the tests pass before your change?

You ask good questions!

kevmoo:
@bwilkerson Same failures before/after.

pq:
> This looks fine, as far as it goes, but Phil also needs to make a small change before we actually publish (unless we can't wait and have to publish twice).

Yes.  Please!

Could you change the constraint on `isolate`?

    isolate: '>=0.2.2 <2.0.0'

Thanks!

bwilkerson (approved):
In that case, I'm going to guess that some previous change caused the break, and I'd say you should go ahead and merge this (possibly after addressing Phil's ask).

kevmoo:
@pq @bwilkerson Reran tests with the latest isolate pkg. No new failures. I think we're good. PTAL

pkg/analyzer/pubspec.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   crypto: '>=1.1.1 <3.0.0'
   glob: ^1.0.3
   html: '>=0.12.0 <1.14.0'
-  isolate: ^0.2.2
+  isolate: '>=0.2.2 <2.0.0'
```
pq: \o/

Thanks!

pq (approved):

bwilkerson:
lgtm

PR #28971, opened 2017-03-06
Fix typo in the @mustCallSuper error message

clairesarsam-wf:
# Ultimate Problem

A library my team maintains ([react-dart](https://github.com/cleandart/react-dart)) recently added several `@mustCallSuper` annotations. One of our consumers noticed that the error message produced states that the annotated method 'does invoke the overridden method', instead of the opposite.

# Solution

Fix the typo in the error message.


clairesarsam-wf:
I signed it!

kevmoo:
Thanks for the fix, @clairesarsam-wf – will let @bwilkerson run with it

bwilkerson (approved):

Issue #29012, opened 2017-03-09
standalone/io/http_server_response_test is flakily crashing


nex3:
From [this vm-linux-debug-simmips-be log](https://build.chromium.org/p/client.dart/builders/vm-linux-debug-simmips-be/builds/3301/steps/checked%20vm%20tests/logs/stdio):

```
FAILED: none-vm-checked debug_simmips standalone/io/http_server_response_test
Expected: Pass 
Actual: Crash
CommandOutput[vm]:

stderr:
../../runtime/vm/scavenger.cc: 239: error: expected: to_->Contains(RawObject::ToAddr(obj))
Dumping native stack trace for thread d39
Aborting re-entrant request for stack trace.
../../runtime/vm/os_thread_linux.cc: 401: error: pthread error: 35 (Resource deadlock avoided)
Aborting re-entrant request for stack trace.


Command[vm]: DART_CONFIGURATION=DebugSIMMIPS out/DebugSIMMIPS/dart --enable_asserts --enable_type_checks --short_socket_write --ignore-unrecognized-flags --packages=/mnt/data/b/build/slave/vm-linux-debug-simmips-be/build/sdk/.packages /mnt/data/b/build/slave/vm-linux-debug-simmips-be/build/sdk/tests/standalone/io/http_server_response_test.dart
Took 0:00:06.209282

Short reproduction command (experimental):
    python tools/test.py -asimmips --checked -t480 standalone/io/http_server_response_test
```

mraleph:
This issue has been last updated 252 days ago. Please consider investigating whether it is still meaningful and if not close it. If the issue is still meaningful please find a relevant owner. Thank you.

PR #29013, opened 2017-03-09
Mark an IO test as flaky on simmips.

nex3:
See #29012


rmacnak-google (approved):
lgtm

(But please remember the SDK uses Rietveld)

efortuna (approved):

PR #29297, opened 2017-04-06
Account for alternate "SDK" layouts.

pylaligand:
The sources for standard Dart libraries are coming via an embedder (Flutter) in Fuchsia, with a slightly different tree layout.

### DIFF
--- a/pkg/analysis_server/lib/src/plugin/plugin_watcher.dart
+++ b/pkg/analysis_server/lib/src/plugin/plugin_watcher.dart
@@ -124,7 +124,8 @@ class PluginWatcher implements DriverWatcher {
   String _getSdkPath(AnalysisDriver driver) {
     AbsolutePathContext context = resourceProvider.absolutePathContext;
     String sdkRoot = driver.sourceFactory.forUri('dart:core').fullName;
-    while (sdkRoot.isNotEmpty && context.basename(sdkRoot) != 'lib') {
+    while (sdkRoot.isNotEmpty && context.basename(sdkRoot) != 'lib' &&
+        context.basename(sdkRoot) != 'dart_sdk') {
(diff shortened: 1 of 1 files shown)


pylaligand:
On Fuchsia we get the Dart standard libraries via Flutter. `core.dart` is located in a tree that looks like:
```
base
  |
  |------- lib
  |          |------- embedder.yaml
  |
  |------- dart_sdk
  |          |------- async
  |          |------- core
  |                     |------- core.dart
  |          |------- etc...
```
When trying to locate the root of the SDK, the while loop would hit `/` as there's never a `lib` folder in `core.dart`'s path and then keep looping forever. The result is that the analysis server would remain stuck and not give any results.

This is a quick fix, but I'm open to other suggestions. For example I think we could put all the stuff that's in `dart_sdk` under `lib` so that the layout resembles that of the standard Dart SDK.

/cc @abarth

abarth:
We can change the layout in flutter to whatever is useful.  I don't think we've given any thought to the current layout.

bwilkerson:
Unless there is a strong need to deviate from the standard SDK layout, I would strongly urge you to not change it. Even then, I would encourage to talk to the Dart team so that we can see whether we can change the standard SDK layout to meet both sets of needs. Having to special case things like this makes it harder for us to test everything thoroughly.

bwilkerson (approved):
Does this need to be cherry-picked into 1.23? If so, please file a merge request.

pylaligand:
I'll make the change on the Flutter side to match the standard layout better.

In any case, there's no need to cherry-pick anything as we build our Dart stuff from sources in Fuchsia.

pylaligand:
https://github.com/flutter/engine/pull/3575

PR #29301, opened 2017-04-07
Wrong heap space data in Observatory

deftomat:
Fixed situation when compacted heap space contains a wrong data (new space instead of an old space)

image

-------------------------

image


johnmccutchan:
Thanks!

a-siva (commented):
lgtm

PR #29364, opened 2017-04-17
GN workspaces now use packages files.

pylaligand:
The previous version would use a global cache of package locations which would hide missing dependencies.
Also fixed an issue in file tracking with non-existing sources that resolvers such as PackageMapUriResolver sometimes return.

### DIFF
--- a/pkg/analyzer/lib/src/generated/gn.dart
+++ b/pkg/analyzer/lib/src/generated/gn.dart
@@ -8,136 +8,15 @@ import 'dart:collection';
 import 'dart:core';
 
 import 'package:analyzer/file_system/file_system.dart';
+import 'package:analyzer/source/package_map_resolver.dart';
 import 'package:analyzer/src/generated/sdk.dart';
 import 'package:analyzer/src/generated/source.dart';
 import 'package:analyzer/src/generated/source_io.dart';
 import 'package:analyzer/src/generated/workspace.dart';
+import 'package:package_config/packages.dart';
+import 'package:package_config/packages_file.dart';
+import 'package:package_config/src/packages_impl.dart';
 import 'package:path/path.dart';
-
-/**
- * Similar to Map#putIfAbsent, except that a value is stored only if it is not
- * null.
- */
-V _putIfNotNull<K, V>(Map<K, V> map, K key, V ifAbsent()) {
-  if (map.containsKey(key)) {
-    return map[key];
-  }
-  V computed = ifAbsent();
-  if (computed != null) {
-    map[key] = computed;
-  }
-  return computed;
-}
-
-/**
- * The [UriResolver] used to resolve `file` URIs in a [GnWorkspace].
- */
-class GnFileUriResolver extends ResourceUriResolver {
-  /**
-   * The workspace associated with this resolver.
-   */
-  final GnWorkspace workspace;
-
-  /**
-   * Initialize a newly created resolver to be associated with the given
-   * [workspace].
-   */
-  GnFileUriResolver(GnWorkspace workspace)
-      : workspace = workspace,
-        super(workspace.provider);
-
-  @override
-  Source resolveAbsolute(Uri uri, [Uri actualUri]) {
-    if (!ResourceUriResolver.isFileUri(uri)) {
-      return null;
-    }
-    String path = provider.pathContext.fromUri(uri);
-    File file = workspace.findFile(path);
-    if (file != null) {
-      return file.createSource(actualUri ?? uri);
-    }
-    return null;
-  }
-}
-
-/**
- * The [UriResolver] used to resolve `package` URIs in a [GnWorkspace].
- */
-class GnPackageUriResolver extends UriResolver {
-  /**
-   * The workspace associated with this resolver.
-   */
-  final GnWorkspace workspace;
-
-  /**
-   * The path context that should be used to manipulate file system paths.
-   */
-  final Context pathContext;
-
-  /**
-   * The cache of absolute [Uri]s to [Source]s mappings.
-   */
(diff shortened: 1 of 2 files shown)


pkg/analyzer/lib/src/generated/gn.dart
```diff
@@ -8,202 +8,126 @@ import 'dart:collection';
...
 
   /**
-   * The path context that should be used to manipulate file system paths.
+   * The path to the directory with source locations.
+   *
+   * Each file in this directory is named after a Dart package and contains the
+   * path to the package's sources.
    */
-  final Context pathContext;
+  String _packagesFilePath;
```
bwilkerson: The comment should be updated. Based on the name I'm guessing this is the path to a file, not a directory.

pkg/analyzer/lib/src/generated/gn.dart
```diff
@@ -8,202 +8,126 @@ import 'dart:collection';
...
-      return null;
-    }
-
-    String sourcePath =
-        context.relative(path, from: workspace.packages[package]);
-
-    return Uri.parse('package:$package/$sourcePath');
+  Map<String, List<Folder>> get packageMap {
+    _packageMap ??= _convertPackagesToMap(packages);
+    return _packageMap;
```
bwilkerson: Consider re-writing to the shorter form of
```
Map<String, List<Folder>> get packageMap => _packageMap ??= _convertPackagesToMap(packages);
```
and similarly for `packages`
```
Packages get packages => _packages ??= _createPackages();
```

pkg/analyzer/lib/src/generated/gn.dart
```diff
@@ -243,26 +155,56 @@ class GnWorkspace extends Workspace {
     return null;
   }
 
-  /**
-   * Locate the Dart sources directory.
-   *
-   * Return `null` if it could not be found.
-   */
-  static String _getPackagesDirectory(ResourceProvider provider, String root) {
+  static String _findPackagesFile(
```
bwilkerson: Is there some documentation we can reference here that explains the structure that's being traversed, why there are (or can be) multiple .packages files, and why it's reasonable to choose one at random? I'm concerned about having to maintain this code when I don't understand why it works the way it does.

pkg/analyzer/lib/src/generated/gn.dart
```diff
@@ -243,26 +155,56 @@ class GnWorkspace extends Workspace {
...
         .firstWhere((Folder folder) {
-      String baseName = basename(folder.path);
+      String baseName = pathContext.basename(folder.path);
       // TODO(pylaligand): find a better way to locate the proper directory.
       return baseName.startsWith('debug') || baseName.startsWith('release');
-    }, orElse: () => null)?.path;
-    return outDirectory == null
-        ? null
-        : provider.pathContext.join(outDirectory, 'gen', 'dart.sources');
+    }, orElse: () => null);
```
bwilkerson: This particular sequence of invocations appears in three places. Given that we control the API of Folder, we should add some utility methods that do some of this work rather than repeating it. (Not a request for a change to this CL, though.)

bwilkerson:
Adding @scheglov for a review of the change in `file_state.dart`.

bwilkerson (approved):
Sorry about the delay, still catching up from being out sick.

pkg/analyzer/lib/src/dart/analysis/file_state.dart
```diff
@@ -716,7 +716,7 @@ class FileSystemState {
 
       // If the URI cannot be resolved, for example because the factory
       // does not understand the scheme, return the unresolved file instance.
-      if (uriSource == null) {
+      if (uriSource == null || !uriSource.exists()) {
```
scheglov: I don't understand the need for this change.
It seems wrong to me.

The URI is either can be resolved and is always resolved to this file (until the `SourceFactory`, `FileSystemState` and `AnalysisDriver` recreation), or it cannot be resolved at all and `SourceFactory` returns `null`. It's OK for the file to not exist, we handle this OK in `FileState.refresh()` and use the empty string as its content. For Bazel if there is no corresponding generated file, we default the URI resolution to the corresponding file in the `lib` folder of the corresponding package.

scheglov (changes requested):

pkg/analyzer/lib/src/dart/analysis/file_state.dart
scheglov: Ah, now I understand why you did this change.
`PackageMapUriResolver` was [radically changed](https://github.com/dart-lang/sdk/commit/6fe8e2278b2cb678ad016b6151c12d016bc7a7db).
So, now it returns either a file, or `null`.
And your change for `file_state.dart` is not needed :-)

pylaligand: Reverted.

pkg/analyzer/lib/src/generated/gn.dart
pylaligand: Done.

pkg/analyzer/lib/src/generated/gn.dart
pylaligand: Done.

pkg/analyzer/lib/src/generated/gn.dart
pylaligand: Added some documentation.

pkg/analyzer/lib/src/generated/gn.dart
pylaligand: Ack!

scheglov (approved):
No change for file_state.dart LGTM :-)

bwilkerson (approved):

PR #29371, opened 2017-04-18
Local type inference informal specification

leafpetersen:


leafpetersen:
This is work in progress, putting up here for discussion and collaboration.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+
+The primary function of expression inference is to determine the parameter and
+return types of closure literals which are not explicitly annotated, and to fill
+in elided type variables in constructor calls, generic method calls, and generic
+literals.
+
+### Expectation contexts
+
+A typing expectation context (written using the meta-variables `J` or `K`) is
+either a type schema `P`, or an empty context `_`.
```
stereotype441: Does an empty context behave differently from a type schema of `?` ?

If so, what is the rationale?  If not, can we simplify the exposition by saying `?` instead of `_` everywhere below?

stereotype441 (commented):
Still digesting this

docs/language/informal/local-inference.md
leafpetersen: I think this might be true.  Type schemas just landed recently, so I had developed the infrastructure using empty contexts, but it seems likely that we could treat a context of `?` and a context of `_` equally.  I'll need to go through and check the details though.  I don't think it's 100% equivalent, but it might be better.  I think as currently phrased, this example demonstrates a difference:
```dart
foo<T>(List<T> x) {}
foo([(b) ? 1 : 2]) // The conditional expression is checked at `?` and hence inferred as Object
var x = (b) ? 1 : 2 // The conditional expression is checked at `_` and hence inferred as int
```
That's probably not a feature though.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+- `C0 + C1` is the concatenation of constraint lists `C0` and `C1`.
+
+### Rules
+
+- The unknown type `?` is a subtype match for any type `Q` with no constraints.
+- Any type `P` is a subtype match for the unknown type `?` with no constraints.
+- A type variable `T` in `L` is a subtype match for any type schema `Q`:
+  - Under constraint `T <: Q`.
+- A type schema `Q` is a subtype match for a type variable `T` in `L`:
+  - Under constraint `M <: Q`.
```
stereotype441: Do you mean `Q <: T` here?

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+- `P` is a subtype match for `FutureOr<Q>` with respect to `L` under constraints
+  `C`:
+  - If `P` is a subtype match for `Future<Q>` with respect to `L` under
+    constraints `C`.
+  - Or `P` is not a subtype match for `Future<Q>` with respect to `L` under
+    constraints `C`
+    - And `P` is a subtype match for `Q` with respect to `L` under constraints
+      `C`
+- A type variable `T` not in `L` with bound `P` is a subtype match for the same
+type variable `T` with bound `Q` with respect to `L` under constraints `C`:
```
stereotype441: I don't understand this rule.  How can the same type variable have two different bounds?

leafpetersen: If you have a type variable `T extends A`, and you do an instance check on something of that type (subject to the usual promotion restrictions):

```dart
   if (x is B) {
     ...
   }
```
then within the scope of the instance check (the `...`) we treat `x` as having type `T extends B`.  You might have other expressions of type `T` in the same scope though, which still have the old bound.

docs/language/informal/local-inference.md
leafpetersen: yes, thanks.

docs/language/informal/local-inference.md
stereotype441: Ruh roh, I can't do this right now because kernel doesn't currently have a way to represent a type variable whose bound is different due to type promotion.  Let's talk with @kmillikin tomorrow when I'm in Seattle about how we want to represent this in kernel.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+  - If `P` is a subtype match for `Q` with respect to `L` under constraints `C`.
+- A type variable `T` not in `L` with bound `P` is a subtype match for a type
+`Q` with respect to `L` under constraints `C`:
+  - If `P` is a subtype match for `Q` with respect to `L` under constraints `C`.
+- A type `P<M0, ..., Mk>` is a subtype match for `P<N0, ..., Nk>` with respect
```
stereotype441: Are we missing a case here?  Something like:

A type `P` is a subtype match for a type varible `T` with respect to `L` under constraints `C`: ...

Or maybe this case is *never* a subtype match?

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+constrained type variable is required to be either a subtype (`T <: M`) or a
+supertype (`M <: T`) of a specific type (or type schema) `M`.  Constraints in
+which the same variable occurs on the left and the right of the constraint are
+ill-formed.
+
+### Closure of subtype and supertype constraints.
+
+The closure of a subtype constraint `T <: M` with respect to a set of type
+variables `L` is the subtype constraint `T <: N` where `N` is the least closure
+of `M` with respect to `L`, and similarly for supertype constraints.
```
kmillikin: We have not explicitly defined supertype constraints.  I guess this means constraints of the form `M <: T`.

'and similarly for supertype constraints' seems redundant with the next paragraph which should not itself say 'and similarly for supertype constraints'.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+computed due to f-bounded subtyping.  If we do not take bounds into
+consideration (and even if we do) we may create types `A<Object>` where `Object`
+does not satisfy the bound for `A`.  If we accept super-bounded types of this
+form (as we have proposed to do) then this is not problematic.  If we don't
+accept these, then we need to be sure that there is eventually an appropriate
+error check on the final inferred types.
+
+## Upper bound
+
+We write `UP(T0, T1)` for the upper bound of `T0` and `T1`and `DOWN(T0, T1)` for
```
kmillikin: Do `T0` and `T1` here range over types?  Type variables?

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+variables `L` is the subtype constraint `N <: T` where `N` is the greatest
+closure of `M` with respect to `L`, and similarly for supertype constraints.
+
+Note that the closure of a type constraint implies the original constraint: that
+is, any solution to the original constraint is a solution to the new constraint.
+
+The motivation for these operations is that constraint generation may produce a
+constraint on a type variable from an outer scope (say `S`) that refers to a
+type variable from an inner scope (say `T`).  For example, ` <T>(S) -> int <:
+<T>(List<T> -> int ` constrains `List<T>` to be a subtype of `S`.  But this
```
kmillikin: Unbalanced parentheses.  I'm a Lisper, I always spot 'em.

`<T>(S) -> int <: <T>(List<T>) -> int` seems to ill-formed according to "constraints in which the same variable occurs on the left and the right of the constraint are ill-formed".

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+#### Grounded constraint solution for a type variable
+
+The grounded constraint solution for a type variable `T` with respect to a
+constraint set `C` is define as follows:
+  - let `Mb <: T <: Mb` be the merge of `C` with respect to `T`.
+  - If `Mb` is known (that is, it does not contain `?`) then the solution is
+    `Mb`
+  - Otherwise, if `Mt` is known (that is, it does not contain `?`) then the
+    solution is `Mt`
+  - Otherwise, if `Mb` is not `?` then the solution is the greatest closure of
```
stereotype441: Should this be changed to "least closure" to mirror the change below?

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+    `Mb` with respect to `?`
+  - Otherwise the solution is the least closure of `Mt` with respect to `?`.
```
stereotype441: As discussed over IM, we want to change this to "greatest closure"

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,500 @@
...
+The closure of a subtype constraint `T <: M` with respect to a set of type
+variables `L` is the subtype constraint `T <: N` where `N` is the least closure
+of `M` with respect to `L`, and similarly for supertype constraints.
+
+The closure of a supertype constraint `M <: T` with respect to a set of type
+variables `L` is the subtype constraint `N <: T` where `N` is the greatest
+closure of `M` with respect to `L`, and similarly for supertype constraints.
+
+Note that the closure of a type constraint implies the original constraint: that
+is, any solution to the original constraint is a solution to the new constraint.
```
stereotype441: I'm confused: is the closure of a type constraint intended to be more strict or less strict than the original constraint?  I think you intend it to be more strict, i.e. closure of (T <: U) with respect to U should be (T <: Null).  This is consistent with saying "the closure of a type constraint implies the original constraint" (i.e. (T <: Null) implies (T <: U)).

But it's not consistent with saying "any solution to the original constraint is a solution to the new constraint".  For the example above, "T=U" is a solution to the original constraint, but not to the new constraint.

Did you perhaps mean to say: "...that is, any solution to the new constraint is a solution to the original constraint."?

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+  - If `e` is inferred as `m` of type `M` in context `_`
+- The expression `<T>(P x) => e` is inferred as `<T>(P x) => m` of type `<T>(P)
+  -> M` in context `<S>(Q) -> N`
+  - If `e` is inferred as `m` of type `M` in context `N[T/S]`
+  - Note: `x` is resolved as having type `P`  for inference in `e`
+- The expression `<T>(x) => e` is inferred as `<T>(dynamic x) => m` of type
+  `<T>(dynamic) -> M` in context `_`
+  - If `e` is inferred as `m` of type `M` in context `_`
+- The expression `<T>(x) => e` is inferred as `<T>(Q[T/S] x) => m` of type
+  `<T>(Q[T/S]) -> M` in context `<S>(Q) -> N`
```
leafpetersen: All uses of `Q[T/S]` here should be replaced with `R` where `R` is the greatest closure of `Q[T/S]` with respect to `?`.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+The incorporation of the type bounds information is asymmetric and procedural:
+it iterates through the bounds in order (`Bi[R0/T0, ..., ?/Ti, ..., ?/Tn]`).  Is
+there a better formulation of this that is symmetric but still allows some
+propagation?
+
+### Inference rules
+
+- The expression `e as T` is inferred as `m as T` of type `T` in context `K`:
+  - If `e` is inferred as `m` in an empty context
+- The expression `x = e` is inferred as `x = m` of type `T` in context `K`:
```
stereotype441: I confess that in my implementation work so far I have been largely ignoring this section and following my intuition because I had a difficult time following the style in which these rules are written.  This morning I figured out the reason I've been confused: because the language used here makes the data flow and the recursion implicit, and mixes inputs and outputs.  For example, on this particular line, it was not clear to me for a long time that `x = e` and `K` are inputs to the inference algorithm, whereas `x = m` and `T` are outputs.  Nor the fact that `T` and `m` are defined to be the outputs of the recursive call to type inference in the line below.

Also I found it confusing that the exposition treats `e` and `m` as two distinct expressions, whereas in my intuitive understanding of the algorithm (as well as in both of our implementations of it) they are represented by the same data structure at different times (`e` is the expression before it is inferred and `m` is the expression afterwards).

Consider rephrasing this section in a way that makes the recursive nature of the algorithm more explicit, makes the data flow more procedural, and uses `e'` rather than `m` to emphasize that it is conceptually the same expression tree, but with type inference information added.  For example, maybe these two lines could be rewritten something like this:

- To infer `x = e` in context `K`:
  - Let `M` be the type of `x`.
  - Apply type inference to `e` in context `M`, producing `e'` with type `T`.
  - Then the result of inference is `x = e'` with type `T`.

(Note: I'm considering trying to rework this section on my own, just to make sure I understand it.  If I do, I'll share the resulting text with you and you can decide whether you want to use any of it.)

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+    same type expectation context as described above.
+  - TODO(leafp): flesh this out.
+- For async and generator functions, the downwards context type is computed as
+  above, except that the propagated downwards type is taken from the type
+  argument to the `Future` or `Iterable` or `Stream` constructor as appropriate.
+  If the return type is not the appropriate constructor type for the function,
+  then the downwards context is empty.  Note that subtypes of these types are
+  not considered (this is a strong mode error).
+- The expression `e(e0, .., ek)` is inferred as `m<M0, ..., Mn>(m0, ..., mk)` of
+  type `N` in context `_`:
```
stereotype441: Are we missing a case here?  What if `e(e0, .., ek)` is inferred in a context other than `_`?

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+- The expression `e as T` is inferred as `m as T` of type `T` in context `K`:
+  - If `e` is inferred as `m` in an empty context
+- The expression `x = e` is inferred as `x = m` of type `T` in context `K`:
+  - If `e` is inferred as `m` of type `T` in context `M` where `x` has type `M`.
+- The expression `x ??= e` is inferred as `x ??= m` of type `UP(T, M)` in
+  context `K`:
+  - If `e` is inferred as `m` of type `T` in context `M` where `x` has type `M`.
+- The expression `await e` is inferred as `await m` of type `T` in context `K`:
+  - If `e` is inferred as `m` of type `T` in context `J` where:
+    - `J` is `FutureOr<K>` if `K` is not `_`, and is `_` otherwise
```
stereotype441: FYI, analyzer's rule is:
- If `K` is `_`, `dynamic`, or `FutureOr<L>` for some `L`, `J` is `K`.
- Otherwise `J` is `FutureOr<K>`

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+- The expression `e0..e1` is inferred as `m0..m1` of type `T` in context `K`
+  - If `e0` is inferred as `m0` of type `T` in context `K`
+  - And `e1` is inferred as `m1` of type `P` in context `_`
+- The expression `e0 ? e1 : e2` is inferred as `m0 ? m1 : m2` of type `T` in
+  context `K`
+  - If `e0` is inferred as `m0` of any type in context `bool`
+  - And `e1` is inferred as `m1` of type `T0` in context `K`
+  - And `e2` is inferred as `m2` of type `T1` in context `K`
+  - Where `T` is the greatest closure of `K` with respect to `?` if `K` is not
+    `_` and otherwise `UP(T0, T1)`
```
stereotype441: The current implementation of type inference in analyzer doesn't follow this rule.  It always uses `UP(T0, T1)` for `T`.

I think analyzer's behavior is more intuitive.  Consider, for example:

    class C<T> {
      C(List<T> x);
    }

    main() {
      bool b = false;
      List<int> l1 = [1];
      List<int> l2 = [2];
      var x = new C(l1);
      var y = new C(l2);
      var z = new C(b ? l1 : l2);
    }

With analyzer's rule, we infer a type of `C<int>` for `x`, `y`, and `z`, which matches my intuitive expectation.  With the rule you've specified here, `z` gets a type of `C<dynamic>`, which IMHO is unexpected.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+### Inference rules
+
+- The expression `e as T` is inferred as `m as T` of type `T` in context `K`:
+  - If `e` is inferred as `m` in an empty context
+- The expression `x = e` is inferred as `x = m` of type `T` in context `K`:
+  - If `e` is inferred as `m` of type `T` in context `M` where `x` has type `M`.
+- The expression `x ??= e` is inferred as `x ??= m` of type `UP(T, M)` in
+  context `K`:
+  - If `e` is inferred as `m` of type `T` in context `M` where `x` has type `M`.
+- The expression `await e` is inferred as `await m` of type `T` in context `K`:
```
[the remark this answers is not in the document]
kmillikin: I agree with @stereotype441 here.

docs/language/informal/local-inference.md
```diff
@@ -0,0 +1,498 @@
...
+Constructor declaration (field initializers)
+Default parameters
+
+## Statement inference.
+
+TODO(leafp)
+
+Return statements pull the return type from the enclosing function for downwards
+inference, and compute the upper bound of all returned values for upwards
+inference.  Appropriate adjustments for asynchronous and generator functions.
```
kmillikin: Something like when returning a value from an `async` function, infer the type of the subexpression, use _flatten_ on that type, and wrap that in a `Future`.

mit-mit:
Closing as stale

PR #29400, opened 2017-04-20
Scan all packages files found in the out/ directory to build mappings

pylaligand:
This accounts for the (common) situation where multiple targets (library, binaries, test) are declared inside a single Dart package.

### DIFF
--- a/pkg/analyzer/lib/src/generated/gn.dart
+++ b/pkg/analyzer/lib/src/generated/gn.dart
@@ -39,9 +39,9 @@ class GnWorkspace extends Workspace {
   final String root;
 
   /**
-   * The path to the .packages file.
+   * The paths to the .packages files.
    */
-  String _packagesFilePath;
+  final List<String> _packagesFilePaths;
 
   /**
    * The map of package locations indexed by package name.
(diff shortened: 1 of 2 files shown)


bwilkerson:
I'm not sure I have the context I need in order to understand this change, but it looks like we're allowing multiple `.packages` files and that we then analyze code in the workspace against the result of merging all of those files. Is that the essence of this change?

If so, the one concern I have is that there's no handling of conflicts between the `.packages` files. What should happen if two or more `.packages` files both contain a mapping for the same package name? At the moment, it appears to be non-deterministic, depending on the order in which files are found.

pylaligand:
Yes, multiple .packages files are generated for a single package. This is because in our build files we're separating libraries, binaries, and tests, and that they all need to generate their own .packages file as part of the build. The analyzer service does not have that much granularity and therefore needs to be provided with all the dependencies at once.

There's definitely a risk for conflicts but in practice it is very low as most binaries use the same exact dependencies as the library in the package: the only real addition comes from test targets which pull test-related dependencies. It roughly amounts to merging "dependencies" and "dev_dependencies".

bwilkerson (approved):
Sounds reasonable to me.

PR #29459, opened 2017-04-25
New Lint quick-fix:  unnecessary_lambdas

JPaulsen:
From [Linter](http://dart-lang.github.io/linter/lints/unnecessary_lambdas.html)

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/fix.dart
+++ b/pkg/analysis_server/lib/src/services/correction/fix.dart
@@ -111,7 +111,8 @@ bool hasFix(ErrorCode errorCode) =>
         (errorCode.name == LintNames.annotate_overrides ||
             errorCode.name == LintNames.avoid_init_to_null ||
             errorCode.name == LintNames.prefer_collection_literals ||
-            errorCode.name == LintNames.unnecessary_brace_in_string_interp));
+            errorCode.name == LintNames.unnecessary_brace_in_string_interp ||
+            errorCode.name == LintNames.unnecessary_lambdas));
 
 /**
  * An enumeration of possible quick fix kinds.
@@ -229,6 +230,8 @@ class DartFixKind {
       'REPLACE_WITH_NULL_AWARE',
       50,
       "Replace the '.' with a '?.' in the invocation");
+  static const REPLACE_WITH_TEAR_OFF = const FixKind(
+      'REPLACE_WITH_TEAR_OFF', 50, "Replace function literal with tear-off");
   static const USE_CONST = const FixKind('USE_CONST', 50, "Change to constant");
   static const USE_EFFECTIVE_INTEGER_DIVISION = const FixKind(
       'USE_EFFECTIVE_INTEGER_DIVISION',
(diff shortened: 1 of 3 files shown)


pkg/analysis_server/test/services/correction/fix_test.dart
```diff
@@ -6106,6 +6106,66 @@ final a = {};
 ''');
   }
 
+  test_tearOffLambda_function_test1() async {
```
bwilkerson: Two nits.

Normally a method is a function associated with a class of objects. It looks like you're making that distinction in the naming of the test methods. If so, and given that `toString` is a method, I would have expected this to be named with `_method_` rather than `_function_`.

Also, we try to encode the case being tested in the name, and things like `test1` don't really fit that pattern. What are the different kinds of things you're trying to test for? I'd guess that includes (a) whether there's a target (the function verses method distinction), and (b) the number of parameters / arguments. If that's true, then perhaps rather than `_test1` you could use something like `_zeroParameters`.

Are there other factors that are important to the tests? For example, does it matter if the target is a global name, such as a class, and we're tearing off a static method? (I suspect not, but it's worth asking.) Is it worth having more than one parameter in at least one of the tests? Should there be a test in which the closure is an argument to another method (which seems like the most common case we're likely to find)?

JPaulsen: Fixed but having problems to create this test:

```dart
  test_tearOffLambda_lambda_asArgument() async {
    String src = '''
bool isPair(int a) => a % 2 == 0;
final finalList = <int>[];
finalList.where(/*LINT*/(number) =>
  isPair(number);
);
''';
    await findLint(src, LintNames.unnecessary_lambdas);

    await applyFix(DartFixKind.TEAR_OFF_LAMBDA);

    verifyResult('''
bool isPair(int a) => a % 2 == 0;
final finalList = <int>[];
finalList.where(isPair);
''');
  }
```

In this test the line

```dart
FunctionExpression ancestor =
  node.getAncestor((a) => a is FunctionExpression);
```

Is being resolved to null, do you know how to solve this?

JPaulsen: @bwilkerson I already compiled my local SDK in my machine and this cases are being well fixed, it looks like the problem is in the test

JPaulsen: Fixed, I surrounded in a function, PTAL

bwilkerson: The tests look good, thanks!

pkg/analysis_server/lib/src/services/correction/fix.dart
```diff
@@ -229,6 +230,8 @@ class DartFixKind {
       'REPLACE_WITH_NULL_AWARE',
       50,
       "Replace the '.' with a '?.' in the invocation");
+  static const TEAR_OFF_LAMBDA =
+      const FixKind('TEAR_OFF_LAMBDA', 50, "Tear-off lambda");
```
JPaulsen: Fixed

bwilkerson (approved):

PR #29469, opened 2017-04-25
New Lint quick-fix:  unnecessary_this

JPaulsen:
From [Linter](http://dart-lang.github.io/linter/lints/unnecessary_this.html)

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
+++ b/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
@@ -22,7 +22,6 @@ import 'package:analysis_server/src/services/correction/namespace.dart';
 import 'package:analysis_server/src/services/correction/source_buffer.dart';
 import 'package:analysis_server/src/services/correction/source_range.dart'
     as rf;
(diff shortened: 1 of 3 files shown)


pkg/analysis_server/lib/src/services/correction/fix_internal.dart
```diff
@@ -20,9 +20,9 @@ import 'package:analysis_server/src/services/correction/levenshtein.dart';
 import 'package:analysis_server/src/services/correction/name_suggestion.dart';
 import 'package:analysis_server/src/services/correction/namespace.dart';
 import 'package:analysis_server/src/services/correction/source_buffer.dart';
+import 'package:analysis_server/src/services/correction/source_range.dart';
```
bwilkerson: This library is already imported (on the line below) with a prefix. Seems like we should have one or the other, but not both.

JPaulsen: Yes, when I apply the sort members in code always move that

bwilkerson (approved):

PR #29472, opened 2017-04-26
New Lint quick-fix:  unnecessary_override

JPaulsen:
From [Linter](https://github.com/dart-lang/linter/pull/614)

### DIFF
--- a/pkg/analysis_server/test/services/correction/fix_test.dart
+++ b/pkg/analysis_server/test/services/correction/fix_test.dart
@@ -5986,6 +5986,71 @@ var x;
 ''');
   }
 
+  test_removeMethodDeclaration_getter() async {
+    String src = '''
+class A {
+  int x;
+}
+class B extends A {
+  @override
+  int get /*LINT*/x => super.x;
+}
+''';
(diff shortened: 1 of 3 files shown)


pkg/analysis_server/test/services/correction/fix_test.dart
```diff
@@ -5986,6 +5986,74 @@ var x;
...
+class B extends A {
+
+}
+''');
+  }
+
+  test_removeMethodDeclaration_method() async {
+    String src = '''
+class A {
+@override
```
JPaulsen: @bwilkerson is there a way to safely remove the indentation and the new line when removing the method?

bwilkerson: I think that `utils.getLinesRange` will do what you're looking for.

JPaulsen: It worked!, Fixed

bwilkerson (approved):
It would be good if one of the existing tests were extended to have members before and after the one being removed.

PR #29473, opened 2017-04-26
New Lint quick-fix:  prefer_conditional_assignment

JPaulsen:
From [Linter](https://github.com/dart-lang/linter/pull/567)

### DIFF
--- a/pkg/analysis_server/test/services/correction/fix_test.dart
+++ b/pkg/analysis_server/test/services/correction/fix_test.dart
@@ -6074,6 +6074,109 @@ class A {
 ''');
   }
 
+  test_replaceWithConditionalAssignment_withCodeBeforeAndAfter() async {
+    String src = '''
+class Person {
+  String _fullName;
+  void foo() {
+    print('hi');
+    /*LINT*/if (_fullName == null) {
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):
It would be good if one of the tests had code before and after the if statement being replaced.

JPaulsen:
@bwilkerson test added

pkg/analysis_server/test/services/correction/fix_test.dart
```diff
@@ -6074,6 +6074,109 @@ class A {
 ''');
   }
 
+  test_replaceWithConditionalAssignment_withCodeBeforeAndAfter() async {
+    String src = '''
+class Person {
+  String _fullName;
+  void foo() {
+    print('hi);
```
bwilkerson: The string literal isn't closed correctly. It must be the case that the parser recovers at the end of the line and the extra errors don't cause issues for the test framework. Still, it would be good to fix it (here and below).

bwilkerson (approved):

pkg/analysis_server/test/services/correction/fix_test.dart

PR #29489, opened 2017-04-27
New Lint quick-fixes: REMOVE_TYPE - 3 different lints

JPaulsen:
This quick-fix just remove a type, it is useful to 3 lints by now, they are:

From [Linter] (http://dart-lang.github.io/linter/lints/avoid_return_types_on_setters.html)
From [Linter] (https://github.com/dart-lang/linter/pull/545)
From [Linter] (https://github.com/dart-lang/linter/pull/551)

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/fix.dart
+++ b/pkg/analysis_server/lib/src/services/correction/fix.dart
@@ -216,6 +216,8 @@ class DartFixKind {
       "Remove parentheses in getter invocation");
   static const REMOVE_THIS_EXPRESSION =
       const FixKind('REMOVE_THIS_EXPRESSION', 50, "Remove this expression");
+  static const REMOVE_TYPE_NAME =
+      const FixKind('REMOVE_TYPE_NAME', 50, "Remove type name");
   static const REMOVE_UNNECESSARY_CAST =
       const FixKind('REMOVE_UNNECESSARY_CAST', 50, "Remove unnecessary cast");
   static const REMOVE_UNUSED_CATCH_CLAUSE =
@@ -236,6 +238,8 @@ class DartFixKind {
       const FixKind('REPLACE_WITH_BRACKETS', 50, "Replace with { }");
   static const REPLACE_WITH_CONDITIONAL_ASSIGNMENT = const FixKind(
       'REPLACE_WITH_CONDITIONAL_ASSIGNMENT', 50, 'Replace with ??=');
+  static const REPLACE_WITH_IDENTIFIER =
+      const FixKind('REPLACE_WITH_IDENTIFIER', 50, "Replace with identifier");
   static const REPLACE_WITH_LITERAL =
       const FixKind('REPLACE_WITH_LITERAL', 50, 'Replace with literal');
   static const REPLACE_WITH_NULL_AWARE = const FixKind(
--- a/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
+++ b/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
@@ -381,8 +381,17 @@ class FixProcessor {
       if (errorCode.name == LintNames.annotate_overrides) {
         _addLintFixAddOverrideAnnotation();
       }
+      if (errorCode.name == LintNames.avoid_annotating_with_dynamic) {
+        _addFix_removeTypeName();
+      }
       if (errorCode.name == LintNames.avoid_init_to_null) {
         _addFix_removeInitializer();
+      }
+      if (errorCode.name == LintNames.avoid_return_types_on_setters) {
+        _addFix_removeTypeName();
+      }
+      if (errorCode.name == LintNames.avoid_types_on_closure_parameters) {
+        _addFix_replaceWithIdentifier();
       }
       if (errorCode.name == LintNames.empty_statements) {
         _addFix_removeEmptyStatement();
@@ -1892,6 +1901,14 @@ class FixProcessor {
     }
   }
 
+  void _addFix_removeTypeName() {
+    final TypeName type = node.getAncestor((node) => node is TypeName);
+    if (type != null) {
+      _addRemoveEdit(rf.rangeStartEnd(type.offset, type.endToken.next.offset));
+      _addFix(DartFixKind.REMOVE_TYPE_NAME, []);
+    }
+  }
+
   void _addFix_removeUnnecessaryCast() {
     if (coveredNode is! AsExpression) {
       return;
@@ -1981,6 +1998,18 @@ class FixProcessor {
       var instanceCreation = coveredNode as InstanceCreationExpression;
       _addReplaceEdit(rf.rangeToken(instanceCreation.keyword), 'const');
       _addFix(DartFixKind.USE_CONST, []);
+    }
+  }
+
+  void _addFix_replaceWithIdentifier() {
+    final FunctionTypedFormalParameter functionTyped =
+        node.getAncestor((node) => node is FunctionTypedFormalParameter);
+    if (functionTyped != null) {
+      _addReplaceEdit(rf.rangeNode(functionTyped),
+          utils.getNodeText(functionTyped.identifier));
+      _addFix(DartFixKind.REPLACE_WITH_IDENTIFIER, []);
+    } else {
+      _addFix_removeTypeName();
     }
   }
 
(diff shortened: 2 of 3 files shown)


pkg/analysis_server/lib/src/services/correction/fix.dart
```diff
@@ -214,6 +214,7 @@ class DartFixKind {
       "Remove parentheses in getter invocation");
   static const REMOVE_THIS_EXPRESSION =
       const FixKind('REMOVE_THIS_EXPRESSION', 50, "Remove this expression");
+  static const REMOVE_TYPE = const FixKind('REMOVE_TYPE', 50, "Remove type");
```
bwilkerson: 'REMOVE_TYPE' --> 'REMOVE_TYPE_NAME'

JPaulsen:
@bwilkerson now this is ready to review, I made some fixes, PTAL

pkg/analysis_server/test/services/correction/fix_test.dart
```diff
@@ -6139,6 +6139,134 @@ class A {
...
+    await findLint(src, LintNames.avoid_annotating_with_dynamic);
+
+    await applyFix(DartFixKind.REMOVE_TYPE);
+
+    verifyResult('''
+bad([defaultValue]) {
+  return null;
+}
+''');
+  }
```
bwilkerson: All of the tests for `avoidAnnotatingWithDynamic` are for parameters. Does this lint get generated for other kinds of variables (local variables, top-level, etc.)?

pkg/analysis_server/lib/src/services/correction/fix_internal.dart
```diff
@@ -1875,6 +1884,25 @@ class FixProcessor {
...
+    final FunctionTypedFormalParameter functionTyped =
+        node is FunctionTypedFormalParameter
+            ? node
+            : node.getAncestor((node) => node is FunctionTypedFormalParameter);
+    if (functionTyped != null) {
+      _addReplaceEdit(rf.rangeNode(functionTyped),
+          utils.getNodeText(functionTyped.identifier));
+      _addFix(DartFixKind.REMOVE_TYPE, []);
+      return;
+    }
```
bwilkerson: Can we get confused if a function-typed formal parameter is *not* in a closure but includes something with a type of `dynamic`? Maybe something like:
```
void callFunction(void f(dynamic p)) {}
```

bwilkerson (commented):
I don't see any changes. Did you upload the changes? (The changes to add avoid_return_types_on_setters were uploaded while I was reviewing, in case that's what you were referring to.)

pkg/analysis_server/test/services/correction/fix_test.dart
JPaulsen: Nope, we decided that for now it was only with parameters, we would like to extend (the lint) to return type also but we decided that not yet because omit them could not maintain the same behavior than typing dynamic

pkg/analysis_server/lib/src/services/correction/fix_internal.dart
JPaulsen: I will double check, it is an interesting test to do, thanks for noticing it

pkg/analysis_server/lib/src/services/correction/fix.dart
JPaulsen: @bwilkerson what do you think about the message given to the user, does 'Remove type' still ok?

JPaulsen: Fixed, I renamed both

JPaulsen:
@bwilkerson yes, that was I referred to

pkg/analysis_server/lib/src/services/correction/fix.dart
bwilkerson: Yes, I think the message is fine.

pkg/analysis_server/lib/src/services/correction/fix_internal.dart
JPaulsen: Yes you were right, the behavior always will prefer to remove the function-typed formal parameter and this is correct for our rules, for example if you have this:

```dart
void callFunction(/*LINT*/void f(dynamic p)) {}
```

The quick-fix (without considering the meaning of the lints) is ambiguous, it could refer the typeName of the returnType (void in this case) or the function-typed formal parameter, and the fix always prefer the function-typed formal parameter and that is because is the behavior needed for fix the lint rules now, so the solution is correct but...

The problem could appear if someone would like to use this fix in the future, so it is relevant to know what you noticed to take a decision, so I added documentation to the quick-fix explaining this behavior.

pkg/analysis_server/lib/src/services/correction/fix_internal.dart
```diff
@@ -1875,6 +1884,29 @@ class FixProcessor {
     }
   }
 
+  /// This fix is not going to work well if what you want is to remove only a
+  /// TypeName inside a FunctionTypedFormalParameter (returnType, parameter or
+  /// generic) so if you want to do that you should create other fix or modify
+  /// this one carefully.
```
bwilkerson: Does this comment (and the lack of a test method for it) mean that the test I suggested fails? If so, I consider that to be a blocker. One possible way to get around that is to pass a `bool` into this method that controls whether it looks for `FunctionTypedFormalParameter`s.

bwilkerson: I didn't see your other comment. I'll respond to it, and you can ignore this comment.

pkg/analysis_server/lib/src/services/correction/fix_internal.dart
bwilkerson: I don't think it works right from a users perspective. If the user sees a diagnostic telling them that they should avoid using `dynamic`, and the fix says it will remove the type, then it's reasonable for them to assume that the type that will be returned will be `dynamic`, not the rest of the function type.

See my other comment for a possible way of implementing the fix the way I think a user would expect it to work.

JPaulsen: Yes, but that is not going to happen, because the avoid using dynamic only can be linted in a parameter of a function expression, it can not happen in the parameter of a function-typed formal parameter, so it is the right solution, the only case that this would fail if it would be necessary a fix to remove a typeName inside a function-typed formal parameter, and for now there is not lint that requires that

bwilkerson: Looking at the code for `AvoidAnnotatingWithDynamic` it looks like it will visit every `SimpleFormalParameter`, even ones inside a function typed formal parameter.

(As a side-note, adding a passing test that proves that we get the right behavior would probably shorten this conversation :-)

JPaulsen: Yes, you are right, sorry for making you waste your time, I mixed the rules in my mind, so yes, AvoidAnnotatingWithDynamic can crash with this fix, avoid_types_on_closure_parameters is the one that do that what I said.

I did not create the tests because I did not know how to do it without a lint, but know how you helped me to figure out what happened can create it will AvoidAnnotatingWithDynamic and fix the fix, thanks!

pkg/analysis_server/test/services/correction/fix_test.dart
JPaulsen: @bwilkerson fixed and test added, I also added a different message when removing the types of a TypedFunctionFormalParameter PTAL

pkg/analysis_server/test/services/correction/fix_test.dart
```diff
@@ -6139,6 +6139,151 @@ class A {
...
+    await findLint(src, LintNames.avoid_return_types_on_setters);
+
+    await applyFix(DartFixKind.REMOVE_TYPE_NAME);
+
+    verifyResult('''
+set speed2(int ms) {}
+''');
+  }
+
+  test_removeTypeName_avoidTypesOnClosureParameters_FunctionTypedFormalParameter() async {
```
bwilkerson: Just one nit: From here down, the new tests are testing the `replaceWithIdentifier` fix rather than the `removeTypeName` fix.

bwilkerson (approved):
Thanks!

pkg/analysis_server/test/services/correction/fix_test.dart

alexeieleusis:
Apparently all travis builds are failing, should we wait or merge? @bwilkerson

bwilkerson:
As far as I can see, one test failed with a timeout:
```
[1A[2K[31mChrome 60.0.3080 (Linux 0.0.0)  lib/html/queryall_test FAILED[39m
	Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
	    at finish (test/browser/language_tests.js:829:11)
	    at Context.<anonymous> (test/browser/language_tests.js:862:13)
Chrome 60.0.3080 (Linux 0.0.0): Executed 2237 of 2537[31m (1 FAILED)[39m (0 secs / 1 min 5.22 secs)
```
I'd ignore it because I'm fairly sure this change didn't cause it.

PR #29490, opened 2017-04-27
New Lint quick-fix: await_only_futures

JPaulsen:
From [Linter](http://dart-lang.github.io/linter/lints/await_only_futures.html)

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
+++ b/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
@@ -393,6 +393,9 @@ class FixProcessor {
       if (errorCode.name == LintNames.avoid_types_on_closure_parameters) {
         _addFix_replaceWithIdentifier();
       }
+      if (errorCode.name == LintNames.await_only_futures) {
+        _addFix_removeAwait();
+      }
       if (errorCode.name == LintNames.empty_statements) {
         _addFix_removeEmptyStatement();
       }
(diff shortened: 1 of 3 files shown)


pkg/analysis_server/lib/src/services/correction/fix_internal.dart
```diff
@@ -1784,6 +1787,18 @@ class FixProcessor {
     _addFix(DartFixKind.ADD_NE_NULL, []);
   }
 
+  void _addFix_removeAwait() {
+    AwaitExpression awaitExpression = node is AwaitExpression
+        ? node
+        : node.getAncestor((node) => node is AwaitExpression);
```
bwilkerson: Will the `node` every *not* be an await expression? Just wondering whether this defensive code could be removed. (Here and in other fix methods.)

pkg/analysis_server/lib/src/services/correction/fix_internal.dart
```diff
@@ -1784,6 +1787,16 @@ class FixProcessor {
...
   }
 
+  void _addFix_removeAwait() {
+    AwaitExpression awaitExpression = node;
+    if (awaitExpression != null) {
+      final awaitToken = awaitExpression.awaitKeyword;
+      _addRemoveEdit(
+          rf.rangeStartEnd(awaitToken.offset, awaitToken.next.offset));
+      _addFix(DartFixKind.REMOVE_AWAIT, []);
+    }
```
bwilkerson: This works, but has the disadvantage that if the node is ever not an await expression we'll get a cast exception at runtime. That's why I was recommending
```
if (node is AwaitExpression) {
  final awaitToken = node.awaitKeyword;
  ...
}
```
Worst case with the code above is that we silently don't produce a fix.

bwilkerson (approved):

alexeieleusis:
Same thing here, I am merging @bwilkerson 
We can revert if needed. Will wait a couple of minutes.

PR #29491, opened 2017-04-27
New Lint quick-fix: empty_statements

JPaulsen:
From [Linter](http://dart-lang.github.io/linter/lints/empty_statements.html)

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
+++ b/pkg/analysis_server/lib/src/services/correction/fix_internal.dart
@@ -384,6 +384,9 @@ class FixProcessor {
       if (errorCode.name == LintNames.avoid_init_to_null) {
         _addFix_removeInitializer();
       }
+      if (errorCode.name == LintNames.empty_statements) {
+        _addFix_removeEmptyStatement();
+      }
       if (errorCode.name == LintNames.prefer_collection_literals) {
         _addFix_replaceWithLiteral();
       }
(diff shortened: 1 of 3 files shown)


pkg/analysis_server/lib/src/services/correction/fix_internal.dart
```diff
@@ -1816,6 +1819,16 @@ class FixProcessor {
     }
   }
 
+  void _addFix_removeEmptyStatement() {
+    EmptyStatement emptyStatement = node is EmptyStatement
+        ? node
+        : node.getAncestor((node) => node is EmptyStatement);
```
bwilkerson: Empty statements can't have anything inside them, so I *know* this is unnecessary. Just replace the null check below with `node is EmptyStatement`.

pkg/analysis_server/test/services/correction/fix_test.dart
```diff
@@ -5943,6 +5943,44 @@ main() {
...
+
+    verifyResult('''
+void foo() {
+  while(true) {
+  }
+}
+''');
+  }
+
+  test_removeEmptyStatement_outOfBlock() async {
```
bwilkerson: Just deleting the empty statement here changes the semantics of the code, which is something we try to avoid. Might be better to replace `;` with `{}` when the parent is not a block. What do you think?

JPaulsen: You are totally right, I noticed the same and the lint marks thoses cases, I agree with replacing it with `{}` in those cases, I think is the best solution.

pkg/analysis_server/lib/src/services/correction/fix_internal.dart

bwilkerson (approved):

PR #29571, opened 2017-05-06
Disable analysis on a few packages.

pylaligand:


zanderso:
Is this to speed up the build, or because the packages aren't under Fuchsia control, or both? Would it make sense to do this for all the packages like this that Fuchsia uses?

zanderso (approved):

pylaligand:
I'm about to turn on Dart analysis in the Fuchsia build and errors will make the build fail. We have a set of default analysis options which don't match the practices in these packages. One option is to disable analysis in Fuchsia for these packages, which is what this change is about; another option is to define the options you want to use for these packages and then the Fuchsia build will use these options.

PR #29589, opened 2017-05-09
double::toString perf improvement for integer case

### LINKED ISSUE #17622: performance: double.toString
_This issue was originally filed by @tatumizer_

---

I will start submitting performance issues gradually, following reply from Srdjan and Peter.

calling 123456.toString() takes 28 nanoseconds (very good! improved by a factor of 3 in the last release!)
calling 123456.0.toString() takes 330 nanoseconds. In V8 javascript, same conversion takes 80 nanoseconds.

joshualitt:
fixes #17622

This change is experimental, but I wanted to get some feedback before continuing. I haven't tested on mobile but on desktop its a pretty clear win in the integer case without a significant penalty in the non integer case.

Open question - if we want to go through with this, should we cache integer values?

### DIFF
--- a/runtime/lib/double.dart
+++ b/runtime/lib/double.dart
@@ -169,16 +169,34 @@ class _Double implements double {
   String _toString() native "Double_toString";
 
   String toString() {
+    if (identical(0.0, this)) {
+      return "0.0";
+    } else if (identical(-0.0, this)) {
+      return "-0.0";
+    } else if (isNaN) {
+      return "NaN";
+    } else if (this == double.INFINITY) {
+      return "Infinity";
+    } else if (this == -double.INFINITY) {
+      return "-Infinity";
+    }
+
+    // This will not throw because we have already tested
+    // for special values. Furthermore, we only take this
+    // path if we are between a range to mimic inaccuracies
+    // in certain floating point calculations.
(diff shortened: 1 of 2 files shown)


joshualitt:
Whoops, there are some bugs in this for handling special double cases. I will fix and reupload.

runtime/vm/object.cc
```diff
@@ -19334,12 +19334,9 @@ RawString* Number::ToString(Heap::Space space) const {
 
 
 const char* Double::ToCString() const {
```
mraleph: This function is called directly from the C++ code bypassing Dart code so this change is not correct.

joshualitt: whoops, of course. Fixed

joshualitt:
Okay, I've updated the cl to localize the changes in double.dart. With some careful refactoring I can share more code with the integer case, but I would have to introduce a few more branches. Thoughts?

joshualitt:
Upon reflection I was able to share the majority of the shareable code just by some careful refactoring, and without needing any additional branches. I'm assuming the compiler can inline static functions, but maybe that is not a correct assumption?

mit-mit:
Closing as this appears to be stale

PR #29604, opened 2017-05-12
DEPS ans status files updated to the last co19 commit

sgrekhov:


whesse:
LGTM.  I will merge it, and then revert after getting some results on the buildbot.

PR #29616, opened 2017-05-15
VM: [Kernel] elide temporaries if the expression result is not used

DmitryOlshansky:
This in a way mimics what Effect/Value/Test hierarchy of normal
FlowGraphBuilder does but relies on guard objects to manipulate
a single flag.


DmitryOlshansky:
I signed it!

PR #29624, opened 2017-05-16
"Enhanced Type Promotion" informal proposal.

munificent:
This is the proposal and tracking issue to improve the language's type promotion rules to allow promotion to apply in more places.

@bwilkerson

### DIFF
--- a/docs/language/informal/enhanced-type-promotion.md
+++ b/docs/language/informal/enhanced-type-promotion.md
@@ -114,8 +114,8 @@ The language specifies type promotion out of four components:
     promotion.** The spec says that promotion only applies if the promoted type
     is a subtype of the declared type.
 
-This proposal keeps the same general pieces, but refines the first three of
-them. The forth is unchanged.
+This proposal keeps the same general pieces, but refines the first two of
+them. The third and forth are unchanged.
 
 ## Facts for False Expressions
 
@@ -259,22 +259,13 @@ some block:
 
     This is also allowed if the if statement has no else clause at all.
 
-## Promoting More Variables
+## Alternatives Considered
 
-The specification places several restrictions on the variables whose type can be promoted. The variable `v`:
+### Allowing assignment inside of the promoted scope
 
-*   Must be either a local variable or a parameter,
-
-*   cannot be potentially assigned to within the scope S that the promotion
-    would apply to,
-
-*   cannot be potentially assigned to within any closure, and
-
-*   if `v` is captured by a closure in the scope S then `v` cannot be
-    potentially mutated anywhere.
-
-The most common case where this eliminates useful promotion is the second rule.
-Consider:
+One frequent annoyance of the current type promotion rules is that assigning to
+the promoted variable anywhere inside the scope where the variable is promoted
+kills the promotion, even before the assignment. For example:
 
 ```dart
 Object trimIfString(Object o) {
@@ -283,13 +274,11 @@ Object trimIfString(Object o) {
 }
 ```
 
-In Dart today, the assignment to `o` prevents `o` from being promoted. That in
-turn means the call to `o.trim()` has a static error since Object has no
-`trim()` method.
+This code is dynamically correct and intuitively should work. But because `o`
+is assigned to inside the then body, no promotion occurs. Since Object doesn't
+define `trim()`, you get a static error.
 
-We accommodate these cases by simply removing the second restriction. We allow
-promotion to occur even if the variable is assigned within the promoted scope.
-This opens the door to code that is not safe:
+This restriction does help avoid some patently wrong code:
 
 ```dart
 test(Object o) {
@@ -300,50 +289,134 @@ test(Object o) {
 }
 ```
 
-Obviously, by the time we execute `o.length`, `o` will no longer be a String and
-the call will fail. We address this by promoting `o`'s static type not just for
-accesses of `o`, but for assignments to it as well.
+But it's too pessimistic since it also disallows valid code like the
+`trimIfString()` example. We considered two middle grounds. The first is to
+allow assignment as long as the type of the assigned value is a subtype of the
+promoted type. That means the first example would promote because the type of
+`o.trim()`, String, is a subtype of the promoted type, also String.
 
-In this example, if means the `o = 123` now generates a static error because an
-int cannot be assigned to a variable of (promoted) type String. There are still
-cases where the assignment is allowed but would lead to unsound code because of
-an implicit downcast:
+The problem is that when we are calculating promotion, we don't know the types
+of the right-hand sides. The two features can interact with type inference in
+complex ways. Consider:
 
 ```dart
-test(Object o) {
-  if (o is String) {
-    o = new Object(); // No longer a String.
-    print(o.length);  // Error!
+if (o is String) {
+  o = f(o);
+}
+```
+
+Is this promotion valid? It depends on the type of `f(o)`. If `f` is a generic
+function, we need to infer its type parameter, which requires knowing the type
+of `o`. But `o`'s type depends on whether or not it is promote, which is what
+we're trying to calculate. We're stuck in a loop.
+
+An even nastier example is:
+
+```dart
+T fn<T, T>(T t, T t) => t;
+
+class A           {  D v; }
+class B extends A { E2 v; }
+class C extends B {}
+
+class D            { A a; }
+class E1 extends D { C a; }
+class E2 extends D {}
+
+foo(A x) {
+  if (x is B) {
+    var y = x.v;
+    if (y is E1) {
+      x = fn(y.a, new B());
+    }
   }
 }
 ```
 
-Here, the assignment `o = new Object()` has no static error, but would lead to
-code that fails later since Object has no `length` getter. We address this like
-all downcasts are addressed in strong mode, by checking at runtime that the
-assignment succeeds. Now, though, the runtime check is against the *promoted*
-type. In the above example, at runtime, we check that `new Object()` can be
-assigned to String. That check fails at runtime, and we never reach the invalid
-call to `o.length`.
+(Determining whether `x` and/or `y` should promote is left as an exercise for
+the reader.)
 
-*Optional: The restriction on mutation within closures could be relaxed to only
-apply to closures defined before the boolean expression that allows for the
-promotion or within the scope that the promotion would apply to. A similar
-relaxation could be made for the locations of closures that capture the
-variable. This would require defining precisely what we mean by "before", which
-might be more work than it's worth.*
-
-*Optional: The restriction on mutation within the scope that the promotion would
-apply to could be relaxed by allowing the promotion to apply to everything up to
-the point of mutation. For example:*
+A simpler alternative approach suggested by Leaf is to simply *always* allow
+assignment inside the promoted scope and then type check it against the promoted
+type. If you assign a value whose type is not a subtype of the promoted type,
+you get a type error but it doesn't nullify the promotion:
 
 ```dart
-if (v is T) {
-  // v has type T here
-  v = someNonTValue;
-  // v has the non-promoted type here
+test(Object o) {
+  if (o is String) {
+    o = 123;  // Error: 123 is not a String.
+    o.trim(); // OK.
+  }
 }
 ```
+
+Unfortunately, there are times when you still want to assign using the original
+declared typed. Consider:
+
+```dart
+class Node {}
+class Tree extends Node {
+  final Node left, right;
+}
+
+Node leftmostLeaf(Node node) {
+  while (node is Tree) {
+    node = node.left;
+  }
+
+  return node;
+}
+```
+
+Promoting `node` to Tree inside the while body lets the `node.left` expression
+work, which is good. But it breaks the assignment. The assignment becomes an
+implicit downcast from Node (the type of `node.left`) to Tree (the promoted type
+of `node`). That downcast is checked at runtime in strong mode and may
+fail.
+
+But the whole point of this code is to allow assigning a non-Tree to `node`.
+That's how you exit the loop. This program would throw a cast exception from otherwise correct code.
+
+Worse, there's no easy workaround. You can't "unpromote" a variable or "upcast"
+an lvalue like:
+
+```dart
+Node leftmostLeaf(Node node) {
+  while (node is Tree) {
+    node as Node = node.left;
+  }
+
+  return node;
+}
+```
+
+The most promising approach is to *change* the type of the promoted variable at
+the assignment. So, in this case, `node` would have the promoted type when
+evaluating `node.left`. And then immediately after the assignment, its type
+switches to Node, the resulting type of the assignment.
+
+But this requires control flow analysis. Consider examples like:
+
+```dart
+test(Object o) {
+  if (o is String) {
+    o = "String";
+  } else {
+    o = 1; // Change type to int.
+    o = o + 1; // OK. int + int.
+    o = "Not String"; // Now change to String.
+    o = o.trim(); // OK. String.trim().
+  }
+
+  return o.toLowerCase(); // OK?
+}
+```
+
+Here, since both paths definitely assign a String to `o`, ideally it would allow
+the `toLowerCase()` call on the result. That requires doing a join on the two
+control paths. This kind of analysis isn't *bad*, but it would make this
+proposal much more complex. We may go there, but we want something simpler we
+can implement soon to remove the ad hoc type propagation code in analyzer.
 
 ## Example Specification Wording
 
--- a/docs/language/informal/enhanced-type-promotion.md
+++ b/docs/language/informal/enhanced-type-promotion.md
@@ -70,7 +70,7 @@ propagation, but it handles the simplest, most common cases.
 
 ## How Type Promotion Works
 
-The language specifies type promotion out of three pieces:
+The language specifies type promotion out of four components:
 
 1.  **Certain expressions cause a fact to be known about a variable.** For
     example, `o` is String deduces the fact "`o` must be a String".
@@ -125,7 +125,12 @@ The language specifies type promotion out of three pieces:
     closure that is passed to it. To avoid cases like this, the spec defines
     certain variables to be off limits for type promotion.
 
-This proposal keeps the same three general pieces, but refines each of them.
+4.  **Which pairs of types -- declared and promoted -- are allowed for
+    promotion.** The spec says that promotion only applies if the promoted type
+    is a subtype of the declared type.
+
+This proposal keeps the same general pieces, but refines the first three of
+them. The forth is unchanged.
 
 ## Facts for False Expressions
 
@@ -218,7 +223,7 @@ has type T:
 ```dart
 while (o is String) oPromotedToStringHere;
 
-for (; o is String;;) oPromotedToStringHere;
+for (; o is String;) oPromotedToStringHere;
 ```
 
 ### Scopes After Exits
@@ -246,8 +251,8 @@ We say a statement "cannot complete normally" if the statement is one of:
 
 *   A block that directly contains a statement that cannot complete normally.
 
-We do not allowed *labeled* statements of the above form. Fortunately, labels
-are so vanishingly rare that this is unlikely to affect users in practice.
+We do not allow *labeled* statements of the above form. Fortunately, labels are
+so vanishingly rare that this is unlikely to affect users in practice.
 
 Then, given an if statement of the form `if (b) s1 else s2` that is contained in
 some block:
@@ -284,7 +289,8 @@ The specification places several restrictions on the variables whose type can be
     potentially mutated anywhere.
 
 This is a little too restrictive. We relax the restriction on potential
-assignments to allow assignments of the same type, as in:
+assignments to allow assignments inside the scope that promotion applies to if
+the assigned value's type is the same as the promoted type, as in:
 
 ```dart
 int stringLength(Object o) {
@@ -295,8 +301,9 @@ int stringLength(Object o) {
 }
 ```
 
-Here, since `trim()` returns a String, the assignment of `o` to its result does
-not disable its promotion.
+Here, the `if (o is String)` promotes `o` to String inside the body of the if.
+Since `trim()` returns a String which is the same as that promoted type, the
+assignment of `o` to its result does not disable its promotion.
 
 We couldn't relax this restriction before strong mode because the static type of
 expressions could not be trusted. Now that we have a sound type system, it's a
@@ -311,7 +318,9 @@ relaxation could be made for the locations of closures that capture the
 variable. This would require defining precisely what we mean by "before", which
 might be more work than it's worth.*
 
-*Optional: The restriction on mutation within the scope that the promotion would apply to could be relaxed by allowing the promotion to apply to everything up to the point of mutation. For example:*
+*Optional: The restriction on mutation within the scope that the promotion would
+apply to could be relaxed by allowing the promotion to apply to everything up to
+the point of mutation. For example:*
 
 ```dart
 if (v is T) {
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+expression evaluates to true. We extend this to include information that can be
+deduced when an expression evaluates to false. As with the true case, there are
+two places that we can do this.
+
+*   An is-expression `e` of the form `v is! T`, shows that `v` has type T when
+    `e` is false.
+
+*   A logical or expression `e` like `e1 || e2` shows that `v` has type T if
+    either `e1` or `e2` shows that when they are false if `e` itself is false.
+    For example, in `o is String || itsMonday`, we know `o` must be a String if
```
floitschG: `o is! String`
@johnniwinther

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+
+*   If `b` shows that `v` has type T when `b` is false, and `v` can be promoted,
+    and `s1` cannot complete normally, then `v` has the type T in the statements
+    after the if statement in the immediately enclosing block.
+
+    This is also allowed if the if statement has no else clause at all.
+
+*Optional: We could expand the definition of complete normally for a block to
+allow for returns or throws before the last statement as long as they are
+unconditionally executed.*
```
floitschG: We should. The following statements should not be treated differently, just because I had to add a block to make the code fit into 80 chars.

``` dart
if (e) return foo;
if (e) {
  return foo;
}
```

Adding curly braces is a style issue.

@johnniwinther

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+
+*Optional: The restriction on mutation within the scope that the promotion would apply to could be relaxed by allowing the promotion to apply to everything up to the point of mutation. For example:*
+
+```dart
+if (v is T) {
+  // v has type T here
+  v = someNonTValue;
+  // v has the non-promoted type here
+}
+```
```
floitschG: Would be good to have.

A common pattern is:
``` dart
Value v = ...
if (v is ProxyValue) {
  v = v.value;
}
```

@johnniwinther

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+        return o.length;
+      }
+    }
+    ```
+
+    After the invocation of `takeCallback()`, will `o` still be a String? Maybe.
+    It depends on whether the function passed to `stringLength()` calls the
+    closure that is passed to it. To avoid cases like this, the spec defines
+    certain variables to be off limits for type promotion.
+
```
floitschG: There is the 4th dimension: what type can be promoted (and how).

For example:
``` dart
Iterable<String> x = ...;
if (x is List) {
  ...
}
```

If we reduce the promotion to simple cases with only one constraint, then almost all cases are covered, and most users would be much happier. Dart2js seems to have some work in it.

@johnniwinther

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+}
+```
+
+We say a statement "cannot complete normally" if the statement is one of:
+
+*   A return statement
+*   A rethrow statement
+*   An expression statement whose expression is a throw expression
+*   An if statement with an else where neither statement can complete normally
+*   A block whose last statement cannot complete normally
```
lrhn: We might want to emphasize that a labeled block statement is not a block statement (it's a labeled statement, and might contain a break to the label - we can also statically determine if it does contain a break to the label, but there is no other reason to have a labeled block, so let's just assume that it does.

lrhn: No `break` or `continue`?
I know they are delimited, but so are throws (by `try/catch`), and the limit can always be found statically for `break` and `continue`.

I guess I'm angling for a proper (function-local) control flow analysis, and not more ad hocery.

docs/language/informal/enhanced-type-promotion.md
lrhn: Agree, we should say: A block statement where *any* of its statements cannot complete normally. The remaining statements would be dead code, but there is no reason to be more stupid than necessary in the presence of dead code.

lrhn: Those two are actually equivalent with the wording above because the return is the last statement of the block.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+int stringLength(Object o) {
+  if (o is String) {
+    o = o.trim();
+    return o.length;
+  }
+}
+```
+
+Here, since `trim()` returns a String, the assignment of `o` to its result does
+not disable its promotion.
```
lrhn: Allowing promotion precisely in the case where you assign the same type again feels ad-hoc.
It's basically saying that we *trust the assignment*, but we only allow it to promote to something we know already.

How about trusting assignments enough to actually let them promote 
the variable in all cases, to the type actually being assigned?

docs/language/informal/enhanced-type-promotion.md
lrhn: one of the "stupidities" of the current system is that it blows up if you have any assignment, even if the assignment is after the use-point.
I know that evaluation-order may be tricky (which is a good reason to fix the evaluation order as we are discussing elsewhere), and it basically amounts to abstract interpretation, but it would also be very useful.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+      return o.length; // Promoted here.
+    }
+    ```
+
+*   If `b` shows that `v` has type T when `b` is false, and `v` can be promoted,
+    and `s1` cannot complete normally, then `v` has the type T in the statements
+    after the if statement in the immediately enclosing block.
+
+    This is also allowed if the if statement has no else clause at all.
+
```
lrhn: I would really like the statement `x as T;` to promote `x` to `T`.
It's basically equivalent to `(x is T) ? x : throw new CastError()` which:
* promotes `x` to `T` in the true branch, and
* cannot complete normally in the false branch.
Any expression or statement that is down-flow from this can trust that `x` is `T`.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,537 @@
...
+For the purposes of type promotion, a statement cannot complete normally if the
+statement is either
+
+*   a return statement,
+*   a rethrow statement,
+*   an expression statement whose expression is a throw expression,
+*   an if statement of the form if (b) s1 else s2 where neither s1 nor s2 can
+    complete normally, or
+*   a block whose last statement cannot complete normally.
```
lrhn: (Or a do-while statement where the body cannot complete normally, but it's probably ridiculous enough that we shouldn't care).

lrhn: Again, I'd add a `break` or `continue` statement.

docs/language/informal/enhanced-type-promotion.md
munificent: Yeah, I think this is super important. The pattern you note Florian is really common, especially with parameters.

docs/language/informal/enhanced-type-promotion.md
munificent: Fixed.

docs/language/informal/enhanced-type-promotion.md
munificent: Agreed. Done.

docs/language/informal/enhanced-type-promotion.md
munificent: I'm up for that, though I worry about how break/continue to a label interferes with this. @bwilkerson, do you think we should add this?

docs/language/informal/enhanced-type-promotion.md
munificent: I'm not opposed to that, but we're trying to keep this proposal small and incremental so that we can get it done in a reasonable amount of time. We'd like to get rid of analyzer's special type propagation code as soon as possible.

Do you think the `as` case comes up enough to warrant adding it now? I can't recall running into it.

docs/language/informal/enhanced-type-promotion.md
munificent: I would like full control flow analysis too, *eventually*. But I'd like to not go there yet because that's going to be a *lot* of work. Can we stick with a smaller subset that covers the common cases in this proposal and then refine it later?

I think we will end up revisiting promotion anyway to handle non-nullable types and possible intersection and union types. That would be a good time to do the full flow analysis.

munificent: Done.

docs/language/informal/enhanced-type-promotion.md
bwilkerson: I chose not to add it to the original proposal because (a) I wanted to keep this as simple as possible, (b) I don't think that case occurs frequently, and (c) it isn't necessary to add all possible enhancements at this point.

I'd leave it out for now, unless adding it will allow us to get this approved more quickly :-)

docs/language/informal/enhanced-type-promotion.md
lrhn: It doesn't come up because it doesn't help you right now, but I expect that it would be useful *if* it causes type promotion, as an assertion-like promotion syntax.

Compare:
```
_foo(Object x) {
  assert(x is Foo);  // no promotion
  ...
}
_bar(Object x) {
  Foo fooX = x;  // unnecessary new name
  ...
}
_baz(Object x) {
  if (x is! Foo) throw ... something;  // Long an cumbersome
  ...
}
_qux(Object x) {
  x as Foo;
  ...;  // Type promotion of x, short and convenient.
}
```
We could even say that `covariant` desugars into an initial `x as Type;` statement.


So, not necessary, but I think it will be very convenient, so at least keep it in mind.

docs/language/informal/enhanced-type-promotion.md
lrhn: If we just go for `break`, it should be safe, as long as we never consider labeled statements to not complete normally.

The only complication I can see is that a block-statement that is the body of a loop can complete normally if it contains a `continue`. We need to handle that *anyway*, because that applies even if the last statement of the block is a return:
```
while (x is! T) {
  if (test) continue;
  if (otherTest) break;
  return 42;
}
// x is T?
```
(I don't remember if we do the contra-indicated promotion on `while` loops the same way as on `if` statements, but if we do, this is a problem. It's also a problem that can be easily solved by statically checking if the loop contains a `break` or `continue` targeting it).

Apart from loop bodies, there is not problem. A `break` or `continue` always terminate the current block, just like a `throw`. If a loop, `switch` or `try` statement is never considered not completing normally, then having control flow going to them from unpredictable places is not a problem.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+    if statement, but not the else clause, and not after the end of the if
+    statement.
+
+3.  **Which variables can have their types promoted.** Some code may cause a
+    fact to no longer be true. In that case, we shouldn't promote it. A simple
+    example is:
+
+    ```dart
+    if (o is String) {
+      o = 123; // Not a string anymore!
```
leafpetersen: Just a thought, but should we really not promote `o` here?  I suspect that the original argument was that promotion should not introduce errors, only eliminate them.  But we're already changing that by allowing promotion from `dynamic`.  So maybe we should just allow in scope re-assignments, and let the chips fall where they may.  The code above becomes an error.

docs/language/informal/enhanced-type-promotion.md
munificent: I'll definitely keep it in mind.

docs/language/informal/enhanced-type-promotion.md
munificent: I didn't follow this. What do you mean by "constraint"?

docs/language/informal/enhanced-type-promotion.md
munificent: Consider:

```dart
foo(Object o) {
  o.length; // Should have error!
  o = "string now";
  o.length; // This would be OK.
}
```

If we allow assignment promoting to other types, then we need to split the scope at each assignment so that we don't promote until *after* the assignment. That's reasonable, but I think it would make this proposal quite a bit more complex.

I do think we want to get there eventually, but I'm hoping to not do so yet. I'd rather defer the interesting flow-sensitive analysis later and try to do it holistically. This proposal is more about what's the minimal amount we can do shortly so that we can eliminate the ad-hoc type propagation that analyzer is doing?

Is that OK?

docs/language/informal/enhanced-type-promotion.md
munificent: See [my comment above](https://github.com/dart-lang/sdk/pull/29624#discussion_r119222452). I'd like to go here eventually, but probably not in this proposal.

docs/language/informal/enhanced-type-promotion.md
bwilkerson: Analyzer's type propagation currently re-assigns the type on assignment (though without the limitations used by promotion). It appears (based on issues opened) to be one of the most confusing aspects of promotion for most users. Perhaps the extra limitations would negate the confusion, but be aware that we've previously run into issues related to this.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+
+*   A rethrow statement.
+
+*   An expression statement whose expression is a throw expression.
+
+*   An if statement with an else where neither statement can complete normally.
+
+*   A block that directly contains a statement that cannot complete normally.
+
+We do not allowed *labeled* statements of the above form. Fortunately, labels
```
lrhn: "do not allowed" -> "do not allow"
... but I'm still not sure what that means.
Maybe: 
"Notice that this does not include a labeled block statement, or a loop statement with a block body, only the block statement itself."

I really think we should include break/continue as well:

  for (var object in objects) {
     if (object is! Future) {
       print(object);
       continue;
     }
     object.then(print);
   }

How do we explain that this doesn't work but replacing `continue` with `return` does. That is an inconsistency that we have to explain, where "if execution of the positive-test branch will definitely not reach the end of the after the i, then the following code is considered to be on the negative test-branch."

docs/language/informal/enhanced-type-promotion.md
bwilkerson: @floitschG

floitschG: In this case, it's pretty clear, that we want to promote the `Iterable<String>` to a `List<String>`. However, the type promotion doesn't even try to.
The following example is currently invalid:

``` dart
Iterable<int> foo() {
  return [1, 2];
}

main() {
  Iterable<int> x = foo();
  if (x is List) {
    print(x[0] + 3);
  }
}
```

@johnniwinther (correctly) stated that this is a 4th dimension in this list.

docs/language/informal/enhanced-type-promotion.md
floitschG: If we all think that this is important, we should not make it optional.

docs/language/informal/enhanced-type-promotion.md
leafpetersen: Sure.  Note though that I'm not really suggesting the same thing that you addressed in your comment above: I'm not suggesting promoting based on the assignment, I'm suggesting just allowing promotion (from the `is` test) even if there is an assignment, and then allowing the assignment to become an error if it changes the type.

bwilkerson: I think that's a much easier semantic to explain to users than either (a) allowing the type to change after an assignment (the current analyzer propagation model) or (b) splitting the scope so that it has a promoted type in some of the code but not in other code. It would be interesting to run an analysis to figure out how much code would be effected by that change in semantics.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+variables and is dependent on the use of the variable in the scope in which the
+promotion would occur. If a variable is one for which type promotion is allowed,
+we say that the variable is promotable in some scope. A variable v is promotable
+in the scope S if all of the following conditions are met:
+
+*   v is either a local variable or a parameter,
+
+*   v cannot be potentially mutated within S unless all of the types of the
+    values that would be assigned to v within S are a subtype of the promoted
+    type of v,
```
stereotype441: I think this is undecidable in general.  For instance, consider:

    if (x is T) {
      x = f(x);
    }

Is `x` promoted within the block?  To answer that we need to know whether `f(x)` evaluates to a subtype of `T`.  If `f` is a generic function, then to answer that question we need to perform type inference on `f(x)`, which requires knowing the type of `x`.  To know the type of `x`, we need to know whether it is promoted within the block.  So we are at an impasse.

The two bullet points that follow this one have similar problems.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+*   A return statement.
+
+*   A rethrow statement.
+
+*   An expression statement whose expression is a throw expression.
+
+*   An if statement with an else where neither statement can complete normally.
+
+*   A block that directly contains a statement that cannot complete normally.
+
```
bwilkerson: My personal opinion is that the specification and the implementation need to agree, because without that I believe that it becomes confusing to users. That could mean updating the spec to match the implementation, or vice versa.

docs/language/informal/enhanced-type-promotion.md
munificent: OK, added some verbiage around this.

docs/language/informal/enhanced-type-promotion.md
munificent: I believe the current proposal already covers this case. The wording ("same type") wasn't clear, but the example was analogous to this one.

I updated the wording to be clear that "same type" means the same as the promoted type.

This doesn't, I think, require abstract interpretation or changing the type of a variable after the assignment.

docs/language/informal/enhanced-type-promotion.md
munificent: Thanks, that's useful to keep in mind.

munificent:
> I don't remember if we do the contra-indicated promotion on while loops the same way as on if statements, but if we do, this is a problem.

The proposal doesn't advocate that and, anecdotally, I don't think it comes up too often in practice.

docs/language/informal/enhanced-type-promotion.md
munificent: > I don't remember if we do the contra-indicated promotion on while loops the same way as on if statements, but if we do, this is a problem.

The proposal doesn't go for that and, anecdotally, I don't think it comes up too often in practice.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+### Loop Scopes
+
+Much like if statements, we extend C-style for and while loops to promote inside
+their bodies when the condition is true and the condition expression shows a `v`
+has type T:
+
+```dart
+while (o is String) oPromotedToStringHere;
+
+for (; o is String;;) oPromotedToStringHere;
```
munificent: If an assignment in the loop body doesn't nullify the promotion then, yes, it seems to me like it should.

@bwilkerson?

docs/language/informal/enhanced-type-promotion.md
munificent: Yeah, I think it should match.

I would up for updating the proposal to match what you describe here. I like the idea of having a unified fact formalism that covers both flavors of facts, and I think tracking reachability at the expression level will end up being useful.

What do you think, @bwilkerson? I'll have to do some homework first, but I could try to sketch out some informal prose to define reachability like @stereotype441 describes.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+*   b shows that a variable v has type T when b is false,
+
+then the type of v is known to be T in s2. If the if-statement in enclosed in a
+block and if s1 cannot complete normally, and v is promotable in the statements
+(S) in the immediately enclosing block that follow the if statement, then v is
+known to have type T in S.
+
+### 17.6.1 For Loop
+
+Given a for statement of the form for (..., c, ...) s, if
```
munificent: Fixed.

docs/language/informal/enhanced-type-promotion.md
bwilkerson: I don't have any problem with promoting `o` in the update. I'm not sure whether this pattern occurs much (I don't recall seeing it before), but it won't hurt anything to add it. But again, I'm less concerned with covering every possible case than with covering the most common cases.

docs/language/informal/enhanced-type-promotion.md
stereotype441: Incidentally, my prototype was inspired by definite assignment analysis, so if you're going to sketch out an informal proposal I'd recommend using definite assignment analysis as a starting point.  I took some shortcuts in the presence of loops to avoid the need to compute a fixed point.

Note that the current (Dart 1.0) type promotion rules can all be re-cast into a definite assignment style framework without changing the current semantics, so if we updated the spec to use a definite assignment style framework we would be able to start adding type promotion features by making parallel incremental changes both in the spec and in the front end implementation. 
 I think that would be really desirable.

I can give you more details about my prototype algorithm any time--let me know if you want to set up a meeting.

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -0,0 +1,540 @@
...
+*   c shows that v has the type T when true,
+
+then v is known to have type T in s.
+
+Given a for statement of the form for (..., c, u) s, if
+
+*   the variable v is promotable in u, and
+*   c shows that v has the type T when true,
+
+then v is known to have type T in u.
```
munificent: Aye.

docs/language/informal/enhanced-type-promotion.md
bwilkerson: IIRC, Lasse was advocating the same thing in the original doc. My only concern then, as now, is the complexity that might add. My primary concern is that this needs to be comprehensible to users (who are not language designers). That said, if you want to take a stab at writing something up that will give us a concrete example to look at to judge the comprehensibility of what it would require.

(IIRC, the original proposal was that every statement and expression have text added of the form: "This construct completes normally if ...", thus defining the notion recursively. Not sure whether that's the most readable form. I think the Java spec has a separate chapter on the subject just to bring all of the parts of the definition closer together.)

bwilkerson: Just to clarify, "same thing" did not mean "definite assignment analysis" when I typed it; it meant "reachability analysis".

docs/language/informal/enhanced-type-promotion.md
munificent: :anguished:

You're exactly right. Digging into it, I managed to come up with this fun one:

```dart
T fn<T, T>(T t, T t) => t;

class A           {  D v; }
class B extends A { E2 v; }
class C extends B {}

class D            { A a; }
class E1 extends D { C a; }
class E2 extends D {}

foo(A x) {
  if (x is B) {
    var y = x.v;
    if (y is E1) {
      x = fn(y.a, new B());
    }
  }
}
```

Here, if we allow the promotion of `x` to B, then it kills the promotion of `y` to E1. If we allow the promotion of `y` to E1, it kills the promotion of `x` to B. So there's stable solutions where either `x` or `y` is promoted (but not both) and the choice is arbitrary. At least, I think that's how it works.

I'd hate for the perfect to be the enemy of the good and go back to making all assignments kill promotion when in most cases, the assignment is innocuous and doesn't depend on or affect inference. But I'm stumped as to how we can identify "safe" and "unsafe" assignments without using any type analysis on the RHS.

Any ideas?

@leafp @bwilkerson

docs/language/informal/enhanced-type-promotion.md
```diff
@@ -125,7 +125,12 @@ The language specifies type promotion out of three pieces:
     closure that is passed to it. To avoid cases like this, the spec defines
     certain variables to be off limits for type promotion.
 
-This proposal keeps the same three general pieces, but refines each of them.
+4.  **Which pairs of types -- declared and promoted -- are allowed for
+    promotion.** The spec says that promotion only applies if the promoted type
+    is a subtype of the declared type.
+
```
johnniwinther: The spec says more specific, not subtype. This is why you cannot promote `Iterable<String>` to `List`.  In simple cases, like this one, there is only one way for a `List` to be an `Iterable<String>`, namely a `List<String>` thus promoting to `List<String>` would avoid the need for an intersection type (`Iterable<String>&List`) in this case.

docs/language/informal/enhanced-type-promotion.md
johnniwinther: The basic problem is the definition of scope. In [Guarded Type Promotion](http://cs.au.dk/~jwbrics/papers/GuardedTypePromotion/paper.pdf) (see section 4, rule for assignment) the scope extends until the first assignment, and a 'reduced' promotion after the assignment. 

This allows for continued promotion after assignment if the assignment upholds the promotion. For instance
```
if (x is String) {
  x = x.trim(); // promotion is continued since .trim() returns String
  x = x.length; // promotion is terminated since .length has type int
}
```

In the example above `x` is promoted to `B` until the assignment - or even after the assignment if we solve the orthogonal question of which pairs of types that can be promoted (component 4); can we promote `y` to `E1&E2` or to `E1` since it was only _inferred_ to be `E2`.

docs/language/informal/enhanced-type-promotion.md
munificent: > The spec says more specific, not subtype.

I think that's the 1.0 spec, but for strong mode / 2.0, I *believe* the intent is to change it to subtype. Is that right, @leafpetersen?

leafpetersen: Yes.

docs/language/informal/enhanced-type-promotion.md
munificent: I worry (and I think Brian has evidence to support it) that changing (by demoting or refining) the type on assignment leads to a more confusing user experience, even if we sidestep intersection types.

For now, at least, I like Leaf's suggestion of having the assignment simply not change the type. Instead, we type check the assigned value against the variable's (now promoted) type and report an error as usual if that's not valid. So, in your example, you'd get an error on `x = x.length` because you're trying to assign an int to a variable of (promoted) type String.

lrhn: I don't think having only one type for a variable is "good enough", we will need to have something similar to intersection types ("multiple independent known types") in order to not look *too* stupid.
If I do `if (x is C) { x.methodFromC(); }` it shouldn't fail. If we can only promote to subtypes of the previously known type, then it can fail if `C` is an unrelated type.

I know that it's tempting to go for "the simplest thing that could work", but that's what got us where we are now (a promotion system that users rightly find mindbogglingly stupid). 
The rules are pretty simple, it's the *exceptions* that hurt, so I want to remove exceptions. An is-test should *always* promote a local variable, without exception. (I'd even argue that it should promote other things than local variables, but that's more tricky).
That means promoting even if the type is unrelated to the static type or other currently known types, because that's easier than explaining to the user when it works and when it doesn't.
For the same reason, I want to be able to know multiple independent types, because there is no good unification of independent types, but the user is able to "know" that `x` is both `B` and `C`.
If one of `B` and `C` is more specific than the other, we *can* unify them, but we should be able to handle the case where we can't.

eernstg (commented):
We discussed this a long time ago, but I promised to take a fresh look at the document.

In general, I think it looks fine. The main issue I can see is that the notion of being "after" some piece of code is complex and somewhat ad-hoc. I really like Paul's approach, but we need to specify it, and I'd recommend that we consider enhancing promotion in two steps: (1) specify and implement simple scopes of promotion (`{.. "x promoted here" ..}`) and (2) specify and implement irregular scopes of promotion (where propotion takes place "after" some point where it is guaranteed that `v is T` because all other situations would force a non-normal completion).

lrhn:
Erik's comment on the "comes after" specification is correct.

I like to think of it as "a test/branch dominates another expression/statement" when all routes of execution from the test to the expression requires the test to have the same value. That is, when reaching the expression/statement implies the result of the test.

Maybe that's a way to specify it. Needs more thinking obviously :)

(Or am I just going down the dark and quirky path of abstract interpretation here :)

mit-mit:
@munificent @lrhn what is the status of this?

munificent:
Back when the final feature list for Dart 2.0 was put together, this didn't make the cut. It's been inactive since then.

munificent:
> open an issue in the language repo for improved type promotion (but don't make it active yet)

https://github.com/dart-lang/language/issues/80

> land this PR in the language repo working directory as a strawman

https://github.com/dart-lang/language/pull/79

> link the strawman from the issue to track it

Done.

PR #29662, opened 2017-05-19
`new Widget` not `new widget`

Skylled:
It seems like these two assists have a lot of work to be done on them, but I can at least fix these typos.


stevemessick:
That's not a typo. "Widget" is never correct but it appears it might be. "widget" looks wrong because it is. That said, maybe the reasoning behind the decision is incorrect. We'll have to think about that.

PR #29684, opened 2017-05-22
Fix inifinite loop in analysis server.

ianloic:
It was walking up a directory heirarchy looking for a directory named
"lib" but didn't terminate when it reached "/" without finding "lib".

### DIFF
--- a/pkg/analysis_server/lib/src/plugin/plugin_watcher.dart
+++ b/pkg/analysis_server/lib/src/plugin/plugin_watcher.dart
@@ -132,8 +132,12 @@ class PluginWatcher implements DriverWatcher {
   String _getSdkPath(AnalysisDriver driver) {
     AbsolutePathContext context = resourceProvider.absolutePathContext;
     String sdkRoot = driver.sourceFactory.forUri('dart:core').fullName;
-    while (sdkRoot.isNotEmpty && context.basename(sdkRoot) != 'lib') {
-      sdkRoot = context.dirname(sdkRoot);
+    while (context.basename(sdkRoot) != 'lib') {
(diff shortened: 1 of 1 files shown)


pkg/analysis_server/lib/src/plugin/plugin_watcher.dart
```diff
@@ -132,7 +132,7 @@ class PluginWatcher implements DriverWatcher {
   String _getSdkPath(AnalysisDriver driver) {
     AbsolutePathContext context = resourceProvider.absolutePathContext;
     String sdkRoot = driver.sourceFactory.forUri('dart:core').fullName;
-    while (sdkRoot.isNotEmpty && context.basename(sdkRoot) != 'lib') {
+    while (sdkRoot.isNotEmpty && sdkRoot != '/' && context.basename(sdkRoot) != 'lib') {
```
bwilkerson: Unfortunately, this won't work on Windows. Perhaps revert the condition and replace the body of the `while` with
```
String parent = context.dirname(sdkRoot);
if (parent == sdkRoot) {
   break;
}
sdkRoot = parent;
```

bwilkerson (commented):
Thanks for catching this!

pkg/analysis_server/lib/src/plugin/plugin_watcher.dart
scheglov: This does not fix the same problem for Windows.

scheglov (changes requested):

ianloic:
I'm bad at pull requests. There's a new one that should work on windows over here: https://github.com/dart-lang/sdk/pull/29685

PR #29735, opened 2017-05-26
Mark analyzer_cli/test/driver_test as slow.

nex3:
This was already marked slow on Windows, but I saw a timeout on Linux as
well so it seems best to just mark it slow for all platforms.


pq (approved):
SGTM.  Thanks!

Issue #25795, opened 2016-02-17
Non-deterministic JS output in core_runtime_types_test


asgerf:
The JS output for `core_runtime_types_test` fails about 10% of the time, and not always with the same stack trace.

To reproduce, compile `tests/corelib/core_runtime_types_test.dart` with CPS and run it 100 times:

``` bash
for i in $(seq 100); do d8 out.js; done
```

Example output:

```
out.js:2754: Expect.throws: Unexpected 'Error: undefined is not a function'
TypeError: undefined is not a function
    at J.$ge$n (out.js:2847:41)
    at CoreRuntimeTypesTest_testComparisonOperatorErrors_closure2.dart.CoreRuntimeTypesTest_testComparisonOperatorErrors_closure2.call$0 (out.js:1792:13)
    at dart.Expect_throws (out.js:2743:11)
    at dart.CoreRuntimeTypesTest_testBinaryOperatorErrors (out.js:1532:9)
    at dart.main (out.js:1575:15)
    at out.js:3152:9
    at init.currentScript (out.js:3132:7)
    at out.js:3147:5
    at out.js:3155:3
      throw H.wrapException(new K.ExpectException(message));
      ^
Expect.throws: Unexpected 'Error: undefined is not a function'
TypeError: undefined is not a function
    at J.$ge$n (out.js:2847:41)
    at CoreRuntimeTypesTest_testComparisonOperatorErrors_closure2.dart.CoreRuntimeTypesTest_testComparisonOperatorErrors_closure2.call$0 (out.js:1792:13)
    at dart.Expect_throws (out.js:2743:11)
    at dart.CoreRuntimeTypesTest_testBinaryOperatorErrors (out.js:1532:9)
    at dart.main (out.js:1575:15)
    at out.js:3152:9
    at init.currentScript (out.js:3132:7)
    at out.js:3147:5
    at out.js:3155:3
    at dart.wrapException (out.js:773:18)
    at dart.Expect__fail (out.js:2754:15)
    at dart.Expect_throws (out.js:2749:11)
    at dart.CoreRuntimeTypesTest_testBinaryOperatorErrors (out.js:1532:9)
    at dart.main (out.js:1575:15)
    at out.js:3152:9
    at init.currentScript (out.js:3132:7)
    at out.js:3147:5
    at out.js:3155:3

out.js:2754: Expect.throws: Unexpected 'Error: undefined is not a function'
TypeError: undefined is not a function
[...]

PR #29916, opened 2017-06-16
Mark core_runtime_types_test as flaky on D8.

nex3:
See #25795


mit-mit:
@nex3 @rakudrama this appears to be stale?

PR #29955, opened 2017-06-20
Informal proposal for making "await" and "yield" reserved words

munificent:


stereotype441:
The spec currently says:

> It is a compile-time error if any of the identifiers `async`, `await` or `yield` is
used as an identifier in a function body marked with either `async`, `async*` or
`sync*`.

Can we make it part of this proposal to strike this paragraph?  It's no longer needed for `await` and `yield` (since we are making them reserved words), and AFAICT, allowing `async` to be used as an identifier introduces no ambiguity (since the special meaning of marking a method as `async` can only occur between a function's formal parameters and its body).

munificent:
@stereotype441, done.

docs/language/informal/reserve-await-and-yield.md
```diff
@@ -0,0 +1,244 @@
...
+## `await` and `yield`
+
+The last two problems apply in particular to `await` and `yield`. Most
+contextual keywords can be disambiguated using very local context. `hide` is
+only a keyword when right inside an import directive. `of` is only a keyword if
+it immediately follows `part` at the top level.
+
+But `await` is a keyword if the surrounding method is declared `async`. That
+modifier may be very far away from the use of `await`. `yield` is similar for
+`sync` methods.
```
lrhn: -> `sync*` methods 
Or better -> generator methods
(that covers both `sync*` and `async*`).

docs/language/informal/reserve-await-and-yield.md
```diff
@@ -0,0 +1,244 @@
...
+
+## What about other contextual keywords?
+
+That still leaves a long list of contextual keywords. Should we reserve any of
+those while we're at it? The tools folks who requested we reserve `await` and
+`yield` also requested `async`. However:
+
+*   "async" is the name of a prominent Dart library, "dart:async". If we ever
+    get a [syntax for imports that doesn't require a quoted string][import],
+    that would cause problems if `async` was a reserved word.
```
lrhn: Probably not. We will not require library names to be identifiers, just character sequences delimited by context.
I.e., this is not a problem for us, but it may still be one for syntax highlighters.

docs/language/informal/reserve-await-and-yield.md
```diff
@@ -0,0 +1,244 @@
...
+
+*   "async" is more commonly used as an identifier than `await` and `yield`. I
+    count 11,046 occurrences in the same corpus as above.
+
+*   "async" appears in the exact same place as "sync" in the grammar, so it
+    would be strange to reserve one without the other. Unfortunately, "sync"
+    shows up as an identifier in some important places, such as `Future.sync()`.
```
lrhn: No count of `sync`-as-an-identifier? (Just curious!) :)

lrhn (approved):
LGTM, ship it!

docs/language/informal/reserve-await-and-yield.md
```diff
@@ -0,0 +1,263 @@
...
+
+[import]: https://github.com/dart-lang/sdk/issues/10018
+
+We could consider reserving some of the other contextual keywords, but so far we
+have had any user or tool author requests to do so, so they don't seem to be
```
eernstg: Typo --> 'have not had'?

eernstg (commented):
LGTM, with a single remark about a typo. We might hesitate because it's "a new thing" to add reserved words at all, or because we'll create problems for a few developers who actually use those identifiers (maybe even in a public interface), but this document is certainly a good presentation of the case for adding these reserved words.

docs/language/informal/reserve-await-and-yield.md
```diff
@@ -0,0 +1,244 @@
...
+
+## `await` and `yield`
+
+The last two problems apply in particular to `await` and `yield`. Most
+contextual keywords can be disambiguated using very local context. `hide` is
+only a keyword when right inside an import directive. `of` is only a keyword if
+it immediately follows `part` at the top level.
+
+But `await` is a keyword if the surrounding method is declared `async`. That
+modifier may be very far away from the use of `await`. `yield` is similar for
```
lrhn: `yield` -> The `yield` identifier 
(to avoid starting a sentence with a lower-case letter).

munificent: Done.

docs/language/informal/reserve-await-and-yield.md
munificent: Done.

docs/language/informal/reserve-await-and-yield.md
munificent: > We will not require library names to be identifiers, just character sequences delimited by context.

I'm not totally convinced of that. Turning *all* of our keywords into contextual keywords seems like creating pain for little value. Pub has always reported an error if you try to use a reserved word in a package name to try to avoid this case.

docs/language/informal/reserve-await-and-yield.md
munificent: 2,212. Added it. :)

docs/language/informal/reserve-await-and-yield.md
munificent: Oops, yes! Fixed.

mit-mit:
@munificent can we merge this given it's approved?

eernstg:
To me it seems useful to flag those situations where a non-async function contains `await` and, accidentally, it looks like an invocation of an undefined function&mdash;and then the developer gets confused about the error message.

However, if we're worried about introducing new reserved words, especially with the policy that puts extremely strong constraints on breaking changes, we could also address the same issue as a lint.

mit-mit:
I thought the conclusion was that this practically wan't very breaking? From the proposal:

> In other words, 0.0001% of identifiers are `await` and 0.0003% are `yield`. A
random use of `await` is 3389 times more likely to be a keyword than an
identifier.

munificent:
Yes, I think in practice it's unlikely to break almost all code. I still think it's a good *idea*, but the current *status* is that it was never approved so it didn't land and wasn't implemented.

@leafpetersen and @lrhn, is this something you think is worth reviving?

matanlurey:
This was identified as a top usability issue in internal trainings by @TedSander and myself.

If it can still be fixed (even in Dart 3, perhaps hinting in Dart 2 it will become a reserved word), that would be great.

leafpetersen:
I thought we'd agreed to do this, but I guess maybe it didn't get done in time?

munificent:
Yeah, I don't recall what the resolution was, if any. I think it was sort of left hanging. If we still want to do it, I still think it's a good idea. I don't know if it's feasible now that 2.0.0 is done, but it might be safe enough to get into a 2.x.

mit-mit:
@munificent can we merge this, AND create a `request` tracking issue in the language repo to make this change?

lrhn:
I would just land it in the language repo directly. There is no need for an "informal" specification in the SDK of a potential feature that hasn't even been decided yet.

Since it is a breaking change, I think we'll have to go through the proper procedures, and not make the change until we have 1) a migration tool of some sort, and 2) language versioning so migration can be gradual.

mit-mit:
> I would just land it in the language repo directly. There is no need for an "informal" specification in the SDK of a potential feature that hasn't even been decided yet.

SGTM

> Since it is a breaking change, I think we'll have to go through the proper procedures, and not make the change until we have 1) a migration tool of some sort, and 2) language versioning so migration can be gradual.

Fully agree. But let's get the request bug opened now so that we don't forget about it.

munificent:
This didn't make it into 2.0, so we aren't actively working on it. Moved the proposal [here](https://github.com/dart-lang/language/pull/174).

PR #30078, opened 2017-07-05
Fix build_pkgs.dart to not mutate arguments list.

munificent:
Also remove build of when package since it's apparently gone now.

### DIFF
--- a/pkg/dev_compiler/tool/build_pkgs.dart
+++ b/pkg/dev_compiler/tool/build_pkgs.dart
@@ -17,7 +17,10 @@ String outputDirectory;
 /// Otherwise, only builds the modules needed by the tests.
 void main(List<String> arguments) {
   var isTravis = arguments.isNotEmpty && arguments.last == "travis";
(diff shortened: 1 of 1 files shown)


pkg/dev_compiler/tool/build_pkgs.dart
```diff
@@ -17,7 +17,10 @@ String outputDirectory;
 /// Otherwise, only builds the modules needed by the tests.
 void main(List<String> arguments) {
   var isTravis = arguments.isNotEmpty && arguments.last == "travis";
-  if (isTravis) arguments.removeLast();
+  if (isTravis) {
+    arguments = arguments.toList();
+    arguments.removeLast();
```
vsmenon: arguments.sublist for a one-liner?

vsmenon (approved):

pkg/dev_compiler/tool/build_pkgs.dart
munificent: Oh, good idea. Done.

PR #30098, opened 2017-07-07
Properly handle Windows file: URLs on Node.js

nex3:
This adds logic to the JS implementation of URI to determine whether
the code is running on Windows under Node.js.

### DIFF
--- a/pkg/dev_compiler/tool/input_sdk/patch/core_patch.dart
+++ b/pkg/dev_compiler/tool/input_sdk/patch/core_patch.dart
@@ -604,8 +604,11 @@ class _Uri {
   @patch
   static bool get _isWindows => _isWindowsCached;
 
-  static final bool _isWindowsCached =
-      JS('bool', 'process !== undefined && process.platform == "win32"');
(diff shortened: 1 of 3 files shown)


pkg/dev_compiler/tool/input_sdk/patch/core_patch.dart
```diff
@@ -602,7 +602,10 @@ class Uri {
 @patch
 class _Uri {
   @patch
-  static bool get _isWindows => false;
+  static bool get _isWindows => _isWindowsCached;
+
+  static final bool _isWindowsCached =
+      JS('bool', 'process !== undefined && process.platform == "win32"');
```
lrhn: I have to check whether "win32" is sufficient (since I'm generally running 64-bit Windows).

lrhn (changes requested):
LGTM if the test is fixed.

pkg/dev_compiler/tool/input_sdk/patch/core_patch.dart
nex3: Done.

PR #30106, opened 2017-07-07
Use "|" as the separator between module path and name.

munificent:
":" is already a path character in Windows. This is why the
dartdevc tests aren't working on Windows. Because they try to do:

-sC:/some/path/blah.dart:blah

And it splits at the first ":". Oops.

### DIFF
--- a/pkg/dev_compiler/lib/src/analyzer/context.dart
+++ b/pkg/dev_compiler/lib/src/analyzer/context.dart
@@ -113,7 +113,7 @@ class AnalyzerOptions {
     return mappings;
   }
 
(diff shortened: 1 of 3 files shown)


jmesserly:
FYI -- this is possibly a breaking change for build. Can you follow up and fix build scripts if needed?

jmesserly (changes requested):
"|" will break unix shells, right? Are there any other options?

munificent:
How about `=`?

jmesserly (approved):
sounds good to me!

PR #30139, opened 2017-07-11
Update status file for now-passing DDC tests.

munificent:


vsmenon (approved):
I need to remember to update these!

PR #30323, opened 2017-08-04
Change DDC presubmit/Travis to use test.dart, similar to other Dart bots

jmesserly:
Sending this as a pull request to get travis to test it.

This might need some additional tweaks to have Travis build the summaries for the test packages, or other build/install prerequisites.


jmesserly:
(I've repushed the branch, this only included the SDK.js rebuild, not the presubmit changes)

jmesserly:
Putting this on the backburner for now. It looks like test.dart requires building the SDK which would really hurt travis cycle times.

Issue #30370, opened 2017-08-08
Analyzer parser recovery: premature closing labels created by code errors


DanTup:
I'm working in implementing a "closing labels" feature in the analysis server (#30285). During testing I noticed that this bad code results in labels being inserted in a bad place:

![Bad label positioning](https://user-images.githubusercontent.com/1078012/29089998-22719258-7c76-11e7-85fd-7464a9f92d9e.png)

The greyed "comments" are being inserted on the line where the associated nodes are ending (according to the AST), the highlighted code is the code I accidentally inserted where a type name should be. It seems like it should be possible to do a better job here (I believe it's the semi-colon throwing things off; but there's a compile error that the semicolon is not expected and if you take it away, I believe the nodes end at the correct places).

@bwilkerson suggests this might not be worth tackling in the current parser, though it might be a good cast to handle/test with the new parser.

bwilkerson:
@danrubel @lukechurch

DanTup:
Here's another one (lmk if you want another case).. Adding ` a` onto the end of this `null` caused a similar thing - a load of invocation nodes to end too early (again, the ` a` is unexpected and if you ignore it completely, you should get the correct endings).

<img width="962" alt="screen shot 2017-08-11 at 14 23 22" src="https://user-images.githubusercontent.com/1078012/29214899-cb455302-7ea0-11e7-8f12-31b287683290.png">

DanTup:
These things are all handled much better now:

<img width="702" alt="Screenshot 2022-05-24 at 12 00 19" src="https://user-images.githubusercontent.com/1078012/170020827-10d392e9-34e8-4dbe-a1ae-b454a00c6ac1.png">

<img width="606" alt="Screenshot 2022-05-24 at 12 05 52" src="https://user-images.githubusercontent.com/1078012/170020834-cf55f4a7-f5f3-42f7-8aea-e12d2d9c7bf0.png">

Issue #30285, opened 2017-07-27
Requesting analyzer support for "trailing/closing tag" feature


sethladd:
We quickly explored the usefulness of ["closing tags" for Flutter code](https://github.com/Dart-Code/Dart-Code/issues/383), and our initial users seem to find it helpful. We'd like to explore how to do this "for real" which we believe requires help/support from the analyzer, in order to improve robustness of the feature and help make it available in IntelliJ.

I'll defer to @DanTup for further explanation and technical requests.

DanTup:
The current implementation linked above uses getHovers to retrieve offset/length information for the constructor calls. To know where constructors are it uses a combination of highlights and outlines. The whole thing is rather inefficient!

The idea is to implement a new notification in the analysis server that can provide all of the require info for rendering these decorations for an entire file in a single message - something like an array of `int offset/int length/string label`.

With assistance, I hope to contribute this feature :-)

DanTup:
This work is now pretty much complete (PRs #30332, #30372 and #30411). There may be some tweaks but I believe the API is stable (or at least, future changes will be non-breaking).

There are some cases where invalid code may result in misplaced labels (see #30370) though hopefully parser improvements will correct these in future.

@bwilkerson I think we're both happy any future changes will be non-breaking, so is it possible to bump the analyzer API version now? I think I'd like to ship an update to Dart Code to switch to this (though still behind a flag) to get some more feedback (eg. @devoncarew has concerns about this appearing in too many places).

bwilkerson:
Alex seems to think the API is good for IntelliJ as well, so yes, I think we should bump the API level and call this done. Thanks for adding this!

DanTup:
@bwilkerson Is it just a case of updating the spec and then re-running the generation script? I can send a PR, or if it's easier for you to do, feel free. Thanks!

bwilkerson:
There's one other place to update, so I did it myself (https://codereview.chromium.org/3002513002/).

DanTup:
Excellent; thanks!

> while still marked experimental

I'm not sure if you saw [my previous comment](https://github.com/dart-lang/sdk/pull/30332#issue-248250214) about this - experimental didn't seem to work on this (a notification) so it's not actually marked as such. I don't think it's a problem here (I don't expect any changes will be breaking), but want to make sure you realise!

devoncarew:
> Note: This does not include experimental="true" because it didn't seem to work on notifications:
[...]

PR #30372, opened 2017-08-08
Implement analysis server closing labels

DanTup:
@bwilkerson I've coded a basic implementation for #30285 with some tests. I'm creating this PR as somewhere to get feedback/review but it'll likely need some work before it's done.

 I don't know if you'd rather wait until they're done before reviewing or review in small chunks (similarly, I don't know whether you'd rather merge multiple PRs or just keep this until it's complete).

I haven't done a thorough test against real world code (valid nor invalid) yet so this is mostly just "the happy path", but I hope to do some proper testing later this week. I don't know how thorough you'd consider these tests but if any test cases/edge cases jump out, lmk and I'll add them.

We might also want to discuss how big an issue bad handling of errors (like #30370) is.

We'll also need to increase the API version when/after landing this so that clients know that this is available too.

### DIFF
--- a/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
+++ b/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
@@ -59,7 +59,8 @@ class _DartUnitClosingLabelsComputerVisitor
   @override
   visitListLiteral(ListLiteral node) {
     if (_spansManyLines(node)) {
-      final typeName = node.typeArguments?.arguments[0]?.toString();
+      final args = node.typeArguments?.arguments;
+      final typeName = args != null ? args[0]?.toString() : null;
 
       if (typeName != null) {
         _addLabel(node, "List<$typeName>");
--- a/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
+++ b/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
@@ -34,4 +34,50 @@ class _DartUnitClosingLabelsComputerVisitor
   final DartUnitClosingLabelsComputer computer;
 
   _DartUnitClosingLabelsComputerVisitor(this.computer);
+
+  @override
+  Object visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (_spansManyLines(node)) {
+      final name = node.constructorName.type.name;
+      final label = name is PrefixedIdentifier ? name.prefix.name : name.name;
+      _addLabel(node, label);
+    }
+
+    return super.visitInstanceCreationExpression(node);
+  }
+
+  @override
+  Object visitMethodInvocation(MethodInvocation node) {
+    if (node.argumentList != null && _spansManyLines(node.argumentList)) {
+      _addLabel(node, node.methodName.name);
+    }
+
+    return super.visitMethodInvocation(node);
+  }
+
+  @override
+  visitListLiteral(ListLiteral node) {
+    if (_spansManyLines(node)) {
+      final typeName = node.typeArguments?.arguments[0]?.toString();
+
+      if (typeName != null) {
+        _addLabel(node, "List<$typeName>");
+      }
+    }
+
+    return super.visitListLiteral(node);
+  }
+
+  bool _spansManyLines(AstNode node) {
+    final start = computer._lineInfo.getLocation(node.offset);
+    // TODO(dantup) Is this valid? We want the last character of the node, not the start of the next?
+    final end = computer._lineInfo.getLocation(node.end - 1);
+
+    return (start.lineNumber - end.lineNumber).abs() > 1;
+  }
+
+  void _addLabel(AstNode node, String label) {
+    computer._closingLabels
+        .add(new ClosingLabel(node.offset, node.length, label));
+  }
 }
--- a/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
+++ b/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
@@ -49,7 +49,7 @@ class _DartUnitClosingLabelsComputerVisitor
 
   @override
   Object visitMethodInvocation(MethodInvocation node) {
-    if (node.argumentList != null && _spansManyLines(node.argumentList)) {
+    if (node.argumentList != null && _spansManyLines(node)) {
       _addLabel(node, node.methodName.name);
     }
 
--- a/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
+++ b/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
@@ -49,6 +49,8 @@ class _DartUnitClosingLabelsComputerVisitor
   @override
   Object visitMethodInvocation(MethodInvocation node) {
     if (node.argumentList != null && _spansManyLines(node)) {
+      // TODO(dantup) This doesn't include prefix/class name for statics (a.Class.method)
+      // This means it's not consistent with constructors.
       _addLabel(node, node.methodName.name);
     }
 
--- a/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
+++ b/pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
@@ -38,10 +38,9 @@ class _DartUnitClosingLabelsComputerVisitor
   @override
   Object visitInstanceCreationExpression(InstanceCreationExpression node) {
     if (_spansManyLines(node)) {
-      // TODO(dantup) Is it right to read constructorName to get nameless constructors types?
-      final name = node.constructorName.type.name;
-      final label = name is PrefixedIdentifier ? name.prefix.name : name.name;
-      _addLabel(node, label);
+      // TODO(dantup) confirm whether calling toString on constructorName is a reasonable
+      // thing to do here.
+      _addLabel(node, node.constructorName.toString());
     }
 
     return super.visitInstanceCreationExpression(node);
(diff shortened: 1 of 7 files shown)


pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
```diff
@@ -0,0 +1,85 @@
...
+class _DartUnitClosingLabelsComputerVisitor
+    extends RecursiveAstVisitor<Object> {
+  final DartUnitClosingLabelsComputer computer;
+
+  _DartUnitClosingLabelsComputerVisitor(this.computer);
+
+  @override
+  Object visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (_spansManyLines(node)) {
+      // TODO(dantup) Is it right to read constructorName to get nameless constructors types?
```
bwilkerson: It won't necessarily give you the actual name of the class of object that will be returned (in the case of a factory constructor), but it will match what's in the code, so I'd say yes.

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
```diff
@@ -0,0 +1,85 @@
...
+      final name = node.constructorName.type.name;
+      final label = name is PrefixedIdentifier ? name.prefix.name : name.name;
```
bwilkerson: I don't think we want to display just the name of the prefix. We can include the prefix if you want, but for a prefixed identifier it's the `identifier` that matters. In other words, I think this should be either
```
final label = name.name;
```
or
```
final label = name is PrefixedIdentifier ? name.identifier.name : name.name;
```

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
```diff
@@ -0,0 +1,85 @@
...
+      _addLabel(node, label);
+    }
+
+    return super.visitInstanceCreationExpression(node);
+  }
+
+  @override
+  Object visitMethodInvocation(MethodInvocation node) {
+    if (node.argumentList != null && _spansManyLines(node.argumentList)) {
```
bwilkerson: I'm curious as to why this isn't `_spansManyLines(node)`? Seems like it matches the contructor case better that way.

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
```diff
@@ -0,0 +1,85 @@
...
+        _addLabel(node, "List<$typeName>");
+      }
+    }
+
+    return super.visitListLiteral(node);
+  }
+
+  bool _spansManyLines(AstNode node) {
+    final start = computer._lineInfo.getLocation(node.offset);
+    // TODO(dantup) Is this valid? We want the last character of the node, not the start of the next?
```
bwilkerson: I believe so, yes. But note that `end` doesn't return the offset of the next token, only the next character (which could be whitespace, including an end-of-line marker).

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
```diff
@@ -0,0 +1,85 @@
...
+    final end = computer._lineInfo.getLocation(node.end - 1);
+
+    return (start.lineNumber - end.lineNumber).abs() > 1;
```
bwilkerson: If you reverse the order of the subtraction you won't need `.abs()`.

pkg/analysis_server/test/analysis/notification_closingLabels_test.dart
```diff
@@ -0,0 +1,311 @@
...
+import '../analysis_abstract.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(_AnalysisNotificationClosingLabelsTest);
+  });
+}
+
+@reflectiveTest
+class _AnalysisNotificationClosingLabelsTest extends AbstractAnalysisTest {
```
bwilkerson: You should include some tests in which some of the type names are prefixed (which will require, iirc, using prefixed named constructors like `new p.C.c(...)`).

That said, I would break these tests up a bit. Having a couple of tests that create a server, subscribe for and listen for notifications, and compare the result (what you have here) is good. But I'd move most of the tests into a set of unit tests for `DartUnitClosingLabelsComputer` because there's less overhead, so the tests will run faster. Not a problem if you don't agree or don't want to do that.

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
DanTup: Using just `name.name` results in `a.Future.delayed` becoming `a.Future`, but I think either `Future` or `Future.delayed` would make sense here.

Confusingly, a non-prefixed named constructor behaves differently (returning `Class.fromThing`) which is what I'd actually been testing when I added this code.

I'll add tests for prefixed imports and find a way to handle both of these cases in a good way.

DanTup: Ok, I see why I'm confused.. For `a.Future.delayed`, the `constructorName.type` seems to be `a.Future` (sounds reasonable) and `constructorName.name` is `delayed` (also seems reasonable).

However, for `Class.fromThing`, `constructorName.type` seems to include `fromThing` and `constructorName.name` is null. This isn't what I'd expect; I'd expect `constructorName.type` to be `Class` and `constructorName.name` to be `fromThing`.

screen shot 2017-08-09 at 18 09 13

What's the reason for this (or what am I getting confused by)?

Btw - doing `node.constructorName.toString()` seems to give a reasonable output here, but it feels slightly weird to be calling `toString` on the AST node for this. WDYT?

bwilkerson: When the parser sees `new a.b` it can't tell whether `a` is a prefix used to access a class named `b` or whether there is a class named `a` with a constructor named `b`. It guesses one and is sometimes wrong. I had plans, at one point, to fix the AST after resolution (when we'd know how we should have parsed it) but that hasn't happened yet.

Yes, it is kind of weird to call `toString`. We use `toString` strictly as a debugging aid so it should never be called from production code. If we need to we can consider adding a `displayName` getter (or some such) to ConstructorName.

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
DanTup: Good question! I thought it was maybe a hangover from when I only excluded things on the same line, so args pushed onto the next line (but on a single line) would otherwise result in a label; however I can't actually reproduce that now (the arg list seems to start on the same line with the paren). I'll change to `node`.

pkg/analysis_server/lib/src/computer/computer_closingLabels.dart
DanTup: Just after posting I realised it can't tell at this point. I presume fixing it up after resolution won't help here, since I don't think it's being resolved.

Thanks for confirming about toString, I'll come back to this!

DanTup:
Ok, I've pushed some changes to address most of the comments here. Outstanding from the comments above are:

1. Stop using toStrings to get labels (I'll build what I need from the other properties for now and we can discuss whether it's worth making `displayNames` from that logic later)
2. Make `prefix.Class.constructor` and `prefix.Class.staticMethod` consistent (currently the constructor gets the full string and the static method only gets the method name)
3. Convert most of the tests to bypass the service and just test the computer directly

DanTup:
Strange failure on Travis just before the last changeset I pushed; seems to be this:

```text
Chrome 60.0.3112 (Linux 0.0.0)  corelib/iterable_fold_test_02_multi FAILED
	Error: Type '(dynamic) -> dynamic' is not a subtype of type '(int) -> bool'
	    at finish (test/browser/language_tests.js:768:59)
	    at Context.<anonymous> (test/browser/language_tests.js:791:11)
```

The error message doesn't look like something that could be flakey, but it also doesn't sound like one of my tests (these don't get run in the browser, right?). Any ideas? (it's building again from the last changeset, so it's possible it'll go green; though that doesn't explain what happened).

bwilkerson:
The test that's failing is a shared test that cannot be impacted by the work you're doing. It's either flaky or someone committed a change that broke the test and your run got scheduled between that and the commit to fix it. Either way, nothing you need to worry about.

DanTup:
Ah; I thought it was building just my PR branch in isolation (which only I'm committing to), but looks like it's merging first. Makes sense!

DanTup:
@bwilkerson I think I've addressed all of the existing comments now, including moving most tests to unit tests.

I still need to do some more thorough testing with some large codebases - I'll add tests for any errors I find and also make a list of any places where invalid code returns unexpected results. LMK if there's anything else you want me to do or any further changes. Thanks!

bwilkerson (approved):

bwilkerson:
Should I merge now (my preference), or were you planning on writing more tests first?

DanTup:
@bwilkerson I might add more tests if anything comes out of the testing, but if you'd prefer to merge it now (and are happy it's in a reasonable state) I can send another PR for any further changes.

We'll need to update the server version too at some point; though maybe better after I've done a little more testing.

DanTup:
@bwilkerson 

> I'm curious as to why this isn't _spansManyLines(node)? Seems like it matches the contructor case better that way.

I found the reason I'd done this... Given this code:

```dart
return _closingLabelsByEndLine.values
    .where((l) => l.any((cl) => cl.spannedLines >= 2))
    .expand((cls) => cls)
    .map((clwlc) => clwlc.label)
    .toList();
```

The ast node for the `map` method invocation has an offset that starts at `_closingLabels...` (incorrectly resulting in a label) whereas the argument list starts where the args are.

So, I'm going to put this back for now - I'll send a PR soon with this and some other fixes; and if you think it's the wrong fix we can discuss there (since this one is merged).

I'll also test similar for constructors to see if it has the same issue if prefix.Class.ctor gets split.

PR #30543, opened 2017-08-24
remove reference to dart:html

sethladd:
This confused a Flutter user. The other classes don't reference dart:html, and these docs appear in docs.flutter.io, so we'd like them to stay focused on dart:io.


sethladd:
cc @zanderso

zanderso (approved):

zanderso:
lgtm

PR #30583, opened 2017-08-30
Added an analysis server argument to set the location of the session …

pylaligand:
…cache.

### DIFF
--- a/pkg/analysis_server/lib/src/socket_server.dart
+++ b/pkg/analysis_server/lib/src/socket_server.dart
@@ -74,10 +74,12 @@ class SocketServer {
 
     PhysicalResourceProvider resourceProvider;
     if (analysisServerOptions.fileReadMode == 'as-is') {
-      resourceProvider = PhysicalResourceProvider.INSTANCE;
+      resourceProvider = new PhysicalResourceProvider(null,
+          stateLocation: analysisServerOptions.cacheFolder);
     } else if (analysisServerOptions.fileReadMode == 'normalize-eol-always') {
       resourceProvider = new PhysicalResourceProvider(
-          PhysicalResourceProvider.NORMALIZE_EOL_ALWAYS);
+          PhysicalResourceProvider.NORMALIZE_EOL_ALWAYS,
+          stateLocation: analysisServerOptions.cacheFolder);
     } else {
       throw new Exception(
           'File read mode was set to the unknown mode: $analysisServerOptions.fileReadMode');
--- a/pkg/analyzer/lib/file_system/physical_file_system.dart
+++ b/pkg/analyzer/lib/file_system/physical_file_system.dart
@@ -40,6 +40,24 @@ List<int> _pathsToTimes(List<String> paths) {
 }
 
 /**
+* The name of the directory containing plugin specific subfolders used to
+* store data across sessions.
+*/
+const String _SERVER_DIR = ".dartServer";
+
+/**
+ * Returns the path to the user's home directory.
+ */
+String _getStandardStateLocation() {
+  final home = io.Platform.isWindows
+      ? io.Platform.environment['LOCALAPPDATA']
+      : io.Platform.environment['HOME'];
+  return io.FileSystemEntity.isDirectorySync(home)
+      ? join(home, _SERVER_DIR)
+      : null;
+}
+
+/**
  * A `dart:io` based implementation of [ResourceProvider].
  */
 class PhysicalResourceProvider implements ResourceProvider {
@@ -50,10 +68,9 @@ class PhysicalResourceProvider implements ResourceProvider {
       new PhysicalResourceProvider(null);
 
   /**
-   * The name of the directory containing plugin specific subfolders used to
-   * store data across sessions.
+   * The path to the base folder where state is stored.
    */
-  static final String SERVER_DIR = ".dartServer";
+  final String _stateLocation;
 
   static _SingleIsolateRunnerProvider pathsToTimesIsolateProvider =
       new _SingleIsolateRunnerProvider();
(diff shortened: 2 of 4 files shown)


pylaligand:
I looked into adding tests but no existing tests seem to be exercising that part of the analysis server (physical storage). Let me know if you want me to add some.

bwilkerson (approved):

bwilkerson:
We generally avoid touching disk when possible. The one exception is in the integration tests.

pkg/analyzer/lib/file_system/physical_file_system.dart
```diff
@@ -62,7 +77,14 @@ class PhysicalResourceProvider implements ResourceProvider {
   final AbsolutePathContext absolutePathContext =
       new AbsolutePathContext(io.Platform.isWindows);
 
-  PhysicalResourceProvider(FileReadMode fileReadMode) {
+  factory PhysicalResourceProvider(FileReadMode fileReadMode) {
+    return new PhysicalResourceProvider.at(fileReadMode: fileReadMode);
+  }
+
+  PhysicalResourceProvider.at({String folder, FileReadMode fileReadMode})
```
scheglov: ResourceProvider is used not only for reading state / cache, but also for any file.
So, specializing it so much for state is not right.
Please rollback changes to this file.

pkg/analysis_server/lib/src/socket_server.dart
```diff
@@ -74,10 +74,12 @@ class SocketServer {
 
     PhysicalResourceProvider resourceProvider;
     if (analysisServerOptions.fileReadMode == 'as-is') {
-      resourceProvider = PhysicalResourceProvider.INSTANCE;
+      resourceProvider = new PhysicalResourceProvider.at(
```
scheglov: Please rollback changes to this file.

scheglov (changes requested):
Instead you should change `String _getByteStorePath()` itself - use the provided cache path to get the folder instead of calling `getStateLocation`.

pkg/analyzer/lib/file_system/physical_file_system.dart
pylaligand: Could you clarify what's not right? PhysicalResourceProvider provides a way to store state, and this change adds a way for clients to get a tiny bit more control over how it's done rather than rely on platform concepts of a "home".
I don't mind making the change, but I'm worried that this would be brittle if state bleeds into a different location.

scheglov: `ResourceProvider` provides a way to access files and folders.

There is also a small addition to access "state location". Making state location a center of `ResourceProvider` by making a special constructor seems like overkill. The location path could be a named parameter of the existing `PhysicalResourceProvider` constructor.

pylaligand: Updated to only have a single constructor.

pkg/analyzer/lib/file_system/physical_file_system.dart
```diff
@@ -62,7 +77,8 @@ class PhysicalResourceProvider implements ResourceProvider {
   final AbsolutePathContext absolutePathContext =
       new AbsolutePathContext(io.Platform.isWindows);
 
-  PhysicalResourceProvider(FileReadMode fileReadMode) {
+  PhysicalResourceProvider({String folder, FileReadMode fileReadMode})
```
scheglov: In general we cannot make `fileReadMode` a named parameter, because this is a breaking change. I'd like to remove it eventually, but let's not mix these two changes.

Also, the name `folder` is too generic. Let's name it `stateLocation` and use instead of `HOME + SERVER_DIR`.

pkg/analyzer/lib/file_system/physical_file_system.dart
```diff
@@ -101,15 +117,9 @@ class PhysicalResourceProvider implements ResourceProvider {
 
   @override
   Folder getStateLocation(String pluginId) {
-    String home;
-    if (io.Platform.isWindows) {
-      home = io.Platform.environment['LOCALAPPDATA'];
-    } else {
-      home = io.Platform.environment['HOME'];
-    }
-    if (home != null && io.FileSystemEntity.isDirectorySync(home)) {
```
scheglov: You have just lost checking for home folder existence.
I don't not whether it is important, but lets keep it for now.

bwilkerson (approved):

scheglov (changes requested):

pkg/analyzer/lib/file_system/physical_file_system.dart
pylaligand: That check is in `_getHomePath`.

pkg/analyzer/lib/file_system/physical_file_system.dart
pylaligand: Done.

pkg/analyzer/lib/file_system/physical_file_system.dart
```diff
@@ -40,6 +40,24 @@ List<int> _pathsToTimes(List<String> paths) {
 }
 
 /**
+* The name of the directory containing plugin specific subfolders used to
+* store data across sessions.
+*/
+final String _SERVER_DIR = ".dartServer";
```
scheglov: This could be a const.

pkg/analyzer/lib/file_system/physical_file_system.dart
```diff
@@ -62,7 +79,10 @@ class PhysicalResourceProvider implements ResourceProvider {
   final AbsolutePathContext absolutePathContext =
       new AbsolutePathContext(io.Platform.isWindows);
 
-  PhysicalResourceProvider(FileReadMode fileReadMode) {
+  PhysicalResourceProvider(FileReadMode fileReadMode, {String stateLocation})
+      : _stateLocation = stateLocation != null
+            ? stateLocation
+            : _getStandardStateLocation() {
```
scheglov: _stateLocation = stateLocation ?? _getStandardStateLocation()

scheglov (approved):

pkg/analyzer/lib/file_system/physical_file_system.dart
pylaligand: Done.

pkg/analyzer/lib/file_system/physical_file_system.dart
pylaligand: Done.

PR #30626, opened 2017-09-05
Fix documentation of Uint32List and Uint64List

renggli:
The currently documented max value is 2^n, but since 0 should be counted too the actual max value is only 2^n - 1.


lrhn:
LGTM

kevmoo:
@lrhn – can you shepard through Gerrit?

PR #30757, opened 2017-09-15
Load exception

grouma:
Dispatch load exception so that others can react without capturing the corresponding error. Will be used in conjunction with https://github.com/dart-lang/test/pull/682


jakemac53 (approved):

jakemac53:
I think you have to send this through gerrit now? not sure

kevmoo:
Talk to @jmesserly and @vsmenon ...

vsmenon:
Looks fine, but yeah, you'll need to use the new gerrit workflow.  Github is now read-only for the sdk.

Note - this file (dart_library.js) is only used internally.  You'll need to do something different for external tests.

kevmoo:
Also – tests?

PR #30866, opened 2017-09-22
Update 20170922.md

a14n:


floitschG:
Thanks. Randal already chatted me on Gitter and I just committed a CL.
(was traveling and couldn't commit before today...)

Issue #27372, opened 2016-09-16
Experimental @sealed annotation


matanlurey:
(I actually wish this was a language feature, but we can start somewhere...)

All over even the Dart SDK codebase I see code like:

``` dart
/// ...
///
/// May not be subclassed or implemented.
class FooAst {}
```

There is nothing _actually_ preventing users from doing this.

I've started more aggressively using factory constructors to prevent subclassing:

``` dart
class DoNotSub {
  factory DoNotSub() => new DoNotSub._sealed();

  DoNotSub._sealed();
}
```

But this still doesn't prevent `implements`, which, IMO is a serious problem for:

**Data models/structures that are not meant to be mocked:**

``` dart
class Point {
  final int x;
  final int y;

  Point(this.x, this.y)
}

// Should never be necessary... Just use 'new Point'
class MockPoint extends Mock implements Point {}
```

**Classes that implement `hashCode` and `equals` comparing private state:**

``` dart
class SnickersBar {
  int _howManyPeanuts;
  bool _includesCaramel;

  SnickersBar();

  @override
  bool operator==(Object o) => 
      o is SnickersBar && 
      o._howManyPeanuts == _howManyPeanuts &&
      o._includesCaramel == _includesCaramel;
}
```

cc @yjbanov

zoechi:
https://github.com/dart-lang/sdk/issues/25462#issuecomment-171814123 seems related

lrhn:
Preventing anyone from implementing your class' interface isn't solving any problems. It might be preventing problems if the class isn't expecting other implementations, and it assumes that all instances of the class that it sees will extend itself. Maybe because the class has private members that it accesses on instances other than `this` (#25462). That means that you can't pretend to be that class to the class itself (or its library), but why prevent anyone from pretending to be that class to other libraries, e.g., for testing?

Preventing anybody from extending or implementing `Point` will prevent the `ColorPoint` problem, but it will also prevent anybody from making a `PolarPoint` that stores its data as angle and magnitude instead of Cartesian coordinates. That would be a perfectly good `Point` implementation that is backwards compatible and even equality will work correctly. 
The more generic a class's concept is, the more reasonable it is to use it as an interface. I probably don't want to use `DateTimePickerMenuEntryWidget` as an interface, but `Point` is perfect for it.
I don't see "preventing mocking" as a goal to strive for - if someone else wants to do it, why should you concern yourself about that.

[...]

PR #30871, opened 2017-09-22
Add noImplement annotation

georgelesica-wf:
We use this with an internal tool to guard against accidental API mis-use which can occur due to the fact that Dart permits any class to be used as an interface. If a class that is consumed as an interface then adding an additional method to that class causes a breaking change but may look like a minor version increment from the standpoint of the package author.

We thought it might be nice to have in a more central location (right now it lives with the tool, which is awkward) and figured we'd try here first.

### DIFF
--- a/pkg/meta/lib/meta.dart
+++ b/pkg/meta/lib/meta.dart
@@ -130,6 +130,22 @@ const _Literal literal = const _Literal();
 ///   without invoking the overridden method.
 const _MustCallSuper mustCallSuper = const _MustCallSuper();
 
+/// Used to annotate a class that may not be no consumed as an interface.
+///
+/// This is useful because changes that would result in minor version
+/// increments require major version increments when applied to interfaces
+/// so packages may want to prevent consumers from implementing certain
+/// classes in order to protect semantic version guarantees.
+///
+/// Tools, such as the analyzer, can provide feedback if
+///
+/// * the annotation is associated with anything other than a class declaration
+/// * a class that has this annotation is used, either directly or
+///   transitively, as an interface (using `implements`)
+///
+/// At this time the analyzer does not enforce this annotation.
+const _NoImplement noImplement = const _NoImplement();
+
 /// Used to annotate a class declaration `C`. Indicates that any type arguments
 /// declared on `C` are to be treated as optional.  Tools such as the analyzer
 /// and linter can use this information to suppress warnings that would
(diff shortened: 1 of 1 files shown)


georgelesica-wf:
My employer has signed a CLA.

pkg/meta/lib/meta.dart
```diff
@@ -130,6 +130,22 @@ const _Literal literal = const _Literal();
...
+/// so packages may want to prevent consumers from implementing certain
+/// classes in order to protect semantic version guarantees.
+///
+/// Tools, such as the analyzer, can provide feedback if
+///
+/// * the annotation is associated with anything other than a class declaration
+/// * a class that has this annotation is used, either directly or
+///   transitively, as an interface (using `implements`)
+///
+/// At this time the analyzer does not enforce this annotation.
```
bencampbell-wf: Should this symbol be added in coordination with native dart analyzer support?

bwilkerson:
The question of whether to support such an annotation has been discussed several times in several forums (see https://github.com/dart-lang/sdk/issues/27372 for one example).

While I personally like the idea of being able to constrain usage in order to maximize my ability to change code without creating breaking changes, it is a hotly contested issue.

I think that we need to re-evaluate the question in light of the current direction of Dart as seen through strong-mode, and would like to not commit this kind of change until we've had a chance to think through the implications more completely. I will start a discussion with the language team to see what their current thinking is.

pkg/meta/lib/meta.dart
```diff
@@ -130,6 +130,22 @@ const _Literal literal = const _Literal();
 ///   without invoking the overridden method.
 const _MustCallSuper mustCallSuper = const _MustCallSuper();
 
+/// Used to annotate a class that may not be no consumed as an interface.
```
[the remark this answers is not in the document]
georgelesica-wf: English is hard... ;-)

pkg/meta/lib/meta.dart
georgelesica-wf: There is a history of adding annotations before they are enforced by the analyzer, so the answer is "maybe", I think.

bwilkerson: It's fine to add an annotation before there is analyzer support, but we do not include the current status of the analyzer's implementation in the documentation for the annotation. This line should be removed.

PR #31103, opened 2017-10-14
Fix typo in exponent representation example

bobek:
> `0.000000998` is printed as `0.98e-7`

I believe, that there is a typo and `0` should read as `9` as `0.000000998` would be typically represented as `9.98e-7`.

/cc @bkonyi

### DIFF
--- a/docs/newsletter/20171013.md
+++ b/docs/newsletter/20171013.md
@@ -34,7 +34,7 @@ In this section I will focus on Dart's `toString` methods of `double`. There are
 - `toStringAsFixed`
 - `toStringAsPrecision`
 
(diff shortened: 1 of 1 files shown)


bobek:
I signed it! :)

floitschG:
Thanks.
I committed your fix, but wasn't able to keep the "Author". Will work with our infrastructure experts to get this fixed next week.

kevmoo:
@floitschG – want to post the sha and close this then?

athomas:
Fixed in bca7b35cd695efa4fa35c6a8e518c94d4acae990.

PR #31247, opened 2017-10-29
Fix dart:collection library link in docs

rafaelalvessa:
Addresses #31246 by fixing the link to the `dart:library` in the docs of the `dart:core` library and `Iterable` class.


kevmoo:
Great idea – we just removed the links entirely - https://github.com/dart-lang/sdk/commit/751de54dc0dc4359725199c64c893162d6ee9140

PR #31614, opened 2017-12-12
Change async stubbing to use thenAnswer

alanrussian:
Mockito now prohibits calling thenReturn with Futures and Streams. dart-lang/build#4833


devoncarew:
@mk13 @bwilkerson @scheglov 

Is pkg/analysis_server_client used? I see the initial commit, but am not sure if anything in the repo was updated to use it. We may consider just removing this package.

bwilkerson:
I believe that Max is using it, but he can tell you for sure. In either case, this seems like a reasonable change.

alanrussian:
I spoke with @keertip and was recommended to ask @mk13 to submit this change. Max, are you able to do this? If so, would you like for me to close this pull request or would you like to modify this one?

mit-mit:
Closing; this appears to be stale

PR #31871, opened 2018-01-12
Specify type for fromIterable

tomyeh:
Better type inference


lrhn:
While I like the idea, and really wish we could do it, that is actually not valid Dart 2 syntax. Methods and functions can be generic, but constructors cannot.
It's one of the features we hope to add before too long, and the `fromIterable` constructor is one of the reasons.

PR #31993, opened 2018-01-29
add quickfix for prefer_const_declarations

a14n:


bwilkerson:
See comments on https://dart-review.googlesource.com/c/sdk/+/37301.

PR #32034, opened 2018-02-02
Increase analysis server version to 1.18.5.

DanTup:
The version hasn't been increased since the server started accepting overlays for HTML files (for angular plugin) so this is to allow editors to know for sure that the server supports them.


bwilkerson:
Being reviewed and committed at https://dart-review.googlesource.com/c/sdk/+/38260

Issue #29451, opened 2017-04-24
Invalid encoding in filename crashes the VM


tvolkert:
### Steps to reproduce:

1. In Linux...
2. Create an empty directory
3. In the directory, create a file named `¶.txt`
4. Install the `convmv` tool: `sudo apt-get install convmv`
5. Convert the filename to Windows encoding: `convmv --notest -f UTF-8 -t ISO-8859-1 ¶.txt`
6. Create a `test.dart` file in the same directory, like so: https://gist.github.com/tvolkert/36ad0283101269d8bdaf76c7311b1370
7. run `dart test.dart`

### Result

The VM crashes with the following message:

```
$ dart test.dart 
Dart_NewStringFromUTF8 expects argument 'str' to be valid UTF-8.
```

### Notes

This may happen on other platforms as well, but the existence of the `convmv` tool makes it easiest to reproduce on Linux.

a-siva:
Are you expecting us to support such filenames or the bug is to ensure the VM produces a sane Dart exception?

tvolkert:
It should produce a sane Dart exception.  I'm operating under the assumption that a VM crash is never the right behavior.

Hixie:
It shouldn't even produce an exception. There's nothing that says file systems should use UTF-8 as far as I can tell. The right behaviour is for the file names to be returned as byte sequences.

aam:
Can you really return just a byte stream without assuming some encoding?
For example, if we assume that you can't have '/' in a filename, does it mean that you parse the filename as a ASCII stream, UTF-8 stream or use some other encoding?

Hixie:
> Can you really return just a byte stream without assuming some encoding?

Sure, that's what Uint8List is.

> For example, if we assume that you can't have '/' in a filename, does it mean that you parse the filename as a ASCII stream, UTF-8 stream or use some other encoding?

For display purposes you can probably assume UTF-8, but the actual file name (which is what this API is dealing with) is just a stream of bytes, as far as I can tell.  See for example the discussion at https://www.j3e.de/linux/convmv/man/

Pacane:
I've hit this too with flutter this week. It was very hard to diagnose.

Hixie:
We are getting an increasing number of reports of this as Flutter extends past US-ASCII environments.

bkonyi:
[...]

PR #32537, opened 2018-03-14
If a directory entry is invalid utf8, consider it a kListError.

powdercloud:
Not to be merged as is - it's just an idea.

Following the "steps to reproduce" by @tvolkert in https://github.com/dart-lang/sdk/issues/29451 before this pr produces this error:
Dart_NewStringFromUTF8 expects argument 'str' to be valid UTF-8.

With this PR:
Unhandled exception:
FileSystemException: Directory listing failed, path = '/tmp/dartcrash/' (OS Error: Success, errno = 0)
#0      _Directory._fillWithDirectoryListing (dart:io-patch/directory_patch.dart:34)
#1      _Directory.listSync (dart:io/directory_impl.dart:214)
#2      main (file:///tmp/dartcrash/test.dart:5:21)
#3      _startIsolate.<anonymous closure> (dart:isolate-patch/dart:isolate/isolate_patch.dart:279)
#4      _RawReceivePortImpl._handleMessage (dart:isolate-patch/dart:isolate/isolate_patch.dart:165)

This is a sketch only (tried this on my Linux Mint machine at home), but if the approach is ok then it could be generalized enough so the error is easier to understand. E.g. "Directory listing failed due to invalid UTF8 in directory listing of '/tmp/dartcrash'" or similar.

### DIFF
--- a/runtime/bin/directory_linux.cc
+++ b/runtime/bin/directory_linux.cc
@@ -23,6 +23,7 @@
 #include "bin/namespace.h"
 #include "bin/platform.h"
 #include "platform/signal_blocker.h"
+#include "vm/unicode.h"
 
 namespace dart {
 namespace bin {
@@ -120,6 +121,11 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
   // ports.
   errno = 0;
   dirent* entry = readdir(reinterpret_cast<DIR*>(lister_));
+  if (entry != NULL) {
+    const uint8_t* d_name = reinterpret_cast<uint8_t *>(entry->d_name);
+    if (!Utf8::IsValid(d_name, strlen(entry->d_name)))
+      return kListError;
+  }
   if (entry != NULL) {
     if (!listing->path_buffer().Add(entry->d_name)) {
       done_ = true;


runtime/bin/directory_linux.cc
```diff
@@ -120,6 +121,11 @@ ListType DirectoryListingEntry::Next(DirectoryListing* listing) {
   // ports.
   errno = 0;
   dirent* entry = readdir(reinterpret_cast<DIR*>(lister_));
+  if (entry != NULL) {
+    const uint8_t* d_name = reinterpret_cast<uint8_t *>(entry->d_name);
+    if (!Utf8::IsValid(d_name, strlen(entry->d_name)))
+      return kListError;
```
zanderso: I wonder if we should instead just skip entries that aren't utf8 formatted. So instead of "return kListError", "return Next(listing)", and update the docs to indicate the behavior.

powdercloud: I think this is reasonable, yes. These were the choices I've thought about a little bit:
1) Make it so that file / directory / ... names can be arbitrary sequences of bytes. Advantage is that it's fully general, e.g. one could then implement a filesystem explorer in Dart that allows the user to delete even a file with incorrect utf8 chars. But I suspect this is a very difficult change to get done. :-)
2) Coerce the names into valid utf8. This is relatively easy and has the advantage that in the hypothetical filesystem explorer, the user could see that there is a file or directory with a weird name. But some operations may not work, e.g., deleting may be hard to implement because the Dart program isn't able to tell the OS precisely which file to delete, and some other operations may be a bit quirky too.
3) Skip over files that aren't valid utf8. This means the user won't even see what's there - this can be puzzling e.g. they may look at a volume like a usb stick in their filesystem explorer and wonder why there's no space available even though there isn't a file to see.
4) Make the error easier to understand. The filesystem explorer may still crash when pointed at a directory, but if someone looks at the stack trace they may be able to understand why. And perhaps sometimes it's possible for the hypothetical filesystem excplorer to catch the FileSystemException and handle it somehow.
I don't think I have enough context to pick. 4) seemed like the least intrusive and leave the door open for any of the others later. :-) I'd be happy to attempt 2) or 3) if you like.

Hixie:
It's not an error for file names to be non-UTF8. The bug is that the Dart libraries assume that file names can be represented using Strings, which is incorrect.

powdercloud:
@Hixie @zanderso @tvolkert I made some proof of concept for supporting uninterpreted directory names: https://github.com/dart-lang/sdk/pull/32583.
I think getting this done in a mostly backward compatible way is pretty invasive because it will require adding additional methods to the APIs, pretty much one for each that currently takes a string. And it may not be convenient to manipulate Uint8Lists for directory / path operations, and also, Uint8List allows mutation whereas String does not... but you all probably know this :-)

Hixie:
@powdercloud I think to properly support this we need a new kind of literal that is backed by an immutable Uint8List rather than String, which interprets ASCII characters is bytes corresponding to those characters.

mit-mit:
Closing as this appears stale

PR #32617, opened 2018-03-20
core: Update Map methods that use Object instead of K

stevenroose:
Currently some of the method signatures use `Object` instead of the `K`
type.  Change them to accept an object of type `K` instead.

### DIFF
--- a/sdk/lib/core/map.dart
+++ b/sdk/lib/core/map.dart
@@ -213,7 +213,7 @@ abstract class Map<K, V> {
    * Returns true if any of the values in the map are equal to `value`
    * according to the `==` operator.
    */
-  bool containsValue(Object value);
+  bool containsValue(V value);
 
   /**
    * Returns true if this map contains the given [key].
@@ -221,7 +221,7 @@ abstract class Map<K, V> {
    * Returns true if any of the keys in the map are equal to `key`
    * according to the equality used by the map.
    */
-  bool containsKey(Object key);
+  bool containsKey(K key);
 
   /**
    * Returns the value for the given [key] or null if [key] is not in the map.
@@ -233,7 +233,7 @@ abstract class Map<K, V> {
    * Methods like [containsKey] or [putIfAbsent] can be use if the distinction
    * is important.
    */
-  V operator [](Object key);
+  V operator [](K key);
 
   /**
    * Associates the [key] with the given [value].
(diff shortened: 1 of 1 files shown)


mraleph:
I think `Object` type on those methods is by design

@lrhn can probably provide elaboration.

eernstg:
In support of @mraleph's remark: With the argument type `K`, it could be a dynamic error to test whether `3.14` is contained in a set (sets, maps, ... are similar in this respect) whose static type is `Set<num>`, because it's dynamically a `Set<int>`. It seems useful to get the result `false` rather than a dynamic error in that situation.

Note that there are connections back to the Java APIs (such as `Collection<E>.contains(Object)`), which makes this a long-standing, multi-language design choice, rather than a newfangled accident in Dart. ;-)

Also note that there's a deeper connection to variance here: Covariance is safe for read-only usage (which is a matter of style in Dart, but a property of the static typing of wildcards in Java), and `contains` and its ilk are a natural part of read-only usage. In Dart we would get additional (arguably gratuitous) dynamic errors with an argument type of `K`, but in Java it would simply make it impossible to call `contains` on a data structure whose type is otherwise optimized for read-only usage (and a cast to a non-wildcarded type wouldn't make it safe, because there is no run-time reification of the type arguments in Java).

So there's definitely a case to be made for keeping these types at `Object`.

stevenroose:
Hmm. That makes sense. For the `[]` and `[]=` operators, I have more doubts, though. I would be really glad to get a warning when putting a type that is not the key type in the square brackets.

But well, if it's done on purpose, I will close the issue.

eernstg:
Fun fact: `void operator []=(K key, V value)` is a mutator and uses stricter types, but `V operator [](Object key)` is read-only and follows the pattern you pointed out. Oh, but thanks for the heads up, that's always valuable for us!

lrhn:
What we would really want here is a `contains` method with a *static* type of `E` and a dynamic type of `Object`.
We can achieve that that by typing the *interface* as `bool contains(E object)` and the *implementation classes* as `bool contains(Object object)`. 
That's a slippery slope to get on, though. There is no help whatsoever in writing the implementation classes, no hint that you shouldn't just implement the actual type. The interface in question is `Iterable`, so it affects every subclass of that, and most of them can safely check `contains` on every object. 
On the other hand, not *all* classes can handle all objects. A `SplayTreeSet` can only check for objects that its comparator can work on, and it would be better served by a parameter type of `E` rather than needing a `bool validKey(Object)` function on the side.

Too late to change now, but if we move `SplayTreeSet` to `package:collection`, we might later want to do a breaking update to make it `bool contains(convariant E object)`.

PR #32714, opened 2018-03-29
Clean up Map#operator[] documentation

burkemw3:


sdk/lib/core/map.dart
```diff
@@ -226,12 +226,10 @@ abstract class Map<K, V> {
   /**
    * Returns the value for the given [key] or null if [key] is not in the map.
    *
-   * Some maps allows keys to have `null` as a value,
-   * For those maps, a lookup using this operator does cannot be used to
-   * distinguish between a key not being in the map, and the key having a null
-   * value.
-   * Methods like [containsKey] or [putIfAbsent] can be use if the distinction
-   * is important.
+   * Some maps allows keys to have `null` as a value. For those maps, a lookup
```
lrhn: Also: "Some maps allow"
I'd prefer to retain the newline between "value" and "For". The separation into individual sentences, and sentences into stand-alone phrases when possible, per line, makes it easier to edit the documentation in the future without having to reflow (and when converted to HTML, extra whitespace doesn't matter anyway).

sdk/lib/core/map.dart
```diff
@@ -226,12 +226,10 @@ abstract class Map<K, V> {
...
+   * using this operator cannot be used to distinguish between a key not being
```
lrhn: "cannot be used to distinguish" -> "cannot distinguish"
(now that you are here).

sdk/lib/core/map.dart
```diff
@@ -226,12 +226,10 @@ abstract class Map<K, V> {
...
+   * in the map and the key having a null value. Methods like [containsKey] or
```
lrhn: null -> `null`

lrhn: Than you for this clean-up. That was some very badly edited doc we had there! :)

burkemw3:
PTAL.

I even found another one, s/use/used/ in the last sentence.

Github advises against force pushing on PRs [0](https://help.github.com/articles/about-pull-requests/). If you want the squashed commit version, [create a new pull request from burkemw3:sdk/patch-3](https://github.com/dart-lang/sdk/compare/master...burkemw3:patch-3?expand=1) and close this one.

lrhn:
New CL created and landing.

Issue #29414, opened 2017-04-21
Analzyer doesn't correctly resolve package paths when inside a folder named `packages`


DanTup:
This was raised by @abarth at https://github.com/Dart-Code/Dart-Code/issues/275 but based on comments from @bwilkerson it sounds like the analysis server should handle this for us. I've stripped the flutter repo right down to make a really simple repro (attached).

The folder structure looks like this:

```text
flutter-set-me-as-analysis-root/
   |- packages/
      |- flutter/
         |- .packages
         |- pubspec.lock
         |- pubspec.yaml
         |- lib/
            |- foundation.dart
            |- src/
               |- widgets/
                  |- drag_target.dart
```

The `drag_target.dart` file just contains `import 'package:flutter/foundation.dart';` and the .packages folder contains `flutter:lib/`. It seems like the pubspec.yaml should cause the `flutter` folder to be treated as a root and the package resolved correctly, however it doesn't seem to (it seems to use the `flutter-set-me-as-analysis-root` folder as a root for resolving packages.

[repro.zip](https://github.com/dart-lang/sdk/files/946501/repro.zip)

DanTup:
Whoops, I forgot the logs. Here's the Dart Code log (analyzer communication) as well as the instrumentation file.

[analyzer.instrum.txt](https://github.com/dart-lang/sdk/files/946535/analyzer.instrum.txt)
[analyzer.txt](https://github.com/dart-lang/sdk/files/946534/analyzer.txt)

bwilkerson:
My first guess is that server might be getting confused because this is inside a directory named 'packages', which used to have special meaning in the Dart ecosystem (being the thing that `.packages` files replaced).

DanTup:
@bwilkerson Looks like you guess right.. Renaming it to `packages2` results in the error going away and an `unused import` warning (in the repro I deleted all the code).

Is this something you think should change in the analyzer or should naming folders `packages` continue to be avoided?

bwilkerson:
It needs to change in analyzer. I'm just not sure how soon that's likely to happen, though there's a chance that we will get to it this quarter as part of a larger arc of work.

DanTup:
Thanks for the info; I'll put some notes about this on the original DC issue. Thanks!

abarth:
@bwilkerson What's involved in removing the special case for directories named "packages"?  My understanding is that the Dart ecosystem has fully converted to `.packages` files at this point.

alexander-doroshko:
[...]

PR #32902, opened 2018-04-16
Increase analysis server version to 1.20.1

DanTup:
So clients can detect when the workaround for #29414 is not required.

### DIFF
--- a/pkg/analysis_server/doc/api.html
+++ b/pkg/analysis_server/doc/api.html
@@ -109,7 +109,7 @@ a:focus, a:hover {
 <body>
 <h1>Analysis Server API Specification</h1>
 <h1 style="color:#999999">Version
-  1.20.0
+  1.20.1
 </h1>
 <p>
   This document contains a specification of the API provided by the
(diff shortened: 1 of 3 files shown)


devoncarew:
It occurs to me that we should have a changelog section in the spec, to be able to call out to readers when a specific feature was added -

bwilkerson (approved):

bwilkerson:
I haven't seen the Gerrit version of this PR yet. It ought to be automatically created, but if not you might try using `git cl upload`.

> ... we should have a changelog section in the spec ...

That's probably true :-), but I don't think that we need to hold up this CL because of it.

DanTup:
> It occurs to me that we should have a changelog section in the spec

A changelog would be great, since there are often behaviour changes clients might want to be aware of (for ex. if you started coding today against the current SDK, you might not know about things like this, so anyone using an older SDK than you might have weird bugs). That said, I think keeping a reliable list of behaviour changes might be a lot of effort!

> I haven't seen the Gerrit version of this PR yet

https://dart-review.googlesource.com/c/sdk/+/51400

I did try to `git cl upload` it, but discovered I don't have any of that set up on my personal MacBook so figured this was easiest (for me 😉) for now.

DanTup:
I can't merge here so I've submitted in Gerrit. Maybe it's worth creating a changelog file somewhere as a start - if it's there, at least there's somewhere to add notes about changes like this. Even if not complete, there's probably value?

PR #33142, opened 2018-05-16
Support building Dart VM for Chromecast

lyceel:
These changes allow the VM to be built using the Chromecast toolchain.


lyceel:
Trying to fix bad author on commit. Will submit a different PR.

PR #33146, opened 2018-05-17
Changes to support Chromecast build

lyceel:
Add a clang toolchain for ARM, with the ability to specify building
against an alternative C++ library (e.g.: libc++).

### DIFF
--- a/build/config/BUILDCONFIG.gn
+++ b/build/config/BUILDCONFIG.gn
@@ -130,6 +130,9 @@ declare_args() {
 
   # Compile for Thread Sanitizer to find threading bugs.
   is_tsan = false
+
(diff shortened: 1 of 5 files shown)


zanderso:
Unfortunately, I think supporting what you're trying to do in a nice way requires some deeper structural changes to the way we support building with different toolchains. I think it might be helpful to have a doc describing your requirements in terms of toolchains, sysroots, and compiler and linker flags.

lyceel:
No problem. I thought that might be the case. I'd be happy to chat about it, if you like.

mit-mit:
Thanks for this PR @lyceel. I'm going to close it, as it doesn't appear actionable at this point.

PR #33318, opened 2018-06-02
update_homebrew:dart: also make a dart@2, for Dart 2 migration

apjanke:
This PR modifies update_homebrew to also generate a dart@2 versioned formula. The new file is dropped out right next to the old `dart.rb` file.

The dart@2 formula is a temporary measure to aid application developers
migrating from Dart 1 to Dart 2.

Supersedes https://github.com/dart-lang/homebrew-dart/pull/52.

This looks right by inspection, but I haven't been able to test it yet: my local `dart` environment is all messed up, and I am a Dart newbie. Still, I'm putting this out here as a PR so the discussion from https://github.com/dart-lang/homebrew-dart/pull/52 can proceed.

### DIFF
--- a/tools/apps/update_homebrew/lib/update_homebrew.dart
+++ b/tools/apps/update_homebrew/lib/update_homebrew.dart
@@ -23,6 +23,7 @@ const dartiumFile = 'dartium/dartium-macos-x64-release.zip';
 const contentShellFile = 'dartium/content_shell-macos-x64-release.zip';
 
 const dartRbFileName = 'dart.rb';
+const dart2RbFileName = 'dart@2.rb';
 
 Future<String> getHash256(
     String channel, String revision, String download) async {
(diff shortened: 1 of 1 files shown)


nex3:
@whesse Please prioritize this; it's blocking Dart Sass moving to Dart 2, and thus the use of Dart Sass within Google.

kevmoo:
"real" review - https://dart-review.googlesource.com/c/sdk/+/58120

nex3:
Is this actually fixed? 2.0.0-dev.61 was released after 61e57e2012ef28471e7196cd44e2bece4ade0488 landed, but https://github.com/dart-lang/homebrew-dart/commit/845e8e5e0ba744e37a097c7b828e086a76612943 didn't create a `dart@2.rb` file.

whesse:
There was a bug in the script, which is now fixed.

nex3:
Thanks! Please let me know when that's done.

athomas:
The new script was used for .61 (with a local fix for the bug) and .62. So this change didn't do everything that was necessary. We need to look into why that happened.

athomas:
Most likely the problem is that the new file isn't part of the repo, so the script fails. I can fix that for the .62 release.

athomas:
@nex3 Should be fixed now: https://github.com/dart-lang/homebrew-dart/blob/master/dart%402.rb

nex3:
Thanks!

PR #33515, opened 2018-06-20
trigger error for const A(B()) if B non-const

a14n:
To unblock dart-lang/sdk#57706


bwilkerson (commented):
I think we actually want to fix `test_constructorInvocation_noArgs` by adding an explicit `const` constructor, and then add the test resulting from your changes (where there's an error) as something like `test_constructorInvocation_noConstConstructor`.

Issue #33327, opened 2018-06-04
Cookie.fromSetCookieValue doesn't handle double quotes


tecbot:
Hi,

I'm using the `HttpClient` from the `dart:io` package within a Flutter project. Some websites returning in the `set-cookie` headers, cookie values with double quotes, e.g. instagram.com if expiring cookies:
`'sessionid=""; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Max-Age=0; Domain=instagram.com; Path=/'.
`

If calling `Cookie.fromSetCookieValue` with the example above, the function will throw an error: `Invalid character in cookie value, code unit: '34'` The value is in this case `""`

I found out that the code doesn't handle double quotes. Any plans to implement this behaviour? Go has an implementation for it https://github.com/golang/go/blob/master/src/net/http/cookie.go#L369

I'm happy to contribute, if wanted.

Flutter (Channel dev, v0.5.1, on Mac OS X 10.13.4 17E199, locale de-DE)
Framework revision c7ea3ca377 (6 days ago), 2018-05-29 21:07:33 +0200
Engine revision 1ed25ca7b7
Dart version 2.0.0-dev.58.0.flutter-f981f09760

thismachinechills:
Also encountered this issue with Instagram and other sites.

jonahwilliams:
Unless I am misreading [RFC 6265](https://tools.ietf.org/html/rfc6265), then a cookie-value that is enclosed in double quotes should be valid, and the current Dart SDK implementation is an oversight.
```ABNF
 cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
 cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                     ; US-ASCII characters excluding CTLs,
                     ; whitespace DQUOTE, comma, semicolon,
                     ; and backslash
```

thosakwe:
Can confirm this occurs; also using Instagram.

Sense545:
The pull request #33765 fails for cookies with empty value but no quotes: 
```CookieName=; expires=somedateinthepast```

kevmoo:
This is related to commit https://github.com/dart-lang/sdk/commit/a9ad427ea22ad81c9a06c4c35c04c1ab029c08df

@zanderso thoughts?

stuartsoft:
@sortie any updates on this? Was your review ever merged in?

sortie:
[...]

PR #33765, opened 2018-07-06
Validate cookie values wrapped in double-quotes

alexdelorenzo:
Servers sometimes send headers with cookie values that are encapsulated in double-quotes. Dart should validate values surrounded with double-quotes instead of throwing a FormatException.

This addresses Issue #33327 directly. I applied the solution to this problem that was [solved in Go's code base](https://github.com/golang/go/blob/master/src/net/http/cookie.go#L369).

### DIFF
--- a/sdk/lib/_http/http_headers.dart
+++ b/sdk/lib/_http/http_headers.dart
@@ -986,7 +986,7 @@ class _Cookie implements Cookie {
       }
     }
     
-    if (value[0] == '"' || value[value.length - 1] == '"')
+    if (value[0] == '"' && value[value.length - 1] == '"')
       value = value.slice(1, value.length - 2)
     
     for (int i = 0; i < value.length; i++) {
--- a/sdk/lib/_http/http_headers.dart
+++ b/sdk/lib/_http/http_headers.dart
@@ -987,7 +987,7 @@ class _Cookie implements Cookie {
     }
     
     if (value[0] == '"' && value[value.length - 1] == '"')
-      value = value.slice(1, value.length - 1)
+      value = value.subString(1, value.length - 1)
     
     for (int i = 0; i < value.length; i++) {
       int codeUnit = value.codeUnits[i];
--- a/sdk/lib/_http/http_headers.dart
+++ b/sdk/lib/_http/http_headers.dart
@@ -987,7 +987,7 @@ class _Cookie implements Cookie {
     }
     
     if (value[0] == '"' && value[value.length - 1] == '"')
-      value = value.slice(1, value.length - 2)
+      value = value.slice(1, value.length - 1)
     
     for (int i = 0; i < value.length; i++) {
       int codeUnit = value.codeUnits[i];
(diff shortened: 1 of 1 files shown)


thismachinechills:
I signed it!

kevmoo:
This is active here - https://dart-review.googlesource.com/c/sdk/+/63920

thismachinechills:
I've made changes as per @zanderso's [comments](https://dart-review.googlesource.com/c/sdk/+/63920#message-4a169018542849cf5bd4ea9f6cb770c4977f64ca).

Please see commits made to this pull request [here](https://github.com/dart-lang/sdk/pull/33765/commits/d7f0907108e8d5b7a4712837c47da9d7c74ea8e1). 

A test was added in PR #34102

zanderso:
@kevmoo Will the new commit get mirrored over to Gerrit?

kevmoo:
I'm not sure – @alexander-doroshko @whesse ?

alexander-doroshko:
> I'm not sure – @alexander-doroshko @whesse ?

@kevmoo, sorry, what's the question for me?

whesse:
No, the last commit to this pull request was not mirrored over to the Gerrit CL  https://dart-review.googlesource.com/c/sdk/+/63920
I don't know why that is.

As always, a pull request on SDK should never be landed directly, but the Gerrit CL created from it should be submitted (via CQ, usually).

It should work to check out the Gerrit CL on a new branch, using git cl patch --force, and to fetch the source of this pull request from the foreign remote, do a local merge (or fast-forward merge), and upload the new commit to Gerrit with git cl upload.

Since this change was small, I edited directly in the Gerrit UI to add it.
The Gerrit CL is now up-to-date, and can be committed.
It would be better to have the added test in the same CL, but that seems hard to do at this point.

whesse:
This pull request has both a syntax error, which prevents it from building, and a logical error, which makes the test fail.  Both of these have been fixed in the linked Gerrit CL, but we should not accept pull requests where the author has neither built the code nor tested it.

whesse:
If this change allows Dart to also send invalid cookies with quote characters, as well as receive them, then we should not accept it as-is.

I think we want to continue validating the cookies that a Dart program sends out according to the strict rules, even if we allow incoming cookies to violate them in this way.

Can the author verify that this won't allow Dart to create and send cookies that violate the standard?  Maybe write a test?

It would be best to do further work on the Gerrit CL, since in order to compile this change and run tests, you need to download depot_tools and make a source checkout anyway.

whesse:
OK, looks like the latest spec allows quotes surrounding the whole value, but then those quotes must be retained as part of the value, not dropped:

http://www.ietf.org/rfc/rfc6265.txt
"plus for compatibility with any poor idiots who actually implemented the earlier RFCs it also banned backslash and quotes, other than quotes wrapping the whole value (but in that case the quotes are still considered part of the value, not an encoding scheme). So that leaves you with the alphanums plus:"

Sense545:
This commit breaks cookies which have an empty value but no quotes when server tires to expire it, e.g.
```CookieName=; expires=somedateinthepast```
The change
```
+ if (value[0] == '"' && value[value.length - 1] == '"') {
+  value = value.substring(1, value.length - 1);
+ }
```
should have checked if value.length > 0 before doing it's thing.

PR #33828, opened 2018-07-11
Mark analyzer_cli pubspec as not-to-publish

kevmoo:


bwilkerson:
Should we just delete the `pubspec.yaml` file? (Do we use it for anything if the package isn't being published?

If not, we should also mark the analysis_server package the same way.

devoncarew:
Let's keep the pubspec.yaml file, even if it just has a `name:` entry. This makes tooling happier in a few places (IntelliJ shows better error paths in the analysis view, and the analyzer does a better job of knowing when to create analysis contexts).

devoncarew (approved):
lgtm; agree that similar changes would make sense for pkg/analysis_server

kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/64721

PR #33829, opened 2018-07-12
Update move.yml

devoncarew:
@JekCharlsonYu @kevmoo @pq

### DIFF
--- a/.github/move.yml
+++ b/.github/move.yml
@@ -18,4 +18,4 @@ keepContentMentions: true
 # Set custom aliases for targets
 aliases:
   flutter: flutter/flutter
-  flutter: flutter/flutter-intellij
+  flutter-intellij: flutter/flutter-intellij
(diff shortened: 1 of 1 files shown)


.github/move.yml
```diff
@@ -16,6 +16,6 @@ mentionAuthors: true
 keepContentMentions: true
 
 # Set custom aliases for targets
-# aliases:
-#   r: repo
-#   or: owner/repo
+aliases:
+  flutter: flutter/flutter
+  flutter: flutter/flutter-intellij
```
pq: I'm new to the syntax but it seems a little odd to have one alias for two targets.  Is that right?

devoncarew: No, good catch - the 2nd one should be `flutter-intellij`.

pq (approved):
👍

devoncarew:
Hmm, I wonder how this gets merged through? I'd assumed that a companion CL would be created on gerrit.

devoncarew:
hmm...

PR #34185, opened 2018-08-18
Dirty Fix for https://github.com/dart-lang/sdk/issues/33889

yury-yufimov:
This is a temporary version of fix for
https://github.com/dart-lang/sdk/issues/33889
we are going to use until build_runner is not migrated on newer version of analyzer / front_end


bwilkerson (commented):
I didn't look specifically to see what might have changed between `_BaseElementBuilder` and `_TypesFixBuilder`, but the general approach seems reasonable to me. I'm sorry that the Dart project is in a state where you need this change, but the change itself looks reasonable.

athomas:
Is this PR still relevant? Reading through #33889 it seems we'll never merge it.

PR #34601, opened 2018-09-27
AwaitExpression.precedence => 14

### LINKED ISSUE #34597: unnecessary_parenthesis confused by await
```dart
import 'dart:async';

void test(Future<int> f) async {
  await (f..timeout(const Duration(seconds: 1), onTimeout: () { print('took longer than one second'); }));
  print('finished');
}

void main() {
  test(Completer<int>().future);
}
```

->

```
   info • Unnecessary parenthesis can be removed • test.dart:4:9 • unnecessary_parenthesis
```

...but removing the parentheses changes the logic, because `await` binds more tightly than `..`.

a14n:
Fixes #34597


a14n:
https://dart-review.googlesource.com/c/sdk/+/76880

Issue #32463, opened 2018-03-08
Re-enable whereType after generic methods are turned on in all tools


leafpetersen:
The behavior of the new whereType method is silently broken on platforms that have not yet turned on generic methods (i.e. dart2js and the VM).  We will make this method throw until generic methods are turned on.

See https://github.com/dart-lang/sdk/issues/32423 for context.

Once generic methods are turned on, we need to re-enable this.

leafpetersen:
CL to disable: https://dart-review.googlesource.com/#/c/sdk/+/45744

lrhn:
We can make the method fail *only* when generic methods are not working. 
We should probably do the same to the `cast` which silently doesn't actually cast.

natebosch:
> We can make the method fail _only_ when generic methods are not working.

This is a little risky because we commonly develop/test with DDC and then release with dart2js. We strive for coverage with dart2js as well but it doesn't always work out. It would definitely be much better than silently ignoring the `dynamic`.

> We should probably do the same to the `cast` which silently doesn't actually cast.

I haven't been able to come up with a where that would matter. The runtimes where it won't cast are also the ones that won't care that it isn't the correct reified type... right?

matanlurey:
There certainly is danger in doing something like:

```dart
if (users is List<AdminUser>) {
  ...
}
```

Or

```dart
// In runtimes without obsfucation/minification.
if ('$users.runtimeType' == 'List<AdminUser>') {
  ...
}
```

but for better or worse, that's been the case since Dart 1.0.0. The `.cast<...>` function is in the same class of issues. I wish we were a little more clear about what it did/didn't do it in the various runtimes, but its all good.

whesse:
Shouldn't this have a milestone?  I was surprised when using the Dart 2 documentation, which says this function exists, and then finding it broken.

bkonyi:
I just got burned by this as well on the latest dev release. Is this not making it into 2.0? If not, this should really be documented as there's no mention whatsoever about this feature not being supported and [we currently recommend using it on the Effective Dart page](https://www.dartlang.org/guides/language/effective-dart/usage#do-use-wheretype-to-filter-a-collection-by-type).

lrhn:
This is making it into Dart 2.0, just as soon as we stop running Dart 1 code (or, hopefully, slightly before).

[...]

PR #35040, opened 2018-11-02
Use whereType in status_file_linter.dart

teresy:
Looks like since #32463 is resolved `whereType` can be supported here (I noted the `// TODO(whereType): When whereType is supported, use that.`)


athomas:
Reviewed on https://dart-review.googlesource.com/c/sdk/+/82707

PR #35212, opened 2018-11-19
Fix viewing generated code in Observatory on Windows

### LINKED ISSUE #30201: Observatory crashes when I try to look at Code objects
I consistently get "ServerRpcException(getObject: invalid 'objectId' parameter: code/17dee5fedec-12bfd1ac460)". The id in the exception varies. This is happening to me on Windows. On Linux I don't hit this problem at all.

I am running Dart 1.24.1 on both Windows and Linux.

ds84182:
For GCC and Clang, `long` and `unsigned long` are 64-bit on 64-bit platforms and 32-bit on 32-bit platforms, while on Windows they're always 32-bit. To fix this, I've changed `GetIntegerId` and `GetUnsignedIntegerId` to use the `long long` variants of `strtol` and `strtoul` on 64-bit platforms.

Closes #30201

### DIFF
--- a/runtime/vm/service.cc
+++ b/runtime/vm/service.cc
@@ -268,7 +268,11 @@ static bool GetIntegerId(const char* s, intptr_t* id, int base = 10) {
   }
   intptr_t r = 0;
   char* end_ptr = NULL;
+#if defined(ARCH_IS_32_BIT)
   r = strtol(s, &end_ptr, base);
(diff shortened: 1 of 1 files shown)


mraleph:
Actual review here https://dart-review.googlesource.com/c/sdk/+/84740

(cc @rmacnak-google)

PR #35464, opened 2018-12-20
Rotate method for ListQueue

uroybd:
Added a `rotate` method in ListQueue class which is expected in double-ended queue-like data structure.
Added a test method to test if the functionality working or not.

### DIFF
--- a/sdk/lib/collection/queue.dart
+++ b/sdk/lib/collection/queue.dart
@@ -807,6 +807,21 @@ class ListQueue<E> extends ListIterable<E> implements Queue<E> {
     return result;
   }
 
+  void rotate(int val){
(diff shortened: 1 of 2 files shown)


uroybd:
I have fixed the issue.

kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/87760

@lrhn – thoughts here?

lrhn:
As commented on the code-recview,  https://dart-review.googlesource.com/c/sdk/+/87760, our stability promises mean that we won't change the interface of a platform library class without very good reason, not even for a very reasonable function like this. So, I'm sorry to have to reject this PR.

Issue #28303, opened 2017-01-09
dart:html - Add `innerText` getter to Node class


greglittlefield-wf:
[`Node.innerText`](https://developer.mozilla.org/en-US/docs/Web/API/Node/innerText) is supported by all major browsers, but is not available via the Dart API.

This is behaviorally different from `textContent` (see link for details), which is what the existing [`Node.text`](https://api.dartlang.org/stable/1.21.0/dart-html/Node/text.html) getter uses, and should be supported.

PR #35471, opened 2018-12-21
Add innerText field to Element class

luaVolk:
As mentioned in #28303 Dart needs the innerText getter/setter.

This seems like all thats needed.


kevmoo:
@terrylucas – I think we'll have to do more here, right?

mraleph:
ping?

kevmoo:
@terrylucas  @vsmenon

mit-mit:
This has landed, thanks for contributing @luaVolk

terrylucas:
This would be removed when dart:html is regenereated.  I will add to the appropriate template file and add the missing @JSName('innerText').

PR #35507, opened 2018-12-27
Fix wording

mdebbar:
The current wording seems to be saying the opposite of what it's supposed to.


kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/88123

PR #35568, opened 2019-01-04
feat: remove constructor to allow for use as mixin

jarrodcolburn:
Lazily set `_frequency` rather than initialize it in Constructor. Without constructor, can use class as a mixin.


JarrodCColburn:
I signed it!

JarrodCColburn:
i made a typo.

PR #35661, opened 2019-01-15
Don't write dill to disk when there is compilation error during dill generation(Flutter's Debug mode)

kangwang1988:
This PR comes from https://github.com/flutter/flutter/issues/23121
I found that app.dill will always generate no matter if there is any compilation problem when generating dill.
1. Create a hello_world project without error
app.dill is generated without reporting any error.
2. Mock a error(First dill compilation)
app.dill is generated with reporting error.
3. Second dill compilation
app.dill is not changed as already compiled and no changes. Henceforth, it reports no error.

@mraleph 
Please have a look if this is a bug or as designed.

### DIFF
--- a/pkg/vm/lib/frontend_server.dart
+++ b/pkg/vm/lib/frontend_server.dart
@@ -349,8 +349,10 @@ class FrontendCompiler implements CompilerInterface {
         transformer.transform(component);
       }
 
-      await writeDillFile(component, _kernelBinaryFilename,
-          filterExternal: importDill != null);
+      if (errors.isEmpty) {
+        await writeDillFile(component, _kernelBinaryFilename,
+            filterExternal: importDill != null);
+      }
 
       _outputStream
           .writeln('$boundaryKey $_kernelBinaryFilename ${errors.length}');
@@ -456,7 +458,9 @@ class FrontendCompiler implements CompilerInterface {
     if (deltaProgram != null && transformer != null) {
       transformer.transform(deltaProgram);
     }
-    await writeDillFile(deltaProgram, _kernelBinaryFilename);
+    if (errors.isEmpty) {
+      await writeDillFile(deltaProgram, _kernelBinaryFilename);
+    }
     _outputStream
         .writeln('$boundaryKey $_kernelBinaryFilename ${errors.length}');
     _kernelBinaryFilename = _kernelBinaryFilenameIncremental;


zoechi:
An alternative might be to keep the errors and just print them again if the dill file is reused.

kmillikin:
It's a fundamental design choice to generate code you could run even when there are static errors.  This means you could make a change to your UI then reload and try it out to see if you're on the right track without spending possibly wasted time fixing errors.

We (the tool) doesn't decide if you are allowed to run your code, you (the developer) decides.

So we don't want this change.

It sounds like a bug in the incremental compiler that we don't signal errors when you recompile.  I think we'll add to a dill file a list of errors that were signaled, so the compiler can report them again without recompiling.

/cc @jensjoha

kangwang1988:
@kmillikin 
Even though it works as expected, It would be better that the developers(like flutter_tools) can determine whether there is an error during this compilation. So that, developers can determine if to allow the dill to get running.
Now, the exitcode with value of 0 make it difficult for developers to determine if their is something wrong in the compilation process.

eernstg:
Ideally, we'd give developers (1) unlimited access to information about program faults (errors, warnings, lints, whatever we can detect) as soon as they are known to exist, and (2) support for exploratory execution of programs that have no errors, as well as programs that only have faults for which a development execution platform has defined a semantics. (Dart 1 defined a semantics for some constructs which are currently errors, e.g., `x is UnknownType` would throw when executed, but in Dart 2 it is an implementation specific property which erroneous constructs can be executed, if any, and how).

So would it be possible to go in this direction? Presumably, this means that it should be possible to generate a dill file which contains code that gives semantics to an erroneous construct, but that dill file would then be marked 'erroneous', and tools would be able to detect this status (for instance, a deployment compilation should never include such a dill).

kmillikin:
> Even though it works as designed, It would be better that the developers(like flutter_tools) can determine whether there is an error during this compilation. So that, developers can determine if to allow the dill to get running.

I didn't say it was working as designed.  I said there was a bug.

kangwang1988:
> > Even though it works as designed, It would be better that the developers(like flutter_tools) can determine whether there is an error during this compilation. So that, developers can determine if to allow the dill to get running.
> 
> I didn't say it was working as designed. I said there was a bug.

Sorry for misleading.

mraleph:
I have filed https://github.com/dart-lang/sdk/issues/35666 to track the actual bug based on @kmillikin's feedback.

I will now close this PR. 

Thanks for bringing our attention to the problem @kangwang1988!

kangwang1988:
@mraleph You're welcome.
@eernstg 
So you mean that though the main procedure remains, it will be possible for the developers to detect the error details(if any) contained in the dill, right?

mraleph:
@kangwang1988 yes, that is correct. Currently it is not possible, but Kernel format will be changed to allow easy listing of all errors it contains.

eernstg:
@kangwang1988 wrote: 
> So you mean ..

@mraleph already answered about the actual status and plans; I think that new Kernel format will work nicely to ensure the ideal that I mentioned (being able to report all known program faults, and being able to run programs even though they have some errors).

PR #35784, opened 2019-01-28
Use setAttribute instead of the IDL property for CSP nonce assignment

mikispag:
We should use ``HTMLElement.setAttribute('nonce', nonceValue)`` instead of directly assigning the IDL property ``element.nonce = nonceValue`` due to Firefox not supporting the nonce IDL property ([Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1389421)).

Fixes: #35784

### DIFF
--- a/sdk/lib/_internal/js_runtime/lib/js_helper.dart
+++ b/sdk/lib/_internal/js_runtime/lib/js_helper.dart
@@ -3508,7 +3508,7 @@ Future<Null> _loadHunk(String hunkName) {
     JS('', '#.type = "text/javascript"', script);
     JS('', '#.src = #', script, uri);
     if (_cspNonce != null && _cspNonce != '') {
-      JS('', '#.nonce = #', script, _cspNonce);
+      JS('', '#.setAttribute("nonce", #)', script, _cspNonce);
     }
     if (_crossOrigin != null && _crossOrigin != '') {
       JS('', '#.crossOrigin = #', script, _crossOrigin);


sigmundch (commented):
Thanks @mikispag for filing the bug and sending this fix!

I have a few suggestions before we merge this in. We should make sure the following 2 workflows work well: setting `nonce` programatically and setting `nonce` as an attribute on the HTML. In particular:
* to support reading the `nonce` attribute from the HTML, we need to update `_computeCspNonce` so we also try to read the data from `getAttribute`, if present.
* to support setting it programatically, we should also read it from the property. One subtle question is whether we should do it only if the attribute was null, or read both and give priority to the property if it is non-null? I'm inclined to do the latter.
* Finally, when users set `nonce` as a property, they may expect it to be available as a property on the deferred scripts as well, so in that case, I suggest we do both, call `setAttribute` and set the property on the new script tag.

What do you think?

I'm not sure if you have the setup to run our tests, if you do, we have a test for this under `tests/compiler/dart2js_extra/deferred_with_csp_nonce_test.dart` that we should update to check also for the attribute.

I'm happy to help complete these changes too, just let me know. Thanks again!

mikispag:
Thanks,

- ``_computeCspNonce`` should use ``.nonce || .getAttribute('nonce')`` in JS. What is the right ``JS(...)`` syntax? Sorry, I am not a Dart coder.
- In recent Blink, `.getAttribute('nonce')` will return ``undefined`` for security reasons, so it is fine to give priority to the property if it is non-null, as you suggested. In general, the IDL property is the reference as it can't be easily changed from outside JS-land.
- We can use ``setAttribute`` and ``.nonce =`` at the same time. It is fine.
- Can you help with the tests? Thank you!

Issue #26087, opened 2016-03-25
List.removeLast() throws an inconsistent error


nex3:
The native List type, at least on the VM, throws a `RangeError` if `removeLast()` is called on an empty list. A list created using `ListBase` or `ListMixin`, on the other hand, throws a `StateError` in the same situation. The following reproduces this:

```dart
import 'dart:collection';

class CustomList extends ListBase {
  final List _inner;

  int get length => _inner.length;
  set length(int value) => _inner.length = value;

  CustomList() : _inner = [];

  operator [](int index) => _inner[index];
  operator []=(int index, value) => _inner[index] = value;
}

void main() {
  try {
    new CustomList().removeLast();
  } catch (error) {
    print(error);
  }

  try {
    [].removeLast();
  } catch (error) {
    print(error);
  }
}
```

I'm using 8922198e78b37c9ebc2d9d7e8362a7dea139c4ce.

rakudrama:
VM:
Bad state: No element
RangeError (index): Invalid value: Valid value range is empty: -1

dart2js:
Bad state: No element
RangeError (index): Index out of range: index must not be negative: -1

lrhn:
The best error is "Bad state: No element". The user has not provided an index explicitly, so complaining about an index is misleading. Having a "Bad state: List is empty" might be more informative than using "Bad state: No element", which is also used for other situations where the list isn't empty.

When using a range error, the implementations are apparently passing different parameters to, or using different constructors of, `RangeError`. To be honest, neither message is great for this case - the former is saying that there is no valid index at all (which best matches that the collection is empty), but the latter claims that a non-negative index could be valid.

rakudrama:
I support changing the warning but it does require implementation work in
dart2js.

`removeLast` is specially inlined to use the same 'bounds check'
instruction as indexing, so the index can be eliminated in code like

```
while (stack.isNotEmpty) {
   process(stack.removeLast());
}
```

The changes would be to make the 'bounds check' instruction also generate
the StateError via js_runtime helper functions, and properly register those
functions since they are called only from generated code. Currently we
pretend that `-1` was used as an index, which matches the VM which relies
on indexing to signal an error i.e.  `a[a.length-1]`.

The first step would be to write tests that demonstrate dart2js gives the
same error on code that is highly optimized or unoptimized.

floitschG:
If it's easy, we should change to StateErrors.
[...]

PR #35894, opened 2019-02-10
List.removeLast() docs error occurs if empty

camsteffen:
Seems like this should be explicitly stated in the docs. I did not indicate the specific type of error because of #26087.

### DIFF
--- a/sdk/lib/core/list.dart
+++ b/sdk/lib/core/list.dart
@@ -537,6 +537,8 @@ abstract class List<E> implements EfficientLengthIterable<E> {
   /**
    * Pops and returns the last object in this list.
    *
+   * The list must not be empty.
+   *
    * Throws an [UnsupportedError] if this is a fixed-length list.
    */
   E removeLast();


lrhn:
Thanks for the suggestion. I'll probably phrase it differently, but it should be said somehow.

mraleph:
@lrhn any ideas about alternative wording? Should this PR be closed if suggested wording does not satisfy your requirements?

ulidtko:
I'd also like to see it documented *which* error exactly will happen. But, since #26087 is `closed-not-planned` (which means WONTFIX I guess), the error will indeed be of varying type.

I'd merge this PR easily **as-is**; for an extra mile of effort, I'd improve it further:

>   * If the list is empty, this method throws an exception (of unspecified type, see https://github.com/dart-lang/sdk/issues/26087).

sdk/lib/core/list.dart
```diff
@@ -537,6 +537,8 @@ abstract class List<E> implements EfficientLengthIterable<E> {
   /**
    * Pops and returns the last object in this list.
    *
+   * An error occurs if the list is empty.
```
lrhn: Just say: 
```
 * The list must not be empty.
```

This is how we document prerequisites in the platform libraries. 
If a "must" requirement is not satisfied, the implementation will *likely* throw an `Error`, but there are situations where the throwing is best-effort only (mainly concurrent modification).

lrhn (commented):
Than you for the PR. I agree that it's a good idea to document the requirement. 
We have a standard way of writing such prerequisites in platform library docs, so I have suggested that wording.

lrhn (commented):
Than you for the PR. I agree that it's a good idea to document the requirement. 
We have a standard way of writing such prerequisites in platform library docs, so I have suggested that wording.

camsteffen:
I changed the wording as suggested.

mit-mit:
Signoff in https://dart-review.googlesource.com/c/sdk/+/107822

PR #36514, opened 2019-04-08
Update `dart2aot` help output

mit-mit:


mraleph:
I really think this script should be written in Dart and be distributed in its own package, rather than in SDK.

mit-mit:
This is now https://dart-review.googlesource.com/c/sdk/+/98820 ; please review there

Issue #28793, opened 2017-02-16
Support python3 in our various build and test scripts


munificent:
We use Python for a bunch of our internal tools and automation. I believe many of them currently don't work under Python 3. That makes it harder for external users that don't have Python 2.7 installed to get up and running.

We should probably get those scripts working under both 2 and 3, deprecate them if they aren't needed anymore, or rewrite them in Dart where appropriate.

eseidelGoogle:
I thought python 2 and 3 were mutually exclusive?  There is a conversion tool, but I think once you've converted you can't go back. :)
https://wiki.python.org/moin/Python2orPython3

jcollins-g:
It's possible to write Python code that runs in both, but the TL;DR is that
it is annoying to do so.  This is a somewhat out of date guide but will
give you the general idea:

http://python-future.org/compatible_idioms.html

If we decide to do it, making our Python tooling work in both will be an
exercise in in using and/or finding compatible idioms along these lines.

Janice

DanTup:
I don't know how much more effort (if any) it'd be to rewrite in Dart over making work in both Pythons, but that would not only eliminate another dependency but it's also good dog-fooding (or "putting your money where your mouth is"). Most people wanting to hack on Dart are going to have an SDK already whereas Python might not be available (for ex., most Windows users trying out Dart probably don't have either Python).

srawlins:
CC @athomas @whesse Python 2 is EOL in < 2 months, on Jan 1, 2020. https://pythonclock.org/

SchoolGuy:
So just a ping from my side as I am packaging in openSUSE if I have time. If we get this Py3 I could try to build an openSUSE/Fedora/CentOS/... Specfile to make Dart available on these distributions if legal is fine with it. But since no Py2 packages are accepted to openSUSE Leap and Tumbleweed Dart will not be available there...

natebosch:
I made some changes in my dev setup and hit this again - python2 has not been maintained for over a year now and we shouldn't be asking authors to install it to contribute to the SDK.

Do we have a timeline for this?

jubalh:
> So just a ping from my side as I am packaging in openSUSE if I have time. If we get this Py3 I could try to build an openSUSE/Fedora/CentOS/... Specfile to make Dart available on these distributions if legal is fine with it. But since no Py2 packages are accepted to openSUSE Leap and Tumbleweed Dart will not be available there...

Actually I wanted to do this > 1 year ago.
[...]

PR #36662, opened 2019-04-17
Fix build when python=python3

josephlr:
Right now most of the dart SDK's python is compatible with python2
or python3. This change fixes a few of the build scripts to make
that completely true (at least when building the standard build on
Linux). There are only four types of changes:
  - Bare `print` statements now use the `print ()` function
  - `commands.getoutput` becomes `subprocess.check_output` with `shell=True`
  - `xrange` becomes `range`
  - `print >> sys.stderr` becomes `sys.stderr.write`

Starts work on addressing (but does not completely fix):
https://github.com/dart-lang/sdk/issues/28793

See related issue:
https://fuchsia-review.googlesource.com/c/fuchsia/+/272925

This change applys to both the `dev` and `master` branches.

Change-Id: Ibd3eb9b1f57520d2d745f05c2ac430b1d20943da

### DIFF
--- a/build/gn_run_binary.py
+++ b/build/gn_run_binary.py
@@ -48,7 +48,7 @@ def main(argv):
     path = './' + argv[2]
 
   if not os.path.isfile(path):
-    print "Binary not found: " + path
+    print ("Binary not found: " + path)
     return error_exit
 
   # The rest of the arguments are passed directly to the executable.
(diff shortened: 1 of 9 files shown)


mraleph:
Gerrit issue: https://dart-review.googlesource.com/c/sdk/+/99707

Note: European part of the Dart team are on Easter holidays until Tuesday.

PR #36683, opened 2019-04-20
Check window.MediaSource before call constructor

gmpassos:
MediaSource is an experimental technology. Not compatible with iOS Safary, even with current last iOS version (v12 April 2019).

This JS  function (polyfill) crashes web compiled Dart load on iOS.

### DIFF
--- a/pkg/dev_compiler/tool/input_sdk/private/ddc_runtime/runtime.dart
+++ b/pkg/dev_compiler/tool/input_sdk/private/ddc_runtime/runtime.dart
@@ -90,8 +90,10 @@ bool polyfill(window) => JS('', '''(() => {
       }
     }
     if (typeof $window.SourceBufferList == "undefined") {
-      $window.SourceBufferList =
-        new $window.MediaSource().sourceBuffers.constructor;
+      if ('MediaSource' in $window) {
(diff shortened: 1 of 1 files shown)


gmpassos:
Note that without this fix Dart crashes in iOS Safari.

JS generated by:
```
webdev serve --hostname 0.0.0.0 -r
```

gmpassos:
The function `polyfill` should use only standard JS, and this should be documented/commented in the code at least.

vsmenon:
Thanks for the context and patch.  We'll need to run this through the standard Dart SDK contribution process (including gerrit to run on our submit queue):

https://github.com/dart-lang/sdk/blob/master/CONTRIBUTING.md

If you'd prefer, one of us can take this over.

gmpassos:
> Thanks for the context and patch. We'll need to run this through the standard Dart SDK contribution process (including gerrit to run on our submit queue):
> 
> https://github.com/dart-lang/sdk/blob/master/CONTRIBUTING.md
> 
> If you'd prefer, one of us can take this over.

@vsmenon 
I saw that the patch is already at master branch. Thanks for the merge.

* Don't forget to add me in the contributors list, I will be very glad.

PR #36927, opened 2019-05-10
Fixed some links

LorenzNickel:
**I fixed some links and some other minor flaws.**

I also found the following outdated links, which I didn't fix:

- [This link](https://github.com/dart-lang/pool/tree/zone.strong) in [this file](https://github.com/dart-lang/sdk/blob/master/docs/newsletter/20170728.md) (didn't fix because don't know new location and this is some sort of archive so the link should probably stay the original)
- [This link](https://www.dartlang.org/tools/analyzer) in [this file](https://github.com/dart-lang/sdk/blob/master/pkg/analyzer_cli/lib/src/options.dart) (didn't fix since the link is still working, it just gets redirected and the link is part of the programm, I don't want to break anything by changing it, all other links are in comments)
- [This link](https://github.com/domokit/mojo/issues/728) in [this file](https://github.com/dart-lang/sdk/blob/master/build/config/compiler/BUILD.gn) (didn't fix since probably has no new location and is part of TODO, which I don't want to change)

While doing all this I also noticed that [these tests](https://github.com/dart-lang/sdk/tree/master/tests/compiler/dart2js_extra) contain a lot of other inconsistencies in their comments and the location of the import-statements (I only fixed one which was not link-related), could be target of another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ See [LICENSE][license] and [PATENTS][patents].
 [repo]: https://github.com/dart-lang/sdk
 [lang]: https://www.dartlang.org/docs/dart-up-and-running/ch02.html
 [tools]: https://www.dartlang.org/tools/
-[codelab]: https://www.dartlang.org/codelabs/darrrt/
+[codelab]: https://dart.dev/codelabs
 [dartbug]: http://dartbug.com
 [contrib]: https://github.com/dart-lang/sdk/wiki/Contributing
 [pubsite]: https://pub.dartlang.org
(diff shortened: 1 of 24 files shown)


kevmoo:
Wow! Awesome! CC @athomas 

there is some magic to get this going through https://dart-review.googlesource.com/q/status:open – stay tuned!

mit-mit:
Thanks much @LorenzNickel for this PR, and sorry that it got stuck!

PR #37313, opened 2019-06-19
Add: Example usage to fillRange method.

abhishekkanojia:


kevmoo:
@lrhn ?

abhishekkanojia:
@kevmoo Added one more commit which updates the insert method documentation.

mit-mit:
Reviewing in https://dart-review.googlesource.com/c/sdk/+/106724

sdk/lib/core/list.dart
```diff
@@ -464,7 +464,7 @@ abstract class List<E> implements EfficientLengthIterable<E> {
    * This increases the length of the list by one and shifts all objects
    * at or after the index towards the end of the list.
    *
-   * An error occurs if the [index] is less than 0 or greater than length.
+   * A `RangeError` occurs if the [index] is less than 0 or greater than length.
```
lrhn: I'd prefer to not say whic error is throws. Instead replace this entire section by:
```
 * The list must be growable. 
 * The [index] value must be non-negative and no greater than [length].
```

sdk/lib/core/list.dart
```diff
@@ -674,6 +674,22 @@ abstract class List<E> implements EfficientLengthIterable<E> {
    * A range from [start] to [end] is valid if `0 <= start <= end <= len`, where
    * `len` is this list's `length`. The range starts at `start` and has length
    * `end - start`. An empty range (with `end == start`) is valid.
+   *
+   *     List<int> list1 = new List(3);
```
lrhn: I'd start with:
```
 * 
 * Example:
```
to set the expectation.

Also, change `list1` to just `list`, the "1" doesn't add anything to the example.

sdk/lib/core/list.dart
```diff
@@ -674,6 +674,22 @@ abstract class List<E> implements EfficientLengthIterable<E> {
...
+   *     list1.fillRange(0, 2, 1);
+   *     print(list); //  [1, 1, null]
+   *
+   * Note: Usage on dynamic list throws RangeError: Invalid Value
+   *
+   *      List<int> list = new List();
+   *      list.fillRange(0, 2, 1); // throws `RangeError: Invalid Value` exception.
+   *
```
lrhn: This is an example of what not to do, which we try not to do (unless the error is particularly interesting, this one isn't).

I'd just drop this example.

sdk/lib/core/list.dart
```diff
@@ -674,6 +674,22 @@ abstract class List<E> implements EfficientLengthIterable<E> {
...
+   *
+   * However, works on dynamic list of fixed length:
+   *
+   *      List<int> list2 = new List()..length = 3;
+   *      list2.fillRange(0, 2, 1);
+   *      print(list2); // [1, 1, null]
```
lrhn: This example is exactly the same as the first one, except that the list is growable. Since this function doesn't care about being growable, the example is not adding anything new. I'd remove this example too.

sdk/lib/core/list.dart
abhishekkanojia: I added this keeping in mind it as an extra note or gotcha while using the fillRange method. But with your suggested comment:
` The list must be growable. `
It give user a sense that using this method might result in some error. 
I will go ahead and remove this example.

sdk/lib/core/list.dart
```diff
@@ -674,6 +674,22 @@ abstract class List<E> implements EfficientLengthIterable<E> {
    * A range from [start] to [end] is valid if `0 <= start <= end <= len`, where
    * `len` is this list's `length`. The range starts at `start` and has length
    * `end - start`. An empty range (with `end == start`) is valid.
+   *
```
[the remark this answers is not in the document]
abhishekkanojia: Noted

sdk/lib/core/list.dart
abhishekkanojia: awesome

sdk/lib/core/list.dart
abhishekkanojia: okay

sdk/lib/core/list.dart
abhishekkanojia: I think this applies to `insertAll` method as well. I am updating it for that as well.

mit-mit:
Thank for contributing @abhishekkanojia

PR #37404, opened 2019-06-30
Fix example syntax

tomgilder:
Update example syntax - HttpClient has a named argument for SecurityContext

### DIFF
--- a/sdk/lib/_http/overrides.dart
+++ b/sdk/lib/_http/overrides.dart
@@ -26,7 +26,7 @@ const _asyncRunZoned = runZoned;
 ///     // Operations will use MyHttpClient instead of the real HttpClient
 ///     // implementation whenever HttpClient is used.
 ///     ...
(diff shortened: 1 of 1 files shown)


mit-mit:
Reviewing in https://dart-review.googlesource.com/c/sdk/+/107760

sjtaelman (approved):

mit-mit:
Thanks for contributing @tomgilder. This should be merging shortly.

mraleph:
Note that the change to the comment made it incorrect so I reverted the change. It explicitly declares a separate class `MyHttpClient` - rather than use an existing `HttpClient` class and then uses overrides to tell `dart:io` to create instances of `MyHttpClient` instead of `HttpClient`. So changing `MyHttpClient` to `HttpClient` in the comment completely changed the meaning of the comment.

sjtaelman:
Hkgy8

PR #37409, opened 2019-07-01
Consistently format breaking changes

mit-mit:
Moved the bug-id to the breaking change prefix. This also makes it easy to validate that an approved breaking change issue exists.


mit-mit:
This is https://dart-review.googlesource.com/c/sdk/+/107820

PR #37436, opened 2019-07-03
Update list.dart

### LINKED ISSUE #37262: Explain that List.== does not compare contents, link to collections.ListEquality
I had difficulty finding out how to compare the contents of two List objects.  My first stop was the List.== operator:
https://api.dartlang.org/stable/2.3.2/dart-core/List-class.html#operators
That page links to Object.== operator.  I spent 15 minutes finding the [answer](https://stackoverflow.com/a/22333042).  How about updating the docs to helpfully state that List.== does not compare contents, but this can be done with the collections.ListEquality class?

MetalDent:
This PR fixes #37262.

### DIFF
--- a/sdk/lib/core/list.dart
+++ b/sdk/lib/core/list.dart
@@ -718,11 +718,11 @@ abstract class List<E> implements EfficientLengthIterable<E> {
    */
   Map<int, E> asMap();
 
-  /** 
+  /**
   * Whether this list is equal to [other].
   *
   * Lists are, by default, only equal to themselves.
-  * Even if [other] is also a list, the equality comparison 
+  * Even if [other] is also a list, the equality comparison
   * does not compare the elements of the two lists.
   */
  bool operator ==(Object other);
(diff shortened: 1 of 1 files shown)


mit-mit:
Review in https://dart-review.googlesource.com/c/sdk/+/108180

mit-mit:
@MetalDent thanks for the contribution! Please see feedback in the above review link

MetalDent:
> Review in https://dart-review.googlesource.com/c/sdk/+/108180

Okay, so you don't want me to make a separate class?

lrhn:
The extra `List` class won't work, there is already one `List` class declared in that library. The only issue is that the existing `List` class inherits the `operator==` method from `Object`, including the documentation.
By adding an abstract `operator==` method to the existing `List` class, we can add a new documentation for the method, without changing the inherited implementation.

lrhn:
Looks good!
Thanks.

PR #37796, opened 2019-08-09
Add defined() guard for checkout_llvm build flag

cbracken:
Adds a guard before uses of the checkout_llvm flag, defaulting it to
false when unset. This prevents breakages when used in trees that hasn't
defined that gn flag.

Also fixed manually in topaz in:
https://turquoise-internal-review.googlesource.com/c/integration/+/133882

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -16,7 +16,7 @@ group("default") {
   deps = [
     ":runtime",
   ]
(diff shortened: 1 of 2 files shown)


cbracken:
Closing to send via git cl upload. It's been a while since I've sent a patch to the SDK it seems :)

PR #37883, opened 2019-08-15
Updating the package:js README

kwalrath:


pkg/js/README.md
```diff
@@ -1,16 +1,28 @@
...
+This package's main library, `js`, provides annotations and functions
+that let you specify how your Dart code interoperates with JavaScript code.
+The Dart-to-JavaScript compilers — dartdevc and dart2js — recognize these
+annotations, using them to connect your Dart code with JavaScript.
+
+**Important:** This library supercedes `dart:js`, so don't import `dart:js`.
+Instead, import `package:js/js.dart`.
+
+A second library in this package, `js_util`, provides low-level utilities
+that you can use if you need to read property values.
```
natebosch: I would phrase this as something like "that you can use as a fallback where it is not possible to wrap with a static annotated API."

pkg/js/README.md
```diff
@@ -23,7 +35,7 @@ import 'package:js/js.dart';
 external String stringify(Object obj);
 ```
 
-#### Classes and Namespaces
+### Using namespaces and defining classes
```
natebosch: I'm not sure if we want to call this "defining classes". The `class` we write here is a representation of the interface of a JS class which is defined elsewhere.

natebosch (approved):

pkg/js/README.md
kwalrath: Maybe: Using JavaScript namespaces and classes
?

kwalrath:
I've addressed your comments (thanks!) and removed/hidden the remaining [PENDING]s.

mit-mit:
@kwalrath don't forget that in the SDK repo you cannot land PRs directly here from the GitHub UI. You have to merge from Gerrit: https://dart-review.googlesource.com/c/sdk/+/113265

PR #38181, opened 2019-09-03
Update CHANGELOG.md

mit-mit:
Clarify cast example for const expressions


mit-mit:
Review is https://dart-review.googlesource.com/c/sdk/+/115260

PR #38396, opened 2019-09-14
explain how to write to `asExternalTypedData`

truongsinh:


mraleph:
Actual review will happen on https://dart-review.googlesource.com/c/sdk/+/117260

mraleph:
I think the resolution here is that this is not going to land.

PR #38558, opened 2019-09-25
Bump analyzer_plugin's maximum supported analyzer version

ReginFell:
analyzer_plugin has max constraint to analyzer version 0.38.0 while the latest version is 0.38.3 and it causes a lot of issues for libraries that use it (built_value for instance)

### DIFF
--- a/pkg/analyzer_plugin/pubspec.yaml
+++ b/pkg/analyzer_plugin/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
   sdk: '>=2.0.0 <3.0.0'
 
 dependencies:
-  analyzer: '>=0.35.3 <0.38.0'
+  analyzer: '>=0.35.3 <0.38.4'
   charcode: '^1.1.0'
   dart_style: '^1.2.0'
   html: '>=0.13.1 <0.15.0'
--- a/pkg/analyzer_plugin/pubspec.yaml
+++ b/pkg/analyzer_plugin/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
   sdk: '>=2.0.0 <3.0.0'
 
 dependencies:
-  analyzer: '>=0.35.3 <0.38.4'
+  analyzer: '>=0.35.3 <0.39.0'
   charcode: '^1.1.0'
   dart_style: '^1.2.0'
   html: '>=0.13.1 <0.15.0'


kevmoo:
CC @devoncarew @stereotype441

bwilkerson (approved):

kevmoo:
This won't help, sadly. Folks will just get the previous version before this restriction.

pkg/analyzer_plugin/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ environment:
   sdk: '>=2.0.0 <3.0.0'
 
 dependencies:
-  analyzer: '>=0.35.3 <0.38.0'
+  analyzer: '>=0.35.3 <0.38.4'
```
devoncarew: `<0.39.0`?

devoncarew:
> This won't help, sadly. Folks will just get the previous version before this restriction.

@ReginFell - can you provide a bit more details about the issue? I don't know if bumping the version range here will help or not.

ReginFell:
Yeah, the main issue at the moment is built_value (https://github.com/google/built_value.dart) 
As I mentioned before, the majority of flutter applications use it. 
It depends on analyzer_plugin which has a top constraint on 0.38.0 so we stuck with this version https://github.com/google/built_value.dart/blob/master/built_value_generator/pubspec.yaml

Several libraries already updated to ^0.38.2 and here where the problem comes in, we can't use built_value with libraries that are on a higher version of analyzer, while built_value can't be updated to the newest analyzer cos it has a dependency on analyzer_plugin

pkg/analyzer_plugin/pubspec.yaml
kevmoo: Oh! I get it now. This should be fine...I think...

devoncarew: @ReginFell, if this looks good, can you make this change.

@bwilkerson - do you mind shepherding https://dart-review.googlesource.com/c/sdk/+/118721 in?

bwilkerson: No, I don't mind.

ReginFell: Yeah, done

ReginFell:
@googlebot I fixed it.

pkg/analyzer_plugin/pubspec.yaml
bwilkerson: Thanks!

simolus3:
Is there anything preventing a release of the `analyzer_plugin` package with the updated dependency range? It would be very helpful to publish this so that we can resolve incompatibilities, thanks!

bwilkerson:
@stereotype441

stereotype441:
@simolus3 
> Is there anything preventing a release of the `analyzer_plugin` package with the updated dependency range? It would be very helpful to publish this so that we can resolve incompatibilities, thanks!

Sorry about that!  I didn't realize the fix hadn't been published.  I'll try to get a new version of analyzer_plugin out today.

stereotype441:
@simolus3
> > Is there anything preventing a release of the `analyzer_plugin` package with the updated dependency range? It would be very helpful to publish this so that we can resolve incompatibilities, thanks!
> 
> Sorry about that! I didn't realize the fix hadn't been published. I'll try to get a new version of analyzer_plugin out today.

Published: https://pub.dev/packages/analyzer_plugin/versions/0.2.1

PR #38614, opened 2019-09-27
WIP: Fix various bugs in nullable propagation

### LINKED ISSUE #38585: nnbd: Analyzer not assigning nullability to generic parameter, inferred return type; bang in index-postfix-expression treated wrongly; propagation in collection elements; broken non-nullable promotion of type parameters
This ticket tracks various bugs in the Dart Analyzer regarding NNBD.

PR for fixing: https://github.com/dart-lang/sdk/pull/38614

**BUGS FIXED IN THE PR** 

```
class ClassT<T> {
  T? nullable() => null;
}

void fooT() {
  final c = ClassT<String>();
  c.nullable()?.length;                  // <-- error here, correct: no error since it can be null
}
```

```
void test() {
  // ignore: unnecessary_statements
  () {
    if ('foo' == 'test') {
      return null;                          // <-- error here, correct: return type should be String?
    }
    return 'test';
  };
}
```

```
class Class<T> {
  T? operator [](String key) => null;
}

void foo() {
  final x = Class<String>();

  x['X']!.length;                   // <-- error here, correct: bang is fine since [] returns String?
}
```

```
class Class {
  final String member;
}

void foo() {
  Class? x;
  x?.member?.length;   // <-- error here, but correct since ? turns `x.member` into `String?` 
}
```

```
typedef Callback = void Function(void);
void foo({Callback? foo}) {}  // <-- error MISSING_DEFAULT_VALUE_FOR_PARAMETER
```

```
List<int> foo(String? s) {
  return [
    if (s != null) s.length,   // <-- error
   ]
}
```

**NOT FIXED YET IN THE PR**

```
void foo<T>(T? v, void Function(T value) action) {
  if (v == null) {
    return;
  }
  action(v);             // <-- error
}
```

derolf:
Fixes https://github.com/dart-lang/sdk/issues/38585


derolf:
okay, this looks like getting into a rabbit hole...

mit-mit:
Closing: this has no signed CLA and appears stale

PR #38706, opened 2019-10-03
Clarify error message for instantiating abstract classes

### LINKED ISSUE #38695: new_with_abstract_class error message is confusing in Dart 2
Running `dartanalyzer` on the following code:

```dart
abstract class Foo {}

final f = Foo();
```
generates:

>   error • Abstract classes can't be created with a 'new' expression at foo.dart:3:11 • new_with_abstract_class
> 1 error found.

This is a bit confusing since I didn't use `new`.  It would be better if that were removed from the error message.

(I'm using Dart 2.5.1 on Linux x64.)

mit-mit:
Fixes https://github.com/dart-lang/sdk/issues/38695

Consistent with terminology in https://dart.dev/guides/language/language-tour#abstract-classes


mit-mit:
Code review is https://dart-review.googlesource.com/c/sdk/+/119880

Issue #24206, opened 2015-08-26
Object.hashCode is slow?


asgerf:
Overriding Object.hashCode with a simple integer counter can provide shocking performance improvements. I.e.

```
class Foo {
  static int _hashCounter = 1;
  final int hashCode = ++_hashCounter; // Better than Object.hashCode
}
```

It seems the VM uses a random-number generator to create fresh hash codes, and then stores them in a weak hash table to associate it with the object. Profiling the dart2js compiler seemed to indicate that `Object_getHash` (which performs the lookup, not the computation) is a bottleneck.

Maybe the hash code could be stored as a field? There is something ironic about storing hash codes in a hash table.

I don't have a minimal reproduction case, but you can reproduce by running revision 90039f64ae of the dart2js compiler with the `--cps-ir` flag on the ImagingGaussianBlurOnce benchmark:

```
git checkout 90039f64ae
time dart -pout/ReleaseIA32/packages pkg/compiler/lib/src/dart2js.dart GOLEM/benchmarks/ImagingGaussianBlurOnce/dart/ImagingGaussianBlurOnce.dart --use-cps-ir
```

Then apply https://codereview.chromium.org/1311673009 to see the difference.

kodandersson:
Regarding the current design: Since most objects will not have their hash code computed, dedicating a word in all of them was deemed wasteful. The default implementation of Object.hashCode uses an RNG rather than object addresses to allow for the GC relocating objects, although currently, we only relocate objects in the young generation.

munificent:
For what it's worth, this was a noticeable performance issue in dartfmt, that I had to [optimize by having my own hash code base class](https://github.com/dart-lang/dart_style/blob/master/lib/src/fast_hash.dart).

iposva-google:
To give you an idea of the overhead Daniel mentioned in his comment I ran dart2js (the default version) compiling itself and collected the allocation information. From that I selected the top-ten object allocations. Based on the average object size I can give you a percentage overhead of adding a field to each of these objects. None of the objects in this list would benefit from an extra hash code field, but since they are all instances you suggest they have one. The size is the cumulative allocated size, to get an idea on how much more memory we would have to allocate and touch.

LinkEntry: +25% on top of 3.0GB
List (fixed length): +10% on top of 1.7 GB
Closures: +25% on top of 877.7+MB (I did not collect cumulative size of all closures, just the top 10) 
List (growable): +25% on top of 447.0MB
[...]

Issue #33796, opened 2018-07-09
Ninja won't find build.ninja


MicroDroid:
I followed exactly what [the wiki](https://github.com/dart-lang/sdk/wiki/Building) says to build the SDK, but building fails:

```
$ gclient sync
... # A lot of syncing here
Running hooks: 100% (14/14), done.
$ cd sdk/
$ ./tools/build.py --mode release --arch x64 create_sdk
ninja -C out/ReleaseX64 create_sdk
ninja: error: loading 'build.ninja': No such file or directory
BUILD FAILED
$ ls out/ReleaseX64
args.gn build.ninja build.ninja.d dart-sdk obj toolchain.ninja
```

- I'm building against the `master` branch, and on Arch Linux with latest packages of everything (gcc, depot_tools, etc.) and depot_tools is within my PATH already.
- Running `ninja` command above manually gives same results. However, doing a `cd` to `out/ReleaseX64` and running `ninja -C . create_sdk` manually there makes it start compiling things. That leads me think that... **`ninja` is ignoring the `-C` parameter?**
- I tried logging the `os.getcwd()` in `./tools/build.py` manually right before opening the process to run the `ninja` command (line 253), and it logs `dart-sdk/sdk` directory. Which seems pretty fine, as well.
- File permissions are fine.
- 
  ```
  $ which ninja
  /opt/depot_tools/ninja
  ```

MicroDroid:
Okay apparently using `/opt/depot_tools/ninja` is what causes the error. Use the `ninja` bundled with the project makes it work.

I'll keep this open because the doc has to be clear on this, as well as this needs a workaround because I believe many folks would have depot_tools in their path anyways.

GabrielBB:
I'm having this issue on Windows 10. I have depot_tools in my path. Ninja says the same error. I don't know where to find `/usr/bin/ninja` on Windows

icatalud:
I had the same issue, the problem was error on some steps during the build on windows. The output (defaults to "sdk/out") folder is created anyways with some files, but its content is incomplete. Whenever there is an error, the build.ninja file is not created.

I suggest to anyone having this error to delete the output folder and build again. Once the build finishes check if there were any errors scrolling up in the console.

In my case I'm building on Windows 10 and the problem was some missing Windows Development Kit features (debugger) not installed by Visual Studio.

To get Visual Studio ready (from scratch) for building the SDK (I believe only 2017 and 2019 are compatible):
[...]

PR #39038, opened 2019-10-22
Fast GetHash implementation for 32 bit architectures

icatalud:
#24206 

This is an outline of the hashCode implementation for 32 bit platforms. I have not even tried to compile. 

- Please check if the logic suits the proposed idea.
- There is a lot of code where the size is retrieved from the object's InstanceSize method. This changes are not compatible with that method and to make it so it would require changing them one by one or to standardize the size retrieval to one method (HeapSize).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,5 +104,6 @@ editor/util/testing/mac/Samples.suite/Results
 /generated/
 /crash_logs/
 /build/config/gclient_args.gni
+/runtime/vm_icatalud
 
 runtime/vm_icatalud/
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,5 @@ editor/util/testing/mac/Samples.suite/Results
 /generated/
 /crash_logs/
 /build/config/gclient_args.gni
+
+runtime/vm_icatalud/
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -40,16 +40,32 @@ DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
 #if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
 #elif defined(FAST_HASH_FOR_32_BIT)
-  RawObject* raw = arguments->NativeArgAt(0);
-  if (raw->HasTrailingHashCode()) {
-    return *reinterpret_cast<RawSmi**>(RawObject::LastPointerAddr(raw));
+ /* RawObject* raw = arguments->NativeArgAt(0);
+  if (raw->HasAppendedHashObject()) {
+    return *reinterpret_cast<RawSmi**>(RawObject::AppendedHashAddress(raw));
   }
   uintptr_t value = (reinterpret_cast<uintptr_t>(raw) << 1) >> 1;
   if (!raw->HashCodeWasRetrieved()) {
     raw->SetHashCodeRetrievedBit();
   }
-  return Smi::New(static_cast<uint32_t>(value));
-  //return Smi::New(arguments->NativeArgAt(0)->GetHash());
+  return Smi::New(static_cast<uint32_t>(value));*/
+  RawObject* raw = arguments->NativeArgAt(0);
+  intptr_t cid = raw->GetClassId();
+  if (true || cid > kInstanceCid || cid < kInstanceCid || cid==kInstanceCid ) { // kContextCid, kContextScopeCid
+    if (cid == kInstanceCid) {
+      RawClass* clazz_ptr = Isolate::Current()->class_table()->At(cid);
+      Class& clazz = Class::Handle(clazz_ptr);
+      //OS::Print("Object_getHash for object %p of class %s, class id: %d, HasAppendedHashCode: %d, hash value: 0x%X\n",
+      //          RawObject::ToAddr(raw), clazz.ToCString(), cid, raw->HasAppendedHashObject(), raw->GetHash());
+    }
+    //OS::Print("Getting hash for class %d\n", raw->GetClassId());
+    return Smi::New(arguments->NativeArgAt(0)->GetHash());
+  } else {
+    Heap* heap = isolate->heap();
+    ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
+    return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
+  }
+  
 #else
   Heap* heap = isolate->heap();
   ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -40,32 +40,7 @@ DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
 #if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
 #elif defined(FAST_HASH_FOR_32_BIT)
- /* RawObject* raw = arguments->NativeArgAt(0);
-  if (raw->HasAppendedHashObject()) {
-    return *reinterpret_cast<RawSmi**>(RawObject::AppendedHashAddress(raw));
-  }
-  uintptr_t value = (reinterpret_cast<uintptr_t>(raw) << 1) >> 1;
-  if (!raw->HashCodeWasRetrieved()) {
-    raw->SetHashCodeRetrievedBit();
-  }
-  return Smi::New(static_cast<uint32_t>(value));*/
-  RawObject* raw = arguments->NativeArgAt(0);
-  intptr_t cid = raw->GetClassId();
-  if (true || cid > kInstanceCid || cid < kInstanceCid || cid==kInstanceCid ) { // kContextCid, kContextScopeCid
-    if (cid == kInstanceCid) {
-      RawClass* clazz_ptr = Isolate::Current()->class_table()->At(cid);
-      Class& clazz = Class::Handle(clazz_ptr);
-      //OS::Print("Object_getHash for object %p of class %s, class id: %d, HasAppendedHashCode: %d, hash value: 0x%X\n",
-      //          RawObject::ToAddr(raw), clazz.ToCString(), cid, raw->HasAppendedHashObject(), raw->GetHash());
-    }
-    //OS::Print("Getting hash for class %d\n", raw->GetClassId());
-    return Smi::New(arguments->NativeArgAt(0)->GetHash());
-  } else {
-    Heap* heap = isolate->heap();
-    ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
-    return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
-  }
-  
+  return Smi::New(arguments->NativeArgAt(0)->GetHash());
 #else
   Heap* heap = isolate->heap();
   ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -32,17 +32,17 @@ DEFINE_NATIVE_ENTRY(Object_equals, 0, 1) {
 }
 
 DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
-// Please note that no handle is created for the argument.
-// This is safe since the argument is only used in a tail call.
-// The performance benefit is more than 5% when using hashCode.
-#if defined(HASH_IN_OBJECT_HEADER)
+  // Please note that no handle is created for the argument.
+  // This is safe since the argument is only used in a tail call.
+  // The performance benefit is more than 5% when using hashCode.
+  // #if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
-#else  // 32bit architectures
+  // #else  // 32bit architectures
   // Heap* heap = isolate->heap();
   // ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
   // return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
-  return Smi::New(Object::GetHash(arguments->NativeArgAt(0)));
-#endif
+  //   return Smi::New(Object::GetHash(arguments->NativeArgAt(0)));
+  // #endif
 }
 
 DEFINE_NATIVE_ENTRY(Object_setHash, 0, 2) {
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -37,10 +37,11 @@ DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
 // The performance benefit is more than 5% when using hashCode.
 #if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
-#else
-  Heap* heap = isolate->heap();
-  ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
-  return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
+#else  // 32bit architectures
+  // Heap* heap = isolate->heap();
+  // ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
+  // return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
+  return Smi::New(Object::GetHash(arguments->NativeArgAt(0)));
 #endif
 }
 
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -13,6 +13,8 @@
 #include "vm/stack_frame.h"
 #include "vm/symbols.h"
 
+//#include <easy/profiler.h>  // TODO: testing temporal remove
+
 namespace dart {
 
 DEFINE_NATIVE_ENTRY(DartAsync_fatal, 0, 1) {
@@ -35,21 +37,24 @@ DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
   // Please note that no handle is created for the argument.
   // This is safe since the argument is only used in a tail call.
   // The performance benefit is more than 5% when using hashCode.
-//#if defined(HASH_IN_OBJECT_HEADER)
+#if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
+#elif defined(FAST_HASH_FOR_32_BIT)
+  RawObject* raw = arguments->NativeArgAt(0);
+  if (raw->HasTrailingHashCode()) {
+    return *reinterpret_cast<RawSmi**>(RawObject::LastPointerAddr(raw));
+  }
+  uintptr_t value = (reinterpret_cast<uintptr_t>(raw) << 1) >> 1;
+  if (!raw->HashCodeWasRetrieved()) {
+    raw->SetHashCodeRetrievedBit();
+  }
+  return Smi::New(static_cast<uint32_t>(value));
   //return Smi::New(arguments->NativeArgAt(0)->GetHash());
-  //#else  // 32bit architectures
-  //intptr_t cid = arguments->NativeArgAt(0)->GetClassId();
-  //OS::Print("Retrieving hash for cid %d\n", cid);
-  //if (cid > 0) { //|| cid == kInstanceCid
-  //  //OS::Print("Cached hash for cid %d\n", cid);
-  //  return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
-  //}
-  //return 0;
-   //Heap* heap = isolate->heap();
-   //ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
-   //return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
-//#endif
+#else
+  Heap* heap = isolate->heap();
+  ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
+  return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
+#endif
 }
 
 DEFINE_NATIVE_ENTRY(Object_setHash, 0, 2) {
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -35,14 +35,20 @@ DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
   // Please note that no handle is created for the argument.
   // This is safe since the argument is only used in a tail call.
   // The performance benefit is more than 5% when using hashCode.
-  // #if defined(HASH_IN_OBJECT_HEADER)
+//#if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
-  // #else  // 32bit architectures
-  // Heap* heap = isolate->heap();
-  // ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
-  // return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
-  //   return Smi::New(Object::GetHash(arguments->NativeArgAt(0)));
-  // #endif
+//#else  // 32bit architectures
+  //intptr_t cid = arguments->NativeArgAt(0)->GetClassId();
+  //OS::Print("Retrieving hash for cid %d\n", cid);
+  //if (cid > 0) { //|| cid == kInstanceCid
+  //  //OS::Print("Cached hash for cid %d\n", cid);
+  //  return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
+  //}
+  //return 0;
+   //Heap* heap = isolate->heap();
+   //ASSERT(arguments->NativeArgAt(0)->IsDartInstance());
+   //return Smi::New(heap->GetHash(arguments->NativeArgAt(0)));
+//#endif
 }
 
 DEFINE_NATIVE_ENTRY(Object_setHash, 0, 2) {
--- a/runtime/lib/object.cc
+++ b/runtime/lib/object.cc
@@ -37,7 +37,8 @@ DEFINE_NATIVE_ENTRY(Object_getHash, 0, 1) {
   // The performance benefit is more than 5% when using hashCode.
 //#if defined(HASH_IN_OBJECT_HEADER)
   return Smi::New(Object::GetCachedHash(arguments->NativeArgAt(0)));
-//#else  // 32bit architectures
+  //return Smi::New(arguments->NativeArgAt(0)->GetHash());
+  //#else  // 32bit architectures
   //intptr_t cid = arguments->NativeArgAt(0)->GetClassId();
   //OS::Print("Retrieving hash for cid %d\n", cid);
   //if (cid > 0) { //|| cid == kInstanceCid
--- a/runtime/platform/utils.h
+++ b/runtime/platform/utils.h
@@ -108,12 +108,12 @@ class Utils {
   static int CountOneBits64(uint64_t x);
   static int CountOneBits32(uint32_t x);
 
-  static int CountOneBitsWord(uword x) {
-#ifdef ARCH_IS_64_BIT
-    return CountOneBits64(x);
-#else
-    return CountOneBits32(x);
-#endif
+  static int CountOneBitsWord(uint64_t x) {
+    #ifdef ARCH_IS_64_BIT
+        return CountOneBits64(x);
+    #else
+        return CountOneBits32(x);
+    #endif
   }
 
   static int HighestBit(int64_t v);
@@ -341,7 +341,7 @@ class Utils {
     ASSERT(n <= kBitsPerWord);
     if (n == kBitsPerWord) {
       static_assert((sizeof(uword) * kBitsPerByte) == kBitsPerWord,
-                            "Unexpected uword size");
+                    "Unexpected uword size");
       return std::numeric_limits<uword>::max();
     }
     return (1ll << n) - 1;
--- a/runtime/vm/class_id.h
+++ b/runtime/vm/class_id.h
@@ -214,6 +214,11 @@ enum ClassId {
   kVoidCid,
   kNeverCid,
 
+  // Used for objects that are an extension of another object. There should
+  // be no pointers referencing them.
+  // FAST_HAST_FOR_32_BIT uses this id for appended hash objects.
+  kAppendedObjectCid,
+
   kNumPredefinedCids,
 };
 
--- a/runtime/vm/compiler/asm_intrinsifier_ia32.cc
+++ b/runtime/vm/compiler/asm_intrinsifier_ia32.cc
@@ -1888,7 +1888,47 @@ void AsmIntrinsifier::StringBaseSubstringMatches(Assembler* assembler,
 
 void AsmIntrinsifier::Object_getHash(Assembler* assembler,
                                      Label* normal_ir_body) {
+  //__ movq(EAX, Address(ESP, +1 * target::kWordSize));  // Object.
+  ////__ movl(EAX, FieldAddress(EAX, target::String::hash_offset()));
+  //__ SmiTag(EAX);
+  //__ ret();
+
+	 //Write the object address in the registry Check if the HasHashCode bit tag is
+  //    set If it is
+  //    : Substract kWordSize to the register value(
+  //          and replace) return the value where the register points to If
+  //          is not jump Set the value HashCodeRetrieved tag bit Shift
+  //              the register value 1 bit to the left return the register value
+  Label use_address_as_hash;
+  //, compute_address_hash;
+ 
   UNREACHABLE();
+  
+  __ cmpl(EAX, Immediate(0));
+  __ j(EQUAL, &compute_hash, Assembler::kNearJump);
+  __ ret();
+  //__ movl(EAX, FieldAddress(EAX, target::Type::hash_offset()));
+
+	
+  //__ movl(EBX, Address(ESP, +1 * target::kWordSize));
+  //__ movl(EBX, FieldAddress(EBX, +1 * target::kWordSize));
+  __ movl(EAX, Address(ESP, +1 * target::kWordSize))
+  __ movl(EDI, FieldAddress(EAX, target::Object::tags_offset()));
+  __ testl(EDI, Immediate(1 << kTrailingHashCodeBit));
+  __ j(NOT_EQUAL, &use_address_as_hash, Assembler::kNearJump);
+  //__ movl(EAX, FieldAddress(ESP, -1 * target::kWordSize));
+  __ movl(EAX, FieldAddress(EBX, -1 * target::kWordSize));
+  __ ret();
+
+
+  __ Bind(&use_address_as_hash);
+  // Set the the HashCodeRetrievedBit
+  __ orl(EDI, Immediate(1 << kHashCodeRetrievedBit));
+  __ movl(FieldAddress(EAX, target::Object::tags_offset()), EDI);
+
+  __ sarl(EAX, Immediate(1));  // shift address 1 bit to the left to convert to RawSmi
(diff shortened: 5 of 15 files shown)


mraleph:
@icatalud Thanks a lot for the contribution! You would need to sign the CLA before we would be able to look at the code. 

Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.

> There is a lot of code where the size is retrieved from the object's InstanceSize method. This changes are not compatible with that method and to make it so it would require changing them one by one or to standardize the size retrieval to one method (HeapSize).

I was thinking that the size in the header can be updated when the hashcode field is added (during the relocation) this way the old way of computing the size would work.

mraleph:
The actual review will happen at https://dart-review.googlesource.com/c/sdk/+/122414

icatalud:
>I was thinking that the size in the header can be updated when the hashcode field is added (during the relocation) this way the old way of computing the size would work.

That's the right way of doing it, it didn't slip through my mind. But that is still not compatible with types that don't read the size from the SizeTag (size is bigger). What do you suggest to deal with those cases? Special consideration in every case? Those types use a specialized InstanceSize.

There are "major" design choices changes that I would make, like standardizing certain common operations. I see among basic operations related to objects that are done in many places:

- Getting heap size
- Getting the size of the VM internals to handle objects (method instead of using sizeof)
- Visiting an object's pointers (there are many versions of this method)
- Visiting objects in a range of memory (with possible premature cancel)

Ideally all of this operations should be done through a unique method. Even better would be to document how and where this should be done and specify (clarifying the differences) if there is more than one method to do same operation. This would make it easier to dig into the code or to produce general changes.

I just saw that the hashCode should be stored in Smi format. I will take another look during the evenings or the weekend, I devoted several days to partially understand the VM-GC internals.

There are different GC approaches that I think would be interesting to experiment, like having a parallel GC that never pauses the threads, assuming Dart pauses considering what I read in general about GCs.

icatalud:
@mraleph , @rmacnak-google (I didn't find a way to reply on Gerrit)
I read the comments, I will address them over the weekend. Most of them are simple, but there is one where I need better understanding of the logic.

As Ryan pointed out, there might be a memory issue because the objects can increase in size. I wonder then, how is the forwarding page selected? Is it safe to assume that if memory was sliding "up" instead of down (increasing memory), then at some point the forwarding page will be itself? In that case it is possible to make an address comparison. If the forwarding address is equal to the address, then the extra size is subtracted since there will be no reallocation. In other words, objects are only reallocated if they are going to slide down, but not up. Would that work?

icatalud:
I addressed the comments, but there is one that I cannot resolve. @rmacnak-google pointed out that there are plans to use the reserved tags for another feature. Please resolve that issue, because if that is decided there is no point in the work I'm doing, we will have to find another workaround. Alternatively you could just permanently add hashCode field to 32 bit paying a small price (only in 32 bit machines) in favor of consistency. It would avoid unexpected performance variations when switching platforms or boost unnoticed performance drains.

Also Ryan, about a question I made about a block changing page, you replied that the last block will not always be completely filled. From what I understand when assigning addresses to a forwarding block in PlanMoveToContiguousSize, that method only assigns address in a page if there is enough free space to cover the whole block size. Could you please clarify how that can happen?

I will address more comments in the code (if there are), check the implementation and attempt to compile over the weekend.

The one issue I still have to resolve, is how to deal with Sizes that are not retrieved from the SizeTag and are retrieved from a method defined in the object itself. As long as that size is not used to iterate over the objects in a certain range of memory it's fine to handle that case just in HeapSizeFromClass.

icatalud:
> It would be good if you could try building it and making it pass all the tests before we start digging deeper into the review.

I will attempt to build on Saturday.

icatalud:
@mraleph I haven't been able to get the dev environment working. When I attempt to build I get the following error:

ninja: error: loading 'build.ninja': The system cannot find the file specified.

I'm working on windows. There is an issue #33796 about this. Do you have any suggestions? I don't see any part of the build chain where this file should get created.

Can you copy the contents of this file so perhaps I can try to add it by hand?

icatalud:
Finally I got it working, never mind. I will post on the issue.

icatalud:
@mraleph I updated the Building wiki file in the following commit. I hope you can update it, I saw that it was you who made the last change on the file. I made a comment about the python section that was added because it actually happened to me that I spent a few hours trying to make the virtualenv work, but it wasn't actually needed.

About the build, currently it seems to pass the first part, however it is failing at:
`build generate_snapshot_bin: phony obj/runtime/bin/generate_snapshot_bin.stamp`

```bash
===== CRASH =====
ExceptionCode=-1073741819, ExceptionFlags=0, ExceptionAddress=00007FF6CC58CA23
version=2.7.0-edge.d2dde9a6b19697f8dc282e1fb4e0826a48ac5405 (Sun Nov 3 23:39:21 2019 -0300) on "windows_x64"
thread=388, isolate=isolate(0000017348F9F4D0)
Stack dump aborted because GetAndValidateThreadStackBounds failed.

[539/691] ACTION //utils/dartdevc:dartdevc_kernel_sdk(//build/toolchain/win:x64)
```

This is going to take me some long time, because I cannot work on this during the week (only a little a bit). I hope you can help me go through this parts faster, because I have a big knowledge gap and it takes me a long time to fill it.

icatalud:
An update of the status of what I have done. Hopefully you already know how to solve some of the current issues that I will point out below. I haven’t pushed code here, as I have the code full of comments and replicated code. It’s not suitable for review, but if you wish to test it, [this is the branch link](https://github.com/icatalud/dart-sdk/tree/32bitHashExperiment).

### Increasing size on Reallocation (working with observations)
Testing in 64bit arch the current reallocation logic that adds the extra trailing hashCode works fine, when either the Scavenger or the Compactor reallocate an object with increased size. However it is implemented so that objects are **always increased in size** when they are reallocated if they didn’t already have the trailing extra size. The reason for this is that I’m not sure about the concurrency of the GC. If the scavenger and/or the compactor run concurrently, then the hashcode could be accessed for the first time in between the Scavenger allocating the size and reallocating the object or when the compactor planned but has not slid yet. At least I read some comments in the code that mention “parallel compaction” and the Scavenger I’m pretty sure is concurrent. Questions here:
- Does the Scavenger run concurrently?
- Does the Compactor run concurrently? If the answer is yes, how can it be when objects are being moved (I didn’t see any code that alerts that an object is getting moved)?
- If there is concurrency, what do you suggest as a workaround? Is it good enough to add trailing hash to all objects that are reallocated? (only medium-long lived objects would get the extra size).  This is highly questionable though, since it could lead to the implementation to take even more space than always adding the hashCode field, because in half of the cases this field takes no extra heap size (when UnalignedSizeOfClass%kObjectAlignment == kWordSize).

I wonder about the extra trailing hash taking extra size in 32bit and not in 64bit. In 64bit architecture if class A has only one field uint32 defined and class B extends class A and adds another field uint32:
- Will both classes take 1 unit of alignment (two units when they are stored in the heap)?
- If previous answer is yes, wouldn't that mean that some Raw classes are taking an extra aligment unit in the heap because raw objects have a uint32 hash_ field?
- If previous answer is yes, wouldn't that mean that adding the extra uint32 hash_ in 64bit arch has on average the same cost that in 32bit arch? More precisely the expected (average) object's size increase in units of alignment of adding a 32bit hash is:
  - In 64 bit arch the field increases its size by one extra alignment (128bit) one quarter of the time and 128/4 = 32.
  - In 32 bit arch the field increases its size by one extra alignment (64bit) one half of the time and 64/2 = 32.

### Address as HashCode (not working)
Using the address as the default hashCode on the other hand is currently unviable. It fails even in 64bit when storing this default hashCode in the hash_ field. This can be tested by uncommenting [line 219](https://github.com/icatalud/dart-sdk/blob/68162022aa4b97d63ee6363c0a76f4c74a314b74/runtime/vm/raw_object.h#L219) in raw_object.h. The error thrown while running the build script is at [415/502] ACTION //runtime/bin:generate_snapshot_bin(//build/toolchain/win:x64)
> ...\dart-sdk\sdk\runtime\vm\object.cc: 3716: error: Unable to use class Library:'dart:core' Class: _List@0047665 which is not loaded yet.

The number after _List@ is actually the address-hashCode, so it seems the hash is defined somewhere else and it has to match a certain value. Over the next weekend I can investigate this issue, however I would like to know first:
- Do you know what is causing this and how to fix it?
- If not, do you know someone that might already know why this is happening and could point to the code that causes this? This could save me a significant amount of time.

### Dart wiki question
I the previous post I made a comment about the Building Dart instructions:
> I updated the Building wiki file in the following commit. I hope you can update it, I saw that it was you who made the last change on the file. I made a comment about the python section that was added because it actually happened to me that I spent a few hours trying to make the virtualenv work, but it wasn't actually needed.

- Do you think the Windows instructions should be added? If not, why not?
- Do you think the python section should left as it is (understanding that it could lead to confusion)

### Final consideration
Please @mraleph take some time to reply to all of the questions. If you don’t know or do not wish to reply, please be explicit about it. It annoys me when people reply some of the questions because when someone asks something it is because they want a reply. At least from my side I always make the effort to reply to everything they ask me directly for respect to the other person. When there is no explicit intention on why the question was not replied, people have to start guessing, it could be either because the person didn’t know, they skipped it by accident, it takes too much time to reply, they don’t want to reply, etc, which leads to using more time (take into consideration the whole cycle time of waiting for a reply) to ask again to make sure there is no intention to reply to certain questions (and yet there is still uncertainty).
I understand that replying takes some time and sometimes we can be short on it, but please bear in mind that I have devoted long hours (entire days of my free time to be precise) to understand the code and test the solution. The least I expect in return of my effort is that you devote some time to write a deferential answer.
As an idea to unify expectations of external developers that collaborate to the dart-sdk, perhaps it would be best to create some kind of “external collaboration code” that standardizes how the external collaboration should be addressed. A reason for this is that for example an excited developer could expect replies within the same day, while a dart developer could think that replying once a month is good enough.

mraleph:
Hi @icatalud,

Sorry we are not always very responsive - but as I indicated before we are very busyx. This also unfortunately means you are more or less on your own debugging the code, we can't really invest too much time walking you through the code. 

> Does the Scavenger run concurrently? / Does the Compactor run concurrently?

The answer to this question is "no". Scavenger stops the mutator. Compactor also stops the mutator - but it can run in multiple threads in parallel (with itself but not with mutator). 

Note: in the context of GCs "concurrent" usually means "concurrent with the mutator", while "parallel" just means "parallelised", e.g. multiple threads doing the same kind of work. 

> I wonder about the extra trailing hash taking extra size in 32bit and not in 64bit.

I am not sure I understand this question. Note that you don't need to implement any of this logic on 64-bit. On 64-bit there is already space in the header which is used for  storing the hash. Look for the code wrapped in `HASH_IN_OBJECT_HEADER` to see the logic - and you don't need to change any of that logic. You only need to change code when `HASH_IN_OBJECT_HEADER` is not defined. 

> Do you know what is causing this and how to fix it?

Not really. Most likely some hashes are going out of sync, somehow. Note for example that string has its own hashing rules and you should not redefine `GetHash` in a way that interferes with it. 

I don't think anybody can tell you exactly what is going on without debugging the issue. 

> I the previous post I made a comment about the Building Dart instructions:

I have updated the Building Dart instructions to mention different approach on Windows.

icatalud:
@mraleph 

I start with the subject that bothers me from our interaction:

> Do you think the Windows instructions should be added? If not, why not?

The most important part that I thought should be included was the visual studio install instructions, which you can see on [this commit ](https://github.com/icatalud/dart-sdk-wiki/commit/f2a41916f66b037c610c35013f9092f660ab39d2). This is the solution to issue #33796 so I believe it can be considered relevant and requires minimal effort to fix.
This is the third time I’m asking to include that explanation with the intention to help Dart and its community and I think that not carefully reading the comment (if that is the case) any of those times, even though in one of those I mentioned you directly (as busy as you could be) is borderline disrespectful.

> @mraleph I updated the Building wiki file in the following commit. I hope you can update it, I saw that it was you who made the last change on the file.

This is the last time I ask for this (I’m annoyed about it) so I expect that either you include it in the file or you give an explicit rational explanation of why you think this should not be included.

### Other comments

Thanks for your input, I can complete some of the original questions:

> I am not sure I understand this question.

To make it crystal clear imagine you have this:

```c++
class A {
  uint32_t a;
}

class B extends A {
  uint32_t b;
}
```

I’m not sure about the consequence, but I’m assuming that sizeof(B) will be 64bit because it has two uint32_t fields. Assuming that statement is true:

What I said is that adding the hash_ field in 64 bit machines is also increasing the size of object allocations in one quarter of the cases on average. For example:

```c++
class RawComplexObject extends RawObject {
  uintptr_t a;   // +64 bit
  uint32_t b;   // +32 bit
}
```

In 64bit architecture, because the size of RawObject is 64 bit, the size of RawComplexObject would be 160 bits, so RawComplexObject takes two alignments in the heap when it could actually take 1 alignment if RawObject had no hash_ field. This means that on average having the extra 32 bit field for the hash_ field is taking 1 extra alignment (128 bits) one quarter of the time, so it is taking 32 bits on average, which is the same extra space that would take on average adding the field on 32bit architectures.
If this whole logic is true, then it is very questionable to not include the hash_ field in 32 bit architectures, since it has the same average memory impact that on 64bit architectures.

> Note that you don't need to implement any of this logic on 64-bit. On 64-bit there is already space in the header which is used for storing the hash. Look for the code wrapped in HASH_IN_OBJECT_HEADER to see the logic - and you don't need to change any of that logic. You only need to change code when HASH_IN_OBJECT_HEADER is not defined.

I implemented the logic on 64bit just to test the whole reallocation and tags logic, because in 64bit arch the extra hash_ field makes it much easier to test the code by reading or setting arbitrary values in the hash_ field. In 32bit I'm stuck with the address-hash (which doesn't work).

mraleph:
> The most important part that I thought should be included was the visual studio install instructions

Note that all of this is already included [here](https://chromium.googlesource.com/chromium/src/+/master/docs/windows_build_instructions.md#install). 

Also if you don't install "Debugging Tools" and you run run `build.py` you get an error telling you to install them. 

So for all intents and purposes this is already reasonably documented.

Nevertheless I also included this into the documentation.  

> If this whole logic is true, then it is very questionable to not include the hash_ field in 32 bit architectures, since it has the same average memory impact that on 64bit architectures.

Sorry, I still don't understand your line of reasoning here. On 64-bit machines we have RawObject defined line this: 

```
class RawObject {
  uint32_t tags_;
  uint32_t hash_;  
};
```

So the `hash_` field is "hiding" in the 64-bit header of every object. So it is not really increasing size of the object at all. 

On 32-bit platform we have a 32-bit header so the `tags_` part of it takes it all and we don't have space to hide 32bit hash. 

You are right to ask the question if this added field can on average be hidden within alignment (i.e. if majority of objects have even numbers of pointer size fields then there would be wasted pointer size field in those objects, because we align object size by two words, thus we could use this wasted space to store the hash).  

However some easy experiments demonstrate that this is not the case. You can simply add a field to `RawObject` on IA32 and then run some programs. When I try that I see 10-25% growth in memory consumption (depending on the workload I run) and  sometimes as much as 50% more GC triggered (e.g. compiling hello world with dart2js triggers 12 GCs with default build and 18 GCs with added field).

icatalud:
> Note that all of this is already included here.

Thank you for adding that and it would be more helpful to be more explicit and at least recognize a careless attitude from your side when addressing my previous comments. By not being explicit, with your answer it is unclear if the Windows section was just included on response to my last comment, or if it was added long ago and it was my mistake not reading properly. Everyone makes mistakes and can sometimes skip or misread, but we should recognize them when me make them and that's how we start changing bad habits (but they should be recognized regardless of the attitude being a bad habit or not). Generally I think it is a good habit to be as explicit as possible to avoid ambiguities. 

I think that change is a fix to bug #33796, the issue should probably be closed or at least linked.

> So the hash_ field is "hiding" in the 64-bit header of every object. So it is not really increasing size of the object at all.

But this is only true when talking about the base RawObject class, there many classes that extend RawObject, RawString, RawClass, etc, that add more fields and are instantiated as those classes in the VM.

> However some easy experiments demonstrate that this is not the case. You can simply add a field to RawObject on IA32 and then run some programs.

But what do experiments say if you remove hash_ field from 64 bit machine? Supose it has a similar impact on 64 bit (it should also reduce memory usage).

icatalud:
@mraleph 

> Note that all of this is already included here.

I think this is an excellent example of how lack of explicitness leads to misunderstandings. I followed the link you put there but it was the same link from depot_tools that I have always seen so I didn't see anything different, believing it was a mistake from you. Later after checking the Building.md document, I followed the instructions and opened the depot tools link again and by chance I scrolled up and saw the Windows instructions, soy I added the Update section in my previous comment. The root of the problem is that the instructions said "depot_tools install instructions", not Visual Studio, so most people from windows are just following the instructions that are written there having already installed a plain version of Visual Studio. Even though the information is there, people was not reaching it, because it was not directly visible, proof of it is that there is a bug about it that has some drag. The correct way instructions should be written is install Visual Studio + depot_tools from the chromium instructions anchored at the Visual Studio section and not depot_tools.

I apologize to you for being incisive, however I must say part of it is your fault (maybe you deserve my annoyance) for not being completely explicit when replying my questions from long time ago. If you already knew that this was clearly stated over there and that is the reason you were not adding what I said you should have said it immediately and also resolve the bug.

mraleph:
> That change is a fix to bug #33796, the issue should probably be closed or at least linked.

Note that #33796 has nothing to do with Windows. The original reporter was building on Arch Linux - and furthermore errors you get when incompletely installing Visual Studio look completely different. 

> But this is only true when talking about the base RawObject class, there many classes that extend RawObject, RawString, RawClass, etc, that add more fields and are instantiated as those classes in the VM
> But what do experiments say if you remove hash_ field from 64 bit machine? Supose it has a similar impact on 64 bit (it should also reduce memory usage).

Vast majority of objects consist of pointer size fields (with very few exceptions) and there is an expectation that these fields are aligned by pointer size. So removing 32-bit field from the header (i.e. making RawObject 32-bit in size) is not going to work the way you think. Instead you will just end up with a 32-bits of padding between the header and the rest of the object. That's exactly why `hash_` field is seating there on 64-bit architectures - it is for all intents and purposes free space.

icatalud:
> Note that #33796 has nothing to do with Windows. The original reporter was building on Arch Linux - and furthermore errors you get when incompletely installing Visual Studio look completely different.

You are right, I got confused because the last post was about windows (I arrived there through search), but once again why didn't you clarify this like 4 posts ago? Perhaps the culture of staying silent tries to express that one should "pay" for mistakes. For the sake of effective communication, please be expressive and immediately point out when you see there is a misunderstanding. The consequence is that same thing keeps dragging because it was not resolved from one side. Why would I assume that you didn't address the comment because I was wrong about something?
If you think otherwise let me know, because it's actually a cultural thing (I have that expectation and perhaps you have the opposite one so we should come to a common ground of expectations).

icatalud:
@mraleph 

The pushed code is now working with the address as hash in 32 bit platform.

- New solution is not effective, GetHash is as slow as the previous hashtables.
- I will debug next weekend, but if you already have an idea point it out
- I you can tell me how to include a library to the project (step by step) that would be much appreciated, so I can use an instrumentation profiling library. A profiling over the executable throws that most of the execution time is taken by
SleepConditionVariableSRW	1.45s	1.45s	86.39%	86.39%	KERNELBASE	[unknown]

(in case that sounds familiar to you)

mraleph:
@icatalud awesome! 

> New solution is not effective, GetHash is as slow as the previous hashtables.
> I will debug next weekend, but if you already have an idea point it out

You need to make the final step and provide intrinsic implementation of `_getHash` on 32-bit platforms. Search for `AsmIntrinsifier::Object_getHash`.

> I you can tell me how to include a library to the project (step by step) that would be much appreciated, so I can use an instrumentation profiling library.

I think this really depends on the kind of library. If you have its sources then you need to write GN BUILD files for it... TBH I am not very good in GN so I can't help you much (I usually do it myself with trial and error by copy&pasting some code around).

Note though that unless this library supports JIT compiled code it would not be of much use. 

We used to have vtune support (which I think is the best profiler available on Windows - plus it has JIT support), but we [deleted](https://codereview.chromium.org/1711163002) it some time ago (because we mostly develop on Linux). If you really want to profile - I'd recommend restoring that and using VTune. 
 
> A profiling over the executable when running a Set and hashCode benchmark throws that most of the execution time is taken by SleepConditionVariableSRW 

This looks a bit suspicious.  I have not seen such stuff - maybe profiler is "focusing" on the wrong threads ( on helper threads which are mostly sleeping) instead of focusing on the main thread that is running?

icatalud:
> You need to make the final step and provide intrinsic implementation of _getHash on 32-bit platforms. Search for AsmIntrinsifier::Object_getHash.

Do you think adding that could boost performance of the function by three or four times? If that is the case, profiling is unnecessary. And if that is so important why was it not used on the previous code with the hash table?

mraleph:
> Do you think adding that could boost performance of the function by three or four times?

Yes, easily. 

> And if that is so important why was it not used on the previous code with the hash table?

Because a) it would require writing a lot of assembly code by hand b) it would not actually resolve all the issues associated with hash table based implementation c) as explained before we don't really consider this issue high-priority. 

Note that this is exactly what [this](https://github.com/dart-lang/sdk/issues/24206#issuecomment-344248313) comment says we could do (the second item on the list). 

> Is there any explanation about those instructions?

I am not sure which instructions you refer to? If you mean assembly instructions used by `AsmIntrinsifier` then you would have to look up things in IA32 Programming Manual (for IA32) and in ARM programming manual for ARM.

mraleph:
@icatalud note that code still does not seem to build (I guess Visual Studio has less strict warnings around unused variables).

icatalud:
@mraleph

Correct me if I'm wrong. My current understanding is that the Intrinsifier is overriding the function Object_getHash for another one that executes direct assembly instructions. The modified function I wrote is not going to be used once the Intrinsifier function is created.

If the previous is correct there is a major flaw in the solution, since it's quite complex to write the GetHash function in assembly instructions. The object tags_ need to be checked and then retrieve the trailing hash if there is one. That part is a major obstacle as it would require writing the RawObject::HeapSize() function which we can assume is not reasonable (it would be a titanic task, even intrinsifying the weak table hash function it’s easier). The alternative would be retrieving the trailing hash from assembly only when the tag size actually measures the size of the object. It's still a bit complex but reasonable and inconsistent (there would still be a "surprise behavior" on big objects).

Under those circumstances it's easier to store the hash not trailing but as a new RawObject just before the object and modify the VisitPointers function to also visit this object (or whichever function is used by the GC to visit objects). The trailing hash is using 1 alignment so it's possible to transform the trailing hash into a RawObject. Then it would be easy to read the hash in assembly.
The previous has actually other advantages like for example the tags_size and HeapSize() function remain intact (only the function visit pointers used by the GC is adapted).

The comment you linked points to this exact solution as inconvenient because of all of the added complexity (true), are you inclined towards adding something like this despite that?

mraleph:
> My current understanding is that the Intrinsifier is overriding the function Object_getHash for another one that executes direct assembly instructions. The modified function I wrote is not going to be used once the Intrinsifier function is created.

Intrinsic version forms a prefix of the full version - intrinsic is a fast path which can fall through to a slow path if it can't handle something. Whether or not intrinsic ever *fallthrough* to the normal implementation is decided on the case by case basis when writing an intrinsic. 

> That part is a major obstacle as it would require writing the RawObject::HeapSize() function which we can assume is not reasonable

Not really, the implementation would be few dozen assembly instructions - important thing here is that you only can encounter few different classes of objects in the intrinsic, and only few classes of variable size objects (normal fixed size list, typed data lists and strings), so you only need to support a subset of heap size computation that runtime supports.

icatalud:
> Not really, the implementation would be few dozen assembly instructions

In my current reality this is out of what I'm willing-capable of doing.

The only way I'm willing to finish the implementation is using the code path I pointed out previously, then the assembly instructions should be 10 or less. Please evaluate the proposal and think if there are any inconsistencies:
>Under those circumstances it's easier to store the hash not trailing but as a new RawObject just before the object and modify the VisitPointers function to also visit this object (or whichever function is used by the GC to visit objects). The trailing hash is using 1 alignment so it's possible to transform the trailing hash into a RawObject. Then it would be easy to read the hash in assembly.

mraleph:
> Please evaluate the proposal and think if there are any inconsistencies

I think the solution with a "prefix" object can be made to work. I actually start to kinda like it - once I think about it longer. You don't really need to visit that object at all (because it does not contain any pointers), all you need to do is to ensure that it survives major GCs, which you can easily do by altering marking to mark it whenever it marks an object in a "hash-taken-and-relocated" state. 

This way the hash will also be considerably faster to access (considerably less branching).  

It would be a bit more wasteful in terms of memory (you always need 2 words instead of potentially 0 due to 2 words alignment of object size), but it would still be better than the current situation.

icatalud:
> It would be a bit more wasteful in terms of memory (you always need 2 words instead of potentially 0 due to 2 words alignment of object size), but it would still be better than the current situation.

But with the current solution 2 words were always being appended, because the code was not evaluating whether the last word in the object was free space or not, it was plainly always adding the 2 words. In terms of memory is the same.

The one concern I have is the Scavenger reallocating the appended hash object separately, but it's probably easy to considerate that as a special case.

mraleph:
> The one concern I have is the Scavenger reallocating the appended hash object separately, but it's probably easy to considerate that as a special case.

Note that scavenger would not relocate appended hash object because it is not referenced from anywhere.

icatalud:
> Note that scavenger would not relocate appended hash object because it is not referenced from anywhere.

That's good news, that's why asking sometimes can be the best shortcut. I will attempt that approach. Thanks.

icatalud:
@mraleph The build never run on the last commit.

I had problems with the hash object solution. Could you confirm with someone else if it is safe to have a lose object in the heap and which object class id would be appropriate for such object? For example there could be a place in the heap where VM base objects are stored (classes, etc) and it is assumed that the object right after object A must be B, but instead it is a hash object.
For ClassId I have tested with kFreeListElement or kForwardingCorpse, is any of those ok? I'm also not marking the hash object, could not marking the object produce problems somewhere else or unmarked objects are only ever freed by the compactor? 

The second thing is that writing the assembly instructions would take me considerable of amount of time, but if you are proficient with the Intrensifier I'm pretty sure you could write the instructions in a few minutes. The exact code-assembly instructions with placeholders would be of great help. What needs to be done is the following:
- Write the object address in the registry
- Check if the HasHashCode bit tag is set
- If it is:
  - Substract kWordSize to the register value (and replace)
  - return the value where the register points to
- If is not jump
  - Set the value HashCodeRetrieved tag bit
  - Shift the register value 1 bit to the left
  - return the register value

mraleph:
> For example there could be a place in the heap where VM base objects are stored (classes, etc) and it is assumed that the object right after object A must be B, but instead it is a hash object.

There are currently no such assumptions. Also note that you should not be really be changing anything  around how base entities are stored. As I have highlighted before you should be focusing on those classes that user level Dart code can see (most notably strings, arrays and user defined classes which have cids above `kInstanceCid`).

> For ClassId I have tested with kFreeListElement or kForwardingCorpse, is any of those ok? 

I would not use any of those. As a starting point I would use something like `RawMint` object (this would waste 2 words of space - but should be good enough for prototyping). Ideally you could create your own class `RawHashWrapper` or something like that - if you want to squeeze size out of the heap. 

You also need to be careful around how you write and read objects from snapshots, because you need to preserve and restore the hash "carrier" object.

> I'm also not marking the hash object, could not marking the object produce problems somewhere else or unmarked objects are only ever freed by the compactor?

As I have mentioned before (see my older [comment](https://github.com/dart-lang/sdk/pull/39038#issuecomment-559250917)) you should most certainly mark this object. Otherwise sweeping would reclaim the space. 

> The second thing is that writing the assembly instructions would take me considerable of amount of time,

As I have highlighted before you are more or less on your own with the code here. You choose for yourself to try and fix this issue - which we do not consider to be a high priority.

icatalud:
> You also need to be careful around how you write and read objects from snapshots, because you need to preserve and restore the hash "carrier" object.

Does that mean that the snapshots do not copy the heap as it is and it's necessary to perform certain checks?

> I would not use any of those. As a starting point I would use something like RawMint object (this would waste 2 words of space - but should be good enough for prototyping).

The problem is that the logic can at most duplicate the object in size, using two alignments instead of one, could triplicate the size. How about using kInstanceCid, is that safe? (otherwise I will just create the RawHashObject)

> As I have mentioned before (see my older comment) you should most certainly mark this object. Otherwise sweeping would reclaim the space.

There is an edge case in the compacting logic which could split the hash object with the object if the hash object and the object are not in the same block (placed there by the scavenger) and the object block is allocated in a new page, that's what I decided to switch the solution to not mark it, so that the compactor would consider the hash object free space (copied anyways later by the object).
I’ll do it the other way around, mark the hash object and treat the split case specially.

> As I have highlighted before you are more or less on your own with the code here. You choose for yourself to try and fix this issue - which we do not consider to be a high priority.

I was only asking in case you were proficient with the Intrensifier and it took you just a few minutes to draft the 5-7 instructions that would be required (particularly the names and inputs of the functions in the Intrensifier) not to actually run them and/or test them. I mentioned that specially considering that this is not high priority and you are not willing to invest an important amount of time on the issue. It is obviously in the best interest of everyone that the issue is solved and if 5 or 10 minutes of your time can save one day of my work (1 day closer to complete the solution), then it can probably be leveraged as worth doing. What slows me down considerable is having to dig in the definition of every function to understand if it is performing exactly the operation that is required. The instructions are not documented and I'm concerned that testing it will be quite challenging, especially if I'm not sure that I'm using the right instruction.

icatalud:
Hi @iposva-google and @mraleph,

I was doing some experiments about adding the hash object and I noticed that creating any new objects in the ScavengePointer function outside from the object being scavenged causes the program to crash during the build.

@iposva-google I pinged you because I see many comments from you in that code. If your are familiar with the scavenger I wonder if you know if it is possible to create an object using the allocated space in the ScavengePointer function. Assume the following is being considered:
- The extra space in being included in the allocation: Allocate...(size + extra_size)
- The new object is being allocated in new_addr and it's size is extra_size
- The scavenged pointer new address is being correctly added to the Promoted stack: scavenger_->PushToPromotedStack(new_addr + extra_size);
- Allocating extra size is not a problem when the extra size is added to the reallocated object, the problem is creating a new object

Appreciate your comments.

icatalud:
Hi @mraleph, I finally got it working. The hashCode speedup is formidable (x20).

Fast hash:
> HashCode for int average time: 1.554243083618278 us
HashCode for Object average time: **5.835472578763127 us**
HashCode for String average time: 3.920815366522932 us
HashCode for A average time: **4.49438202247191 us**
HashCode for Identifiable average time: 5.065856129685917 us

Old: 
> HashCode for int average time: 4.758801141769743 us
HashCode for Object average time: **230.13636363636363 us**
HashCode for String average time: 6.138735420503377 us
HashCode for A average time: **210.79166666666666 us**
HashCode for Identifiable average time: 7.378318584070796 us


The test of adding objects to a Set speeds up around 4 times. At least now the speeds normalized and the bottleneck is where it should be, at storing the object and not at hashCode.

I made quite a few changes, so there might be more unused variables. Let’s see what build says.

About the problems I had, I inserted a bug while I was testing in 64bit and a tag bit was incorrectly being set. That said, it still odd that in 32 bit when adding an appended object to all reallocated objects, the program worked fine, while in 64bit it crashed at the Scavenger unless I removed a subset of class ids that were the majority of objects being reallocated (class id greater than 100). As a pre-clarification, in 64 bit I was simply appending an object to test the logic, the object was not being used for anything.

There is still a long road to finish, a refactoring and tests are missing, so I think it’s taking me at least one more weekend (and more).

This whole problem made me question the consistency of the hashmap operations. If the runtime types are the same and there is a hashCode collision, then a wrong object could be incorrectly placed or retrieved in the unlikely case where the hashes match. With this solution collisions are more likely than with the previous one. Am I correct about this or is there something I’m not considering?
If what I say is correct, the hashCode should be initialized with an increasing global id.

mraleph:
> Does that mean that the snapshots do not copy the heap as it is and it's necessary to perform certain checks?

Snapshots serialise object graph rather than just copy raw memory around (See for example clustered_snapshot.cc). That is why you would need to make some changes to ensure that hash code is serialized and deserialized properly.

> Hi @mraleph, I finally got it working. The hashCode speedup is formidable (x20).

This is starting to look really good! 

> If the runtime types are the same and there is a hashCode collision, then a wrong object could be incorrectly placed or retrieved in the unlikely case where the hashes match.

I am not sure I understand the question: hashmap use hash code plus some equality comparison to retrieve objects, even if there are two keys with the same hash if they are unequal (e.g. by default `==` is used) then hashmap will never confuse them.

icatalud:
I pushed again, try to build please. If you know how to easily check for unused vars and stuff like that let me know.

> That is why you would need to make some changes to ensure that hash code is serialized and deserialized properly.

So although the solution is working, snapshots are broken with this implementation?
I hope you can direct me more on the snapshot part so that I can move forward faster. I truly have devoted far more time than what I had expected.

> I am not sure I understand the question: hashmap use hash code plus some equality comparison to retrieve objects.

My bad, I remember reading that equality was implemented as hashCode equality && runtimetype equality, but that is the suggestion when implementing ==, the default implementation does perform object pointer equality as you noted it. If it was like the suggested implementation, hash-maps would currently be broken and that's what I was trying to explain.

icatalud:
Hi @mraleph ,

I couldn't compile. The build throws an error about Dart extension methods. I'm hoping this is just a Windows thing.

I fixed the comparison error, let's see if it build on Gerrit. Are you waiting for the program to compile to start a review?

icatalud:
Hi @mraleph ,

We should formalize our work cycle, such that we can finalize this issue, otherwise your and mine responses are too unreliable and this demotivates me (probably you too). This is what comes into my mind:

- Address comments or questions once week (this can take longer time).
- Every day when you arrive to your office you trigger a new build to the latest patchset. I'm assuming that triggering a build takes less than 3 minutes, but if that is not the case please clarify and propose an alternative.

The above should be a reasonable time-priority trade off.
We are very close to the finish, I showed you that the solution is working. I really want to end this and take this issue out of my head. After rebasing the other day I haven't been able to build because there is a complaint about an experimental extension methods flag.

If you can trigger a build every day, then I can consistently check every evening what has failed. The compiler complaints are usually trivial changes that I can make in a few minutes.

Let me know what you think.

mraleph:
@icatalud I am away from work for the next month or so. I will try to answer your questions from time to time - but I can't guarantee any regular cadence. 

> So although the solution is working, snapshots are broken with this implementation?

Yeah, I think they are _a bit_ broken. I think it would work if you just clear hash related bits when writing snapshots.

It seems that `clustered_snapshot` implementation is not affected because that writes tags very manually. 

However `snapshot.cc` / `raw_object_snapshot.cc` are affected. Look for `WriteTags` method there. I think it should just clear the hash related bits out. 

> I couldn't compile. The build throws an error about Dart extension methods. I'm hoping this is just a Windows thing.

Did you try doing `gclient sync` after you rebased?

icatalud:
> I am away from work for the next month

Can you pass the responsibility to someone else during that time?

>  I will try to answer your questions from time to time - but I can't guarantee any regular cadence.

Is it possible to trigger a new build on a daily basis? It's just a matter of writing it down in the calendar, like a 3 minute meeting (assuming it's that easy to do).

About the reply cadence, I think it would greatly benefit the success of our joint project that you define a minimal amount of time you will invest once a week or once every two weeks to this issue. One good reason for it, is that it displays commitment and gives importance to the work I’m doing. Another reason is that it helps me create the right expectations and program my schedule better. Although this has low priority, it is still relevant and it deserves a proportional amount of time to be invested. Imagine your are doing external job, wouldn’t you prefer that it was somehow regulated? Everyone that invest their time in something would like to have certain guarantees and to not be left to depend on the mood of the counterpart.

Ideally you can address the issue before the cadence timeline, the timeline serves as a guarantee that you will take a look and invest let’s say 30 min (once a week or every two weeks) regardless of the stack of higher responsibilities that you have at any point in time, just like a meeting you have set up in your calendar.

> However snapshot.cc / raw_object_snapshot.cc are affected. Look for WriteTags method there. I think it should just clear the hash related bits out.

Any script or command you can give me so I can test that?

> Did you try doing gclient sync after you rebased?

I didn't, I tried and it worked. Thanks! Trigger a build please, it should work.

PR #39165, opened 2019-10-29
Fixed Dart->C++ mapping of IPPROTO_IPV6 socket option

rencevio:
Fixes mapping of Dart `RawSocketOption` in `RawSocketOption_GetOptionValue` function.


mraleph:
Thanks for the contribution. CQ is running here https://dart-review.googlesource.com/c/sdk/+/123440

PR #39249, opened 2019-11-05
Modify typedef for catchError callback function

sdstolworthy:


sdstolworthy:
@googlebot I signed it!

sdk/lib/async/future.dart
```diff
@@ -635,7 +635,8 @@ abstract class Future<T> {
   // - (dynamic, StackTrace) -> FutureOr<T>
   // Given that there is a `test` function that is usually used to do an
   // `isCheck` we should also expect functions that take a specific argument.
-  Future<T> catchError(Function onError, {bool test(Object error)});
+  Future<T> catchError(void onError(Object error, [StackTrace stackTrace]),
```
mraleph: This is a breaking change: note that the type you are using here is in complete disagreement with types listed in the comment above. `onError` is supposed to be either `FutureOr<T> Function(dynamic)` or `FutureOr<T> Function(dynamic, StackTrace)` - your type does not capture this (and strictly speaking can't capture this). 

An important thing to realise here is that you can't pass a function of type `FutureOr<T> Function(dynamic)` where a function of type `FutureOr<T> Function(dynamic, [StackTrace])` is expected.

mraleph:
Thank you for your contribution. Unfortunately (as explained above) we can't take it - because it is a breaking change to the dart core libraries. I am going to close this PR.

PR #39328, opened 2019-11-11
Fix typo

mnordine:
I believe it should be `dart_tool`


mit-mit:
Review in https://dart-review.googlesource.com/c/sdk/+/124720

Thanks for contributing! Good catch :-)

PR #39350, opened 2019-11-12
[dart:core] Update DartTime.parse to support arbitrary precision fractional seconds.

### LINKED ISSUE #24205: DateTime.parse() does not strictly include RFC 3339: limited millisecond length
RFC 3339 defines the "time-secfrac" component of a timestamp to be `"." 1*DIGIT`, which permits second fractions of unlimited length. `DateTime.parse()` only accepts a second fraction component of up to length 6.

The length limitation on "millis_opt" should be removed, or the documentation should be updated to note that DateTime.parse() does not strictly support RFC 3339.

slightfoot:
Update DartTime.parse to support arbitrary precision fractional seconds. So it now conforms to ISO8901 and RFC3339 standards.

Closes #24205


slightfoot:
Poke @lrhn I see you modified this code before..

kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/124903

PR #39417, opened 2019-11-17
Support latest analyzer in analyzer_plugin

simolus3:
I've raised the upper constraint for the analyzer to `<0.40.0` so that `0.39.1` can be supported. The latest `build_resolvers` package requires analyzer `^0.39.0`, it would be nice to use those packages together.

### DIFF
--- a/pkg/analyzer_plugin/CHANGELOG.md
+++ b/pkg/analyzer_plugin/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.2.2
+- Bump maximum supported version of the analyzer to `<0.40.0`
+
 ## 0.2.1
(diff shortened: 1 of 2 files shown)


mit-mit:
Code review: https://dart-review.googlesource.com/c/sdk/+/125491

pdf:
This was closed by the merge a month ago, but the updated package was never published to pub.dev

simolus3:
@pdf If you depend on it directly, you can use  https://pub.dev/packages/analyzer_plugin_fork which is the exact same package with a different name and analyzer 0.39 support. It's still bad if `analyzer_plugin` is a transitive dependency you can't change...

PR #39453, opened 2019-11-19
Fix syntax error in Symbol doc example

mtcliatt:


mit-mit:
Thanks for contributing, this is https://dart-review.googlesource.com/c/sdk/+/125722

PR #39462, opened 2019-11-20
Hide dart:cli in api docs

mit-mit:
This is experimental, and not ready for general consumption, so we shouldn't be advertizing it


mit-mit:
Review in https://dart-review.googlesource.com/c/sdk/+/125800

PR #39479, opened 2019-11-21
Update math.dart:atan2 in relation to atan comment

dimpen:
It's a **typo**. It should say **atan(a/b).** 
The way the _atan2(a,b)_ is defined _x=b, y=a => atan2(y,x) = atan(y/x) if x is positive_. 

So _atan2(a,b) = atan(a/b) if b is positive._

Indeed:
```
import 'dart:math'; 

void main() {
    double a = 1, b = 2;
    assert(atan2(a, b) == atan(a / b));
    assert(atan2(a, b) != atan(b / a));
}
```

### DIFF
--- a/sdk/lib/math/math.dart
+++ b/sdk/lib/math/math.dart
@@ -77,7 +77,7 @@ external T max<T extends num>(T a, T b);
 /// and the vector ([b],[a]).
 /// The result is in the range -PI..PI.
 ///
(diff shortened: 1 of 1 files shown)


dimpen:
@googlebot I signed it!

kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/125926

CC @lrhn

PR #39676, opened 2019-12-06
Update iterator.dart to fully spec moveNext()

jtlapp:
The `moveNext` docs should contain all the information necessary to implement it. There is otherwise no mention that this method is also used to set `current` to the initial element.

### DIFF
--- a/sdk/lib/core/iterator.dart
+++ b/sdk/lib/core/iterator.dart
@@ -33,21 +33,21 @@ part of dart.core;
  * in the [library tour](http://www.dartlang.org/docs/dart-up-and-running/contents/ch03.html)
  */
 abstract class Iterator<E> {
-  /**
-   * Moves to the next element.
+   /**
+   * Advances the iterator to the next element of the iteration.
+   * 
+   * Should be called before reading [current]. It the call to `moveNext` returns
+   * `true`, then [current] will contain the next element of the iteration until 
+   * `moveNext` is called again. If the call returns `false`, there are no further
+   * elements and [current] should not be used any more.
    *
-   * Returns true if [current] contains the next element.
-   * Returns false if no elements are left.
+   * It is safe to call [moveNext] after it has already returned `false`, 
+   * but it must keep returning `false` and not have any other effect.
    *
-   * Invoke [moveNext] to move to the first element and to each
-   * subsequent element.
-   *
-   * It is safe to invoke [moveNext] even when the iterator is already
-   * positioned after the last element.
-   * In this case [moveNext] returns false again and has no effect.
-   *
-   * A call to `moveNext` may throw if iteration has been broken by
-   * changing the underlying collection.
+   * A call to `moveNext` may throw for various reasons, including a concurrent change
+   * to an underlying collection. If that happens, the iterator may be in an inconsistent 
+   * state, and any further behavior of the iterator is unspecified, 
+   * including the effect of reading [current]. 
    */
   bool moveNext();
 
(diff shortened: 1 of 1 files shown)


jtlapp:
@googlebot I signed it!

mit-mit:
Code review in https://dart-review.googlesource.com/c/sdk/+/127455

sdk/lib/core/iterator.dart
```diff
@@ -39,6 +39,9 @@ abstract class Iterator<E> {
    * Returns true if [current] contains the next element.
    * Returns false if no elements are left.
    *
+   * Invoke [moveNext] to move to the first element and to each
+   * subsequent element.
+   *
```
lrhn: This repeats the behavior mentioned in the class description, but it does make it easier to find this information while only having an `Iterator` typed variable available.

I'd prefer to phrase it slightly differently (we rarely use "invoke" in documentation, it's too formal). Maybe:

> Use the [moveNext] method to move to the next element. This will change the
> current value to that of the next element if there is a next element, which is when
> [moveNext] returns `true`. Also use [moveNext] to start the iteration before accessing
> the first element.

lrhn: Hmm, I misread this as a comment on `current`.

I'm not sure it's suitable as an *addition* on `moveNext`, because the documentation does say that already, but the existing text could probably be rewritten to express the same thing more *clearly*.

Maybe:
> ```dart
>  /**
>  * Advances the iterator to the next element of the iteration.
>  * 
>  * Should be called before reading [current]. It the call to `moveNext` returns
>  * `true`, then [current] will contain the next element of the iteration until 
>  * `moveNext` is called again. If the call returns `false`, there are no further
>  * elements and [current] should not be used any more.
>  *
>  * It is safe to call [moveNext] after it has already returned `false`, 
>  * but it must keep returning `false` and not have any other effect.
>  *
>  * A call to `moveNext` may throw for various reasons, including a concurrent change
>  * to an underlying collection. If that happens, the iterator may be in an inconsistent 
>  * state, and any further behavior of the iterator is unspecified, 
>  * including the effect of reading [current]. 
>  */

How does this read to you?

mit-mit:
@jtlapp thanks for your contribution, and sorry the delay. Please see review feedback above.

jtlapp:
I'm having trouble parsing your proposed alternatives. They aren't clear English. The `moveNext` should itself say what it does. Not having it there cost me time implementing it. This doesn't need to be complicated. Substitute "Call" for "Invoke" if you like. It is a concept all programmers know.

lrhn:
If you have trouble parsing it, then it's probably not good documentation :).

The problem with describing what `moveNext` does is that all it does is to change internal state. As such, it's necessary to first describe the internal state (abstractly, because iterators can be implemented in many ways), and then say how calling `moveNext` updates that state.

So, maybe:

> Advances the iterator to the next element of the iteration.
>
> An iterator maintains a position in a sequence of elements being iterated.
>
> The position starts out before the first element, and calls to `moveNext` advances the
> position to the next element of the sequence, as long as there is a next element.
> Returns true if there was a next element and the call advanced the position to that element,
> and returns false if there are no further elements to iterate, and then the position
> is after the last element.
>
> The current elements of the iteration can be accessed as the [current] value *after* 
> a call to [moveNext] has returned true. It is not available before calling `moveNext`
> the first time on an iterator, nor after such a call has returned false.
> 
> It is safe to call `moveNext` after it has already returned `false`, 
>  but it must keep returning `false` and not have any other effect.
>
> A call to `moveNext` may throw for various reasons, including a concurrent change
> to an underlying collection. If that happens, the iterator may be in an inconsistent 
> state, and any further behavior of the iterator is unspecified, 
> including the effect of reading [current].

jtlapp:
That looks fine to me. You put more care into the explanation than I did.

mit-mit:
@jtlapp @lrhn I updated this PR with the docs comment I believe you both agreed to?

PR #39860, opened 2019-12-19
Sets the loading path for tzdata

filmil:
Fuchsia has a fixed path in which tzdata is offered to
running programs.  This change sets the env variables
that point to the data.


mraleph:
Actual review here https://dart-review.googlesource.com/c/sdk/+/129003

filmil:
Fixed at #39861, I think this PR is OK to close now.

Issue #27695, opened 2016-10-30
Negative endIndex support for `String.substring()`


tejainece:
Would be nice if `String.substring()` supported negative endIndex. Then one can do:

```dart
String get something => getAnotherString().substring(1, -1);
```

Rather than:

```dart
String get something {
  final String str = getAnotherString();
  return str.substring(1, str.length-1);
} 
```

or

```dart
String get something => getAnotherString().substring(1, getAnotherString().length-1);
```

lrhn:
Yes and no.
The problem with accepting more things is that it can also more easily hide mistakes.

floitschG:
There is also a question of consistency: if `substring` allows `-1` why not `[]`. If it works for strings, why not for lists (`sublist`) ...

woozbooz:
Would be nice if Dart developers buy yourself a brains ans make end index inclusive.

PR #39898, opened 2019-12-21
Support for String.substring negative indexes

urusaich:
#27695

### DIFF
--- a/sdk/lib/_internal/vm/lib/string_patch.dart
+++ b/sdk/lib/_internal/vm/lib/string_patch.dart
@@ -384,6 +384,13 @@ abstract class _StringBase implements String {
   String substring(int startIndex, [int endIndex]) {
     endIndex ??= this.length;
 
+    if (startIndex < 0) {
+      startIndex = this.length + startIndex;
+    }
+    if (endIndex < 0) {
+      endIndex = this.length + endIndex;
+    }
+
     if ((startIndex < 0) || (startIndex > this.length)) {
       throw new RangeError.value(startIndex);
     }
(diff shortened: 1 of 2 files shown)


nanodesu88:
@googlebot I signed it!

kevmoo:
CC @lrhn

lrhn:
Thank you for your interest.

We don't have any current plans to change the behavior of index operations to allow negative numbers.

I know it's inconvenient to have to write `s.substring(2, s.length - 2)` instead of just `s.substring(2, -2)`. I can see the lure.

However, it's not without cost. If nothing else, every user will now pay for two more `< 0` checks on every call to `substring`, whether they use it or not (which is so far entirely "not"). There is also the risk of hiding errors when a value is accidentally negative. A negative argument is already a hard-to-catch error because it only gets caught at run-time, we do not have a type for "positive integers".

Also, if we do decide to add the ability to use negative indices, we would do it *everywhere*, not just for `substring`. That makes it a large change affecting user-implementable interfaces (like `List` with its `sublist` and `[]` methods), and it would make existing subclasses no longer satisfy the interface contract when they don't accept the negative arguments. So, not a small change.
Doing it only for a single method is not good for consistency, and could lead users to believe it works elsewhere, which doubles the trouble of errors only being caught at run-time.

So, for now, I think we'll prefer to not change `substring`.

nanodesu88:
> Also, if we do decide to add the ability to use negative indices, we would do it _everywhere_, not just for `substring`. That makes it a large change affecting user-implementable interfaces (like `List` with its `sublist` and `[]` methods), and it would make existing subclasses no longer satisfy the interface contract when they don't accept the negative arguments. So, not a small change.
> Doing it only for a single method is not good for consistency, and could lead users to believe it works elsewhere, which doubles the trouble of errors only being caught at run-time.

Good point of view. I think is a good reason for not adding a ability to use negative indeces. Taken

PR #40025, opened 2020-01-08
Analysis server: Highlight null and throw keywords

simolus3:
The `DartUnitHighlightsComputer2` does not highlight the `null` and `throw` keywords. This PR implements the two missing cases and adds a simple test case for them.


simolus3:
I'll just use git cl directly

PR #40431, opened 2020-02-01
Add support for @Unsafe to analyzer

mehmetf:
Add support for `@Unsafe` annotation as `isUnsafe` and `hasUnsafe`.

See: https://github.com/dart-lang/sdk/pull/40429 for the definition of the annotation.

### DIFF
--- a/pkg/analyzer/CHANGELOG.md
+++ b/pkg/analyzer/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.39.6
+* Add support for `@Unsafe` annotation as `isUnsafe` and `hasUnsafe`.
+
 ## 0.39.5-dev
 * Deprecated `ClassElement.instantiateToBounds()` and
   `FunctionTypeAliasElement.instantiateToBounds()`. With the null-safety
--- a/pkg/analyzer/lib/src/dart/element/element.dart
+++ b/pkg/analyzer/lib/src/dart/element/element.dart
@@ -2392,6 +2392,9 @@ class ElementAnnotationImpl implements ElementAnnotation {
   /// The name of the top-level variable used to mark a class as being sealed.
   static const String _SEALED_VARIABLE_NAME = "sealed";
 
+  /// The name of the top-level variable used to mark a method as unsafe.
+  static const String _UNSAFE_VARIABLE_NAME = "Unsafe";
+
   /// The name of the top-level variable used to mark a method as being
   /// visible for templates.
   static const String _VISIBLE_FOR_TEMPLATE_VARIABLE_NAME =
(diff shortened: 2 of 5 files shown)


pkg/analyzer/lib/src/dart/element/element.dart
```diff
@@ -2392,6 +2392,9 @@ class ElementAnnotationImpl implements ElementAnnotation {
   /// The name of the top-level variable used to mark a class as being sealed.
   static const String _SEALED_VARIABLE_NAME = "sealed";
 
+  /// The name of the top-level variable used to mark a method as unsafe.
```
pq: Is this just methods or other members?

pkg/analyzer/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.39.6
```
pq: I *think* this should be 0.39.5?

/cc @stereotype441

pkg/analyzer/lib/src/dart/element/element.dart
bwilkerson: nit: `Unsafe` is a class, not a variable.

pkg/analyzer/lib/src/dart/element/element.dart
```diff
@@ -2540,6 +2543,12 @@ class ElementAnnotationImpl implements ElementAnnotation {
       element.name == _SEALED_VARIABLE_NAME &&
       element.library?.name == _META_LIB_NAME;
 
+  @override
+  bool get isUnsafe =>
+      element is PropertyAccessorElement &&
```
bwilkerson: This seems wrong to me. I would have expected that it would be a `ClassElement`.

bwilkerson (commented):
We obviously need to resolve the question of where the annotation lives before committing this.

pkg/analyzer/lib/src/dart/element/element.dart
mehmetf: Hint: I don't know what I am doing :-) So thanks for all the help!

I have not setup my environment to test this yet. When the annotation is agreed upon and pushed to master, I can tinker with this.

pkg/analyzer/CHANGELOG.md
mehmetf: Yeah, I am not sure what `-dev` means....

pkg/analyzer/lib/src/dart/element/element.dart
mehmetf: Fixed.

mehmetf:
Closing this PR as there's a better proposal being discussed.

PR #40549, opened 2020-02-08
Draft PR for introducing a way to ban http

mehmetf:
See proposal: https://github.com/dart-lang/sdk/issues/40548

Tests and documentation is missing as it is a POC.

### DIFF
--- a/sdk/lib/_http/http_impl.dart
+++ b/sdk/lib/_http/http_impl.dart
@@ -2126,6 +2126,8 @@ class _ConnectionTarget {
 typedef bool BadCertificateCallback(X509Certificate cr, String host, int port);
 
 class _HttpClient implements HttpClient {
+  static const bool _isHttpAllowedByDefault = true;
+
   bool _closing = false;
   bool _closingForcefully = false;
   final Map<String, _ConnectionTarget> _connectionTargets =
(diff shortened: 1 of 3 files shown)


sdk/lib/_http/http_impl.dart
```diff
@@ -2285,6 +2288,11 @@ class _HttpClient implements HttpClient {
     }
 
     bool isSecure = (uri.scheme == "https");
+    bool isMobileClient = Platform.isAndroid || Platform.isIOS;
```
zanderso: If you remove the restriction that the flag only works on mobile platforms, then I suspect it will be easier to write tests. A regular Dart program could just install an HttpOverrides in a zone with the flag flipped, and check that the right exception is thrown.

mehmetf: There are two reason for including a mobile client distinction:
- We intend to flip the default in google3 but only for mobile. The default should be `allow` for non-mobile applications even in google3. Perhaps, this can be solved by having defaults per Platform.
- It reflects the principle of the change more accurately. On iOS and Android, we should not be asking the apps to install an Override to get this behavior. Both OSes ban cleartext HTTP by default for their native apps.

I introduced platform specific defaults to capture both points and removed the `isMobileClient` check.

sdk/lib/_http/http_impl.dart
```diff
@@ -2151,6 +2152,13 @@ class _HttpClient implements HttpClient {
 
   _HttpClient(this._context);
 
+  static bool get _isHttpAllowedByDefault {
+    if (Platform.isIOS) return true;
```
vanesyan: Maybe opposite value, i.e. `false`?

mehmetf: That would be a breaking change upstream.

I would be happy to do it but @mit-mit needs to agree.

mehmetf:
This requires a rewrite. Closing.

PR #40929, opened 2020-03-09
Remove duplicate check for space character (32)

tp:
`_CharCode.SP` is 32, which is already covered by the `byte > 31` part.

Should I do this change also in the NNBD version, or how are these kept in sync?


kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/138787

sortie:
lgtm. Sorry about the delay on this one, it fell through the cracks. Thank you for this improvement! The change is obviously correct. The sdk_nnbd directory has been merged into the regular sdk at this time.

PR #41127, opened 2020-03-20
Update deps to new version of package_config

### LINKED ISSUE #879: Use a lower default language version if there is a .packages file
The current spec for language versioning says that if no package_config.json file is found then libraries default to the latest language version supported by the SDK. Since the package_config.json file is itself new, we may find ourselves in this situation:

1. User upgrades to the latest Dart SDK which supports NNBD.
2. Since they have not run `pub get` yet, they have a `.packages` file, but not `package_config.json`.
3. The SDK doesn't see a `package_config.json` file, so it treats their libraries as all being at the latest version, which is NNBD.
4. They get tons of errors because their code is now treated as opted in.

This will be less of an issue if we ship pub support for `packages_config.json` *before* NNBD. But even then users who skip that SDK version and jump straight to the version that releases NNBD can be in this state.

Should we consider refining the proposal by saying that if there is no `packages_config.json` file, but there *is* a `.packages` file, then use something like 2.7 as the default language version?

@lrhn @leafpetersen @eernstg 

cc @sigmundch

mit-mit:
Fixes https://github.com/dart-lang/language/issues/879

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -112,7 +112,7 @@ vars = {
   "mustache_tag" : "5e81b12215566dbe2473b2afd01a8a8aedd56ad9",
   "oauth2_tag": "1.2.1",
   "observatory_pub_packages_rev": "0894122173b0f98eb08863a7712e78407d4477bc",
-  "package_config_tag": "87a8b5184020ebcc13b34ee95dde58f851b68ca3", # should be 1.9.0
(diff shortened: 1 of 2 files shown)


mit-mit:
Review in https://dart-review.googlesource.com/c/sdk/+/140244

PR #41374, opened 2020-04-07
Update gn_helpers.py

nemathahmed:
Just a minor change in a comment.


nemathahmed:
@googlebot I signed it!

mraleph:
Thank you for the contribution. However the original comment is already grammatically correct. It means "Helper functions (which are) useful ...".

Issue #41902, opened 2020-05-14
Allow `min` and `max` to work with Comparable's like DateTime


ChristianKleineidam:
I have a list of DateTime string and want to know the latest DateTime in my list. If DateTime would not only provide `.isBefore()` and` .isAfter()` I should be able to get the latest DateTime in my list with max (from dart:math). 

While I do grant that  `.isBefore()` and` .isAfter()` are normally more readable, they aren't as easy to use with existing functionality like `max` and `him` so it would be useful to have the syntax for both.

mraleph:
Note that `min` and `max` only work with numbers (their type parameter has `num` as a bound) and not with arbitrary comparable objects. So overriding `<` on `DateTime` would not permit you to use it with `min`/`max`.

There is no convenient method in the core library, but `kt_dart` package has `min` and `max` extension methods on `Iterable<T extends Comparable<T>>`, so if you have a `List<DateTime> l` you should be able to just do `l.min()` and `l.max()`. See https://pub.dev/documentation/kt_dart/latest/kt_collection/KtComparableIterableExtension.html

ChristianKleineidam:
It seems > and < aren't implemented because compareTo() and == have different behavior. == not only cares about the time the DateTime happens but also about the timezone that's set. 

Looking a bit more at the issue it seems the problem is the max working on `num `and not on `comparable`. 

Is there a good reason for `min` and `max` to not work for comparable?

lrhn:
I'd prefer to not make `min` and `max` from `dart:math` work on other types than numbers.

I'd rather add static `min` and `max` methods on `Comparable`:

```dart
class Comparable<T> {
  static V min<V extends Comparable<V>>(V first, V second) => 
    first.compareTo(second) <= 0 ? first : second;
  static V max<V extends Comparable<V>>(V first, V second) => 
    first.compareTo(second) >= 0 ? first : second;
  ...
```
Maybe even `V minAll<V extends Comparable<V>>(Iterable<V> values) => ...`.

ChristianKleineidam:
Okay, then it makes sense to have the functionality for min/max with Comparables outside of the math package.

How about a getter on `Iterable<Comparable>`? Then I could call `[Duration(hours: 1),Duration(hours: 2)].max` which is very concise and also easy to discover?

lrhn:
That is something I'm considering adding, potentially as part of package:collection.

ChristianKleineidam:
I did try to define this as an extension:

[...]

PR #42000, opened 2020-05-20
Added Iterable<Comparable> Extensions: .min and .max

ChristianKleineidam:
As discussed in https://github.com/dart-lang/sdk/issues/41902 I created two extensions to Iterable<Comparable>. .min that returns the smallest element and .max return s the largest one.

When it comes to the question of what to do when multiple elements have the same value I orientied myself with Pythons implementation of min() and max() that returns that both return the first value. 

I copied the headers from other files in their respective folders.

### DIFF
--- a/sdk/lib/collection/iterable_comparable_extensions.dart
+++ b/sdk/lib/collection/iterable_comparable_extensions.dart
@@ -1,0 +1,39 @@
+// Copyright (c) 2012, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// @dart = 2.6
+
+part of dart.collection;
+
+extension IterableComparableExtensions<T extends Comparable> on Iterable<T> {
+  /*
+  * Returns the largest value in the iterable determined via
+  * [Comparable.toCompare()].
+  *
+  * The iterable must have at least one element, otherwise
+  * [IterableElementError] gets thrown.
+  * If it has only one element, that element is returned.
+  *
+  * If multiple items are maximal, the function returns the first one encountered.
+  */
+  Comparable get max {
+    return this.reduce((Comparable value, Comparable element) =>
+        value.compareTo(element) >= 0 ? value : element);
+  }
+
+  /*
+  * Returns the largest value in the iterable determined via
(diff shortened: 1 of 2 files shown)


parlough:
For reference the CL for this is here: https://dart-review.googlesource.com/c/sdk/+/148744

lrhn:
This is not a bad idea. I'd like to consider implications a little more before landing our *first* extension methods in the platform libraries.

For example, should it be a method instead of a getter? Then we can do:

```dart
extension IterableExtension<T> on Iterable<T> {
  T max(int compare(T a, T b)) { 
    var it = this.iterator;
    if (!it.moveNext()) throw StateError("no element");
    var max = it.current;
    while (it.moveNext()) {
      var next = it.current;
      if (max.compareTo(next) < 0) max = next;
    }
    return max;
  }
}
extension IterableComparableExtension<T extends Comparable<T>> on Iterable<T> {
  T max([int compare(T a, T b)]) => 
      IterableExtension(this).max(compare ?? (a, b) => a. compareTo(b));
}
```
That would be  a more general API, and would still allow you to get the maximum of comparables without providing a comparator.

(I think we might indeed want that, and I'm currently working on something similar aimed for `package:collection`).

ChristianKleineidam:
To me it appears as if the max() you define on the general Iterable doublicates the functionality of the existing `.reduce()`-method and as such provides no additional value.

lrhn:
It's not exactly the same as `reduce`.

If you have 
```dart
T Function(T, T) maxBy<T>(int compare(T a, T b)) => (T a, T b) => compare(a, b) < 0 ? b : a;
```
then you can indeed do `values.max(compare)` as `values.reduce(maxBy(compare))`. It's still more convenient to have it directly, which is why this is a extension method rather than a method on `Iterable` itself. 

I have no problem having more than one way to do a thing, as long as the more specialized one is often useful and more readable.

ChristianKleineidam:
You are right that gives additional value. Given that sort() also takes such a compare function, it would also reuse existing patterns to do it that way. 

Having a function would also mean that you can pass a default value for cases where the Iterator is empty like Pythons max and min functions have.

PiN73:
Using method instead of getter might make sense because it's O(N)

mit-mit:
Closing as assumed stale

PR #42153, opened 2020-06-02
[nnbd_migration] Fix typos/mistakes

creativecreatorormaybenot:
The text says "We can fix this all by adding a single `/*!*/` hint.", but the code uses `/*?*/`. I am pretty sure that this is wrong.


mit-mit:
Code review in https://dart-review.googlesource.com/c/sdk/+/149685

mit-mit:
Merging now, thanks for your contribution @creativecreatorormaybenot !

PR #42157, opened 2020-06-02
Move extension methods experiment to shipped list

mit-mit:


mit-mit:
Review in https://dart-review.googlesource.com/c/sdk/+/149686

PR #42209, opened 2020-06-06
Doc: DateTime accepted format for parsing date

jagomf:
Add to the inline documentation the most simple and often most used simple date format that can be parsed, that was missing.


athomas:
Would you mind rebasing your change on master? It's currently based on a dev branch commit which confuses our automation tooling.

jagomf:
@athomas rebased to master

lrhn:
Thanks for the CL.

We also need this change in the sdk_nnbd source code. If we wait a few days, the two might be merged first, and then we won't have to do double work.

jagomf:
> Thanks for the CL.
> 
> We also need this change in the sdk_nnbd source code. If we wait a few days, the two might be merged first, and then we won't have to do double work.

Not sure how do you want me to proceed

lrhn:
Either make the same change in the `sdk_nnbd` directory or wait a few (work) days and see if it becomes unnecessary. Your call :smile:

jagomf:
Change done 👍

athomas:
Unfortunately, landing this will require golden files for the SDK to be updated as well (because line numbers changed), this command should do the trick:
```
dart pkg/dev_compiler/tool/check_nnbd_sdk.dart  --update-golden
```

jagomf:
@athomas can't seem to get that command working

image

athomas:
To do it, you'd have to install depot tools (see https://github.com/dart-lang/sdk/wiki/Building#linux) and run `gclient sync` (that will download all the dependencies needed by the Dart SDK).

I've run the tool for you so that in case the above is too much effort you could just apply this patch:
```
diff --git a/pkg/dev_compiler/tool/dartdevc_nnbd_sdk_error_golden.txt b/pkg/dev_compiler/tool/dartdevc_nnbd_sdk_error_golden.txt
index 2f6957baa8..3b7f59a28b 100644
--- a/pkg/dev_compiler/tool/dartdevc_nnbd_sdk_error_golden.txt
+++ b/pkg/dev_compiler/tool/dartdevc_nnbd_sdk_error_golden.txt
@@ -1,8 +1,8 @@
-ERROR|COMPILE_TIME_ERROR|CONST_CONSTRUCTOR_THROWS_EXCEPTION|lib/core/core.dart|3722|5|94|Const constructors can't throw exceptions.
-ERROR|COMPILE_TIME_ERROR|CONST_CONSTRUCTOR_THROWS_EXCEPTION|lib/core/core.dart|7912|5|97|Const constructors can't throw exceptions.
-ERROR|COMPILE_TIME_ERROR|CONST_CONSTRUCTOR_THROWS_EXCEPTION|lib/core/core.dart|940|5|95|Const constructors can't throw exceptions.
-ERROR|COMPILE_TIME_ERROR|CONST_CONSTRUCTOR_THROWS_EXCEPTION|lib/core/core.dart|973|5|94|Const constructors can't throw exceptions.
-ERROR|SYNTACTIC_ERROR|CONST_FACTORY|lib/core/core.dart|3720|3|5|Only redirecting factory constructors can be declared to be 'const'.
-ERROR|SYNTACTIC_ERROR|CONST_FACTORY|lib/core/core.dart|7910|3|5|Only redirecting factory constructors can be declared to be 'const'.
-ERROR|SYNTACTIC_ERROR|CONST_FACTORY|lib/core/core.dart|938|3|5|Only redirecting factory constructors can be declared to be 'const'.
-ERROR|SYNTACTIC_ERROR|CONST_FACTORY|lib/core/core.dart|971|3|5|Only redirecting factory constructors can be declared to be 'const'.
+ERROR|COMPILE_TIME_ERROR|NULLABLE_TYPE_IN_CATCH_CLAUSE|lib/io/io.dart|6330|10|7|A nullable type can't be used in an 'on' clause because it isn't valid to throw 'null'.
+ERROR|STATIC_WARNING|ARGUMENT_TYPE_NOT_ASSIGNABLE|lib/io/io.dart|4015|66|3|The argument type 'int?' can't be assigned to the parameter type 'int'.
+ERROR|STATIC_WARNING|ARGUMENT_TYPE_NOT_ASSIGNABLE|lib/io/io.dart|4044|59|3|The argument type 'int?' can't be assigned to the parameter type 'int'.
+ERROR|STATIC_WARNING|USE_OF_NULLABLE_VALUE|lib/_internal/js_dev_runtime/private/ddc_runtime/runtime.dart|3631|22|5|The expression is nullable and must be null-checked before it can be used.
+ERROR|STATIC_WARNING|USE_OF_NULLABLE_VALUE|lib/io/io.dart|3951|18|3|The expression is nullable and must be null-checked before it can be used.
+ERROR|STATIC_WARNING|USE_OF_NULLABLE_VALUE|lib/io/io.dart|4024|18|3|The expression is nullable and must be null-checked before it can be used.
+ERROR|STATIC_WARNING|USE_OF_NULLABLE_VALUE|lib/io/io.dart|4046|9|3|The expression is nullable and must be null-checked before it can be used.
+ERROR|STATIC_WARNING|USE_OF_NULLABLE_VALUE|lib/io/io.dart|4050|28|3|The expression is nullable and must be null-checked before it can be used.

```

jagomf:
Thanks @athomas for your patch. Just applied it.

athomas:
Sorry, but yesterday the change to remove sdk_nnbd landed, and now the change to  sdk_nnbd/lib/core/date_time.dart creates a conflict. You could fix that yourself by just deleting that file from your change (the other changes should still be there). Alternatively, I can take it from here, but I'd rather not take credit for your work.

jagomf:
Change removed from sdk_nnbd/lib/core/date_time.dart

Issue #34962, opened 2018-10-29
DateTime unsafely uses private members of arguments.


sanderAtMaxdoro:
Thank you for taking the time to file an issue!

In order to route, prioritize, and act on this, please include:

* Dart SDK Version (`dart --version`)
Flutter 0.10.2-pre.47 • channel master • https://github.com/flutter/flutter.git
Framework • revision 9dfc0f3aaa (5 days ago) • 2018-10-24 07:22:08 +0100
Engine • revision 2586e94122
Tools • Dart 2.1.0-dev.7.1.flutter-45f9462398

* Whether you are using Windows, MacOSX, or Linux (if applicable)
MacOSX

* Whether you are using Chrome, Safari, Firefox, Edge (if applicable)
NA

Missing some or all of the above might make the issue take longer or be
impossible to act on.

----

Is it really an issue? For general questions consider starting with Stack
Overflow:
https://stackoverflow.com/questions/tagged/dart

Also consider our Gitter channel for light-weight/quick discussions:
https://gitter.im/dart-lang/sdk

-----
When you let a class implement another class using

class MyClass implements YourClass

the

final <Type> _propertyName;

vars are part of the public class interface. In particular, MyClass is expected to implement  _propertyName by the implementation of YourClass, even though there is no proper way to implement that when you look at the public interface. You can of course see it in the code if you have the code and add it, but that is fraud (in the Stroustrup sense).

Try this by implementing your own MyDateTime as an implementation of DateTime. MyDateTime is expected to implement final int _value, and a NoSuchMethod is thrown if it is not found by DateTime.

lrhn:
If I understand this correctly, the issue is that `DateTime` assumes that every implementation of the interface also has a `_value` getter. That's obviously not true if the interface is implemented in a different library. The code that makes this assumption should be written to work in terms of the public interface of the class.

Dart does allow implementing interfaces that have private and inaccessible members. That is generally unsafe, but only to the library that declared this public interface with private members, the rest of the world can't see the members anyway.

lrhn:
The proper behavior is to call `Bar.noSuchMethod` with `Invocation.getter(#_f)` where `#_f` is the symbol from `foo.dart`. (I'll write that symbol as `#foo::_x` for now, there is obviously no way to write it directly in `bar`)

More precisely: The class `Bar` does not implement `int get _x` or `set _x(int _)`, and since the members are inaccessible, that is accepted and noSuchMethod-forwarders are inserted:
```dart
[...]

PR #42351, opened 2020-06-15
Fix remaining instance where DateTime accesses the private member _value

matthewlloyd:
This does not work if the DateTime accessed is an implementation of
the DateTime interface, rather than an instance of the same class.

Related issues:

* https://github.com/dart-lang/sdk/issues/34962
* https://github.com/srawlins/timezone/issues/57

Related changes:

* https://dart-review.googlesource.com/c/sdk/+/81828


mit-mit:
Gerrit review is https://dart-review.googlesource.com/c/sdk/+/151322

PR #42564, opened 2020-07-02
Uri.https and Uri.http constructors queryParams type

jagomf:
Currently `Uri.http` and `Uri.https` factory constructors do not allow for `queryParameters` to be `Map<String, dynamic>`, but only `Map<String, String>`, when in fact the main `Uri` constructor does work with `queryParameters` as `Map<String, dynamic>`.


mraleph:
Gerrit review at https://dart-review.googlesource.com/c/sdk/+/153182

jagomf:
@mraleph @lrhn anything I can help with so that this goes though?

mraleph:
@jagomf in case you have not seen review comments on Gerrit: this needs CHANGELOG.md entry and tests to land according to @kevmoo comment.

jagomf:
@mraleph I have updated Changelog file (hopefully in the right place).

Regarding tests, they are way far from my Dart knowledge. I hope that someone can contribute so that this inconsistency can be solved.

jagomf:
@mraleph I have updated Changelog file (hopefully in the right place).

Regarding tests, they are way far from my Dart knowledge. I actually can't even think of tests for types inconsistency @kevmoo.

I hope that someone can contribute so that this inconsistency can be solved.

mraleph:
Thanks @jagomf! @lrhn do you think tests are absolutely necessary to land this?

@athomas CopyBara can't sync this into Gerrit. Any idea why?

lrhn:
I want tests, because otherwise it probably doesn't work. (Law of nature: If it isn't tested, it's broken.)

If we land this and someone promises to write test for it quickly, we can probably do it in two steps, but we *must not* release this code without adding tests.

athomas:
@mraleph What is not being synced? You posted the link to the Gerrit review yourself and I can see e.g. the CHANGELOG.md there and the diff looks the same to me? I also only see successful jobs for this PR.

mit-mit:
@lrhn can you provide some guidance to @jagomf regarding how to test this?

lrhn:
I'll add some tests. It shouldn't be much, just checking that it actually works for each constructor.

Issue #37685, opened 2019-07-31
Iterable<E>.contains() argument should be templated?


duzenko:
Currently the `contains(Object element)` method accepts any argument value.
I wonder why not limit `element` to template type `E`?

mit-mit:
cc @lrhn

duzenko:
I have the same complain towards the Map [] operator.
```
final m = Map<num,num>();
final x = m['what???'];
```

duzenko:
Somewhat related - Map.fromIterable could use templating:
`Map.fromIterable(Iterable iterable,      {K key(element), V value(element)})`
=>
`Map.fromIterable<T>(Iterable<T> iterable,      {K key(T element), V value(T element)})`
Right now constructions like 
`Map<num,String>.fromIterable(myList, key: (x)=>x.id, value: (x)=>x.name)` are asking for bugs due to abuse of dynamic downcasts.

lrhn:
This has been asked before, e.g. https://github.com/dart-lang/sdk/issues/30405.
In short, there are reasons for the current design, and changing it now would be breaking.

The `Object` typed parameters are there because of covariant generics. You can treat a `Set<int>` as a `Set<num>` or `Set<Object>`, and all non-mutating functions still work. That means that `contains` needs to be able to accept any `Object`.

The option of typing the parameter as `E` in the interface, and then make all implementations actually accept `Object`, is not really viable. It's too easy to create something with the interface type.

The map constructors are a consequence of the lack of generic constructors (https://github.com/dart-lang/sdk/issues/30355).

lrhn:
It can count as an individual request, but this particular request is solidly in the "unlikely to occur" bucket. 
The one duplicate issue that was linked was not the only issue ever, just the most recent one. We also have, e.g., #26852 and #22781 which are basically the same issue.

Changing the member declaration now is breaking. Even if we do so, it will mean that there are use-cases that are no longer supported. Unless we introduce a language feature which makes it possible to declare a method parameter to have a static type of `E` and a run-time type of `Object`, ensuring that all implementations of the method will support this, then we are unlikely to change the current design. That is why this request is closed. I'm adding a "not planned" along with the "duplicate" closing marker, just to make it clear why it is closed.

PR #42867, opened 2020-07-29
Update list.dart

tomyeh:
It shall be consistent with `List<E>` since List is what the caller access, and it makes the overriding easier:

```
class Foo<E extends Base> with ListMixin<E> implements List<E> {
  @override
   int indexOf(E element, [int start = 0]) {
      //then we can access Base's member directly without casting it to E first
```

### DIFF
--- a/sdk/lib/collection/list.dart
+++ b/sdk/lib/collection/list.dart
@@ -482,7 +482,7 @@ abstract class ListMixin<E> implements List<E> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


mraleph:
It is intentional that these parameters are typed as `Object` rather than `E`, also making this change now is breaking so unlikely to happen. See https://github.com/dart-lang/sdk/issues/37685#issuecomment-525257492 for explanation.

Issue #42930, opened 2020-08-04
Expose additional Dart APIs through ffi


blagoev:
An issue from the MongoDb Realm team:

We got a lot of requests for creating a Flutter SDK for Realm mobile database (https://github.com/realm/realm-object-server/issues/55 now has over a 1000 upvotes) and initially made a prototype with FFI. Unfortunately, our progress is blocked for over a year now on https://github.com/dart-lang/sdk/issues/37022 and https://github.com/dart-lang/sdk/issues/35770.

We started to look at an alternative route and found one, but it would require adding official support for a set of the Dart VM API’s in Flutter. 

Realm JS SDK is designed as a Node.js native extension with an abstraction layer that abstracts away the JavaScript VM and allows for targeting different JavaScript VMs on various platforms we support. 

The similarity between the Dart VM APIs and the JavaScript VM APIs allows us to reuse a major part of Realm database code that is already tested and provide only a platform specific layer for Dart and Flutter. Moreover this approach would be superior to calling Realm through the FFI mechanism.

We made a [POC](https://drive.google.com/file/d/1_Ns_55vnxv49IHKavyOo5xMAp6fkmUPs/view?usp=sharing) with a custom built Flutter engine dynamic library which exposes a set of the Dart VM API’s and allows us to reuse a major part of Realm JavaScript engine abstraction layer with the Dart and Flutter platforms. We achieved it by exposing some of the Dart API’s from the libflutter.so library by editing the https://github.com/flutter/engine/blob/master/shell/platform/android/android_exports.lst.

In the light of the recent changes https://dart-review.googlesource.com/c/sdk/+/145592, we would like to request additional Dart APIs exposed officially so it can be consumed in Flutter through the Dart ffi. This will unblock us on delivering Flutter support and potentially will also enable other Node.js native extensions built on N-API and V8 API’s to be ported to Flutter and Dart in a straightforward manner, which would be beneficial to the Flutter and Dart community as a whole.

Here is a PR with the changes in question: https://github.com/dart-lang/sdk/pull/42931

cheers,
blagoev

mraleph:
> Unfortunately, our progress is blocked for over a year now on #37022 and #35770.

I think both of the issues now have their own workarounds (for asynchronous case there are send ports and for finalizers there are weak handles). 

Are you still blocked even given those work-arounds? 

> we would like to request additional Dart APIs exposed officially so it can be consumed in Flutter through the Dart ffi

[...]

PR #42931, opened 2020-08-04
Expose dart functions to extenders

blagoev:
This enabled extenders like MongoDb Realm to use the Dart API in Flutter. Relevant issue is here: https://github.com/dart-lang/sdk/issues/42930


blagoev:
@googlebot I signed it!

kevmoo:
https://dart-review.googlesource.com/c/sdk/+/157061

mraleph:
Thank you for the contribution @blagoev, however as I have indicated both on the Gerit CL and on the issue you have filed I don't think exposing Dart VM C API through FFI aligns with our vision for FFI - so we are not going to expose these functions. 

We however would be very happy to work with you to ensure that Realm can achieve everything you want through FFI itself. 

I believe that right now you should not have any blockers - as both finalizers and async communication is exposed. Let us know if you need any questions.

Issue #42946, opened 2020-08-05
VM const evaluation fails to recognize `2 == 2.0`.


lrhn:
See: https://stackoverflow.com/questions/63261894/int-and-double-equality-with-const-in-dart

Example:
```dart
main() {
  const x = (2 == 2.0);
  print(x); // prints `false`
}
```
This should print `true`, and it does so for dart2js (according to dartpad), so it seems to be a VM issue.

Tested with the following VMs:
* 2.10.0-3.0.dev
* 2.9.0-21.4-beta

on Windows (Chocolatey installs).

mraleph:
Constant evaluation is done in the front-end, so it's a front-end issue.

lfkdsk:
fix it https://github.com/dart-lang/sdk/pull/42949

PR #42949, opened 2020-08-05
[front-end] bugfix: fix bug in constant eval .

lfkdsk:
fix bug in constant eval.
https://github.com/dart-lang/sdk/issues/42946

``` dart
main() {
  const x = (2 == 2.0);
  print(x); // prints `true`
}
```

@mraleph


mraleph:
Thank you for the contribution. Could you add a regression test? You can place it into `tests/language_2/regress/` following the naming scheme there (`tests/language_2/regress/regress_42946_test.dart`). You can look at how other regression tests look like in that folder (e.g. they use `package:expect/expect.dart` for checking things).

lfkdsk:
update regress test

Issue #42997, opened 2020-08-09
Compiler crashed with "type 'AmbiguousBuilder' is not a subtype of type 'Expression'"


Hixie:
A tried to compile the attached code and the compiler crashed as follows:
```
$ flutter run
Downloading android-arm-profile/linux-x64 tools...                  0.6s
Downloading android-arm-release/linux-x64 tools...                  0.4s
Downloading android-arm64-profile/linux-x64 tools...                0.4s
Downloading android-arm64-release/linux-x64 tools...                0.5s
Downloading android-x64-profile/linux-x64 tools...                  0.4s
Downloading android-x64-release/linux-x64 tools...                  0.4s
Downloading Web SDK...                                              1.4s
Running "flutter pub get" in dummy_51752...                         0.8s
Launching lib/main.dart on Lenovo TB 8504F in debug mode...
Running Gradle task 'assembleDebug'...
Unhandled exception:
Crash when compiling file:///home/ianh/dev/dummy_51752/lib/main.dart,
at character offset 2511:
type 'AmbiguousBuilder' is not a subtype of type 'Expression'
#0      BodyBuilder.doBinaryExpression (package:front_end/src/fasta/kernel/body_builder.dart:1767:46)
#1      BodyBuilder.endBinaryExpression (package:front_end/src/fasta/kernel/body_builder.dart:1718:7)
#2      Parser.parsePrecedenceExpression (package:_fe_analyzer_shared/src/parser/parser_impl.dart:4644:20)
#3      Parser.parseExpression (package:_fe_analyzer_shared/src/parser/parser_impl.dart:4466:13)
#4      Parser.parseForLoopPartsMid (package:_fe_analyzer_shared/src/parser/parser_impl.dart:6180:15)
#5      Parser.parseForStatement (package:_fe_analyzer_shared/src/parser/parser_impl.dart:6118:13)
#6      Parser.parseStatementX (package:_fe_analyzer_shared/src/parser/parser_impl.dart:4258:14)
#7      Parser.parseStatement (package:_fe_analyzer_shared/src/parser/parser_impl.dart:4217:20)
#8      Parser.parseFunctionBody (package:_fe_analyzer_shared/src/parser/parser_impl.dart:4122:15)
#9      DietListener.buildFunctionBody (package:front_end/src/fasta/source/diet_listener.dart:990:14)
#10     DietListener._endClassMethod (package:front_end/src/fasta/source/diet_listener.dart:714:5)
[...]

PR #42998, opened 2020-08-10
[front-end] fix bug in BodyBuilder 

lfkdsk:
https://github.com/dart-lang/sdk/issues/42997

@mraleph 

> Also, how to construct tests that generate AmbiguousBuilder and catch error?

### DIFF
--- a/pkg/front_end/lib/src/fasta/kernel/body_builder.dart
+++ b/pkg/front_end/lib/src/fasta/kernel/body_builder.dart
@@ -1763,6 +1763,10 @@ class BodyBuilder extends ScopeListener<JumpTarget>
       } else if (left is Generator) {
         push(left.buildBinaryOperation(token, name, right));
       } else {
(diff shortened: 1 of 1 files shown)


mraleph:
Thanks for the contribution. 

Gerrit review here: https://dart-review.googlesource.com/c/sdk/+/157941

I have forwarded it to the right people on the CFE team for review - they should also be able to help you with how to create a regression test.

johnniwinther:
Do you have a small (self contained) repro for the crash? If so, normally we add this as an "expectation test" (in `pkg/front_end/testcases/general/`). This is a bit involved but I you (already) have a small repro I can add it to the CL.

lfkdsk:
https://github.com/lfkdsk/dummy_51752 I created such a Flutter project as described in issues.

johnniwinther:
Thanks. I'll try to make it self-contained (not needing the flutter import) and add it to the CL.

PR #43480, opened 2020-09-18
[vm, service] Dump complete gc roots.

zhiwei-wu:
Incomplete gc roots causes many objects to be placed in the uncollected garbage when load the Snapshot file.


zhiwei-wu:
@googlebot I signed it!

PR #43481, opened 2020-09-18
[vm, service] Dump complete gc roots.

zhiwei-wu:
Incomplete gc roots causes many objects to be placed in the uncollected garbage when load the Snapshot file.


mit-mit:
Gerrit CR in https://dart-review.googlesource.com/c/sdk/+/163361

athomas:
According to the Gerrit CL, these changes have been applied in a different commit.

PR #43491, opened 2020-09-20
Export CommentToken in token.dart

rrousselGit:
This exports `CommentToken`

That class may be reached when doing:

```dart
AstNode node;
CommentToken = node.beginToken.precedingComments;
```

### DIFF
--- a/pkg/analyzer/lib/dart/ast/token.dart
+++ b/pkg/analyzer/lib/dart/ast/token.dart
@@ -5,4 +5,4 @@
 /// Defines the tokens that are produced by the scanner, used by the parser, and
 /// referenced from the [AST structure](ast.dart).
 export 'package:_fe_analyzer_shared/src/scanner/token.dart'
-    show Keyword, Token, TokenType, LanguageVersionToken;
+    show Keyword, Token, TokenType, LanguageVersionToken, CommentToken;


mraleph:
Gerrit review https://dart-review.googlesource.com/c/sdk/+/163561

/cc @devoncarew @bwilkerson

bwilkerson:
I have already replied to the CL on Gerrit and am waiting for a response to the question I asked.

mraleph:
@bwilkerson thanks!

FWIW the type is part of the public interface https://pub.dev/documentation/analyzer/latest/dart_ast_token/Token/precedingComments.html so it leaks to the user (user can in fact access its public members), but user can't declare any variable with this type explicitly. 

```
final /* CommentToken */ a = tok.precedingComments; // ok, static type inferred CommentToken.
final CommentToken a = tok.precedingComments; // error
```

I don't know if this is desirable.

bwilkerson:
As I said on the CL, the original intention was for `Token` to be the only class in the public interface and for all of the behavior to be available from there. The reference to `CommentToken` was unintentional and, imo, should be changed to `Token`. What I'm trying to ascertain is how big of a breaking change it would be to change the return type of `precedingComments` by understanding what, if any, `CommentToken`-specific behavior is needed by clients.

mit-mit:
@rrousselGit thanks for the contribution, can you please respond to the review comment in https://dart-review.googlesource.com/c/sdk/+/163561 ?

rrousselGit:
Sorry, I didn't see the comment.

May I suggest a bot that synchronize comments on dart-review with github?

So that people watching the PR get notified

mit-mit:
@rrousselGit are you still planning on working on this?

rrousselGit:
I've already answered the comments:

I don't really care about CommentToken. It just was used by a public API but not available for use

So I don't really care about whether CommentToken would be exported or whether the APIs that use it change to use Token

So we can probably close this

PR #43631, opened 2020-10-01
Update formatting in CHANGELOG.md

agilob:

PR #43636, opened 2020-10-01
tiny syntax fix

boukeversteegh:


mraleph:
Thanks for the contribution! Fixed by #43631 which was submitted a bit earlier.

Issue #43474, opened 2020-09-18
please rename nativewrappers.dart to native_wrappers.dart


Silentdoer:
otherwise it's not consistent with other files naming style.
such as:
native_typed_data.dart

Silentdoer:
and it will tips by IDE: Typo: In word 'nativewrappers'

mraleph:
There is not much value in such rename and too much breakage.

Silentdoer:
ahh, i wish dart is a perfect, more consistent language ..

PR #43700, opened 2020-10-07
make file and library name consistent with others

Silentdoer:
like title.


mraleph:
It's a breaking change which does not really improve anything (it does not matter how this library name is spelt). As such we are not interested in doing this (as was already indicated before on https://github.com/dart-lang/sdk/issues/43474).

Silentdoer:
it is a few breaking change, and i think has very little peoples will use nativewrappers, rename it
 can improve  dart's consistent.

PR #43731, opened 2020-10-09
Support latest analyzer in analyzer_plugin

dkrutskikh:
I've raised the upper constraint for the analyzer to <0.41.0.


dkrutskikh:
@googlebot I fixed it.

PR #44190, opened 2020-11-13
Override contains in LinkedList to make it constant time

### LINKED ISSUE #44189: Improve LinkedList contains performance
Currently `LinkedList` does not override the `contains` method of `Iterable`
However there is a very simple optimisation that can be done by `LinkedList` to make `contains` take constant time instead of linear time by simply comparing the `list` property of the entry to itself.

Kavantix:
Fixes #44189

Q: are there any tests for `LinkedList` because I couldn't find them.
If not, should I create a new `linked_list_contains_test.dart' file in `tests/corelib/`?

### DIFF
--- a/sdk/lib/collection/linked_list.dart
+++ b/sdk/lib/collection/linked_list.dart
@@ -122,6 +122,17 @@ class LinkedList<E extends LinkedListEntry<E>> extends Iterable<E> {
     } while (!identical(current, _first));
   }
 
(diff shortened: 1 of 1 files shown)


mit-mit:
cc @lrhn

athomas:
Thanks for the PR and the idea! @lrhn has created a similar CL the we will land instead: https://dart-review.googlesource.com/c/sdk/+/172161

Issue #11189, opened 2013-06-10
DateTime.parse should throw an error on  invalid date...


DartBot:
_This issue was originally filed by adrian.avil...&#064;gmail.com_

---

**What steps will reproduce the problem?**
1. Put this line and run: DateTime pDate = DateTime.parse(&quot;20130547&quot;);
2. It will return: 2013-06-16 00:00:00.000
**3.**

**What is the expected output? What do you see instead?**
Since the say is 47 it should throw an exception.

**What version of the product are you using? On what operating system?**
Dart Editor version 0.5.13_r23552
Dart SDK version 0.5.13.1_r23552
Windows 8 64bits.

lrhn:
I agree that `DateTime.parse` should throw on invalid inputs, where invalid inputs are anything that isn't a valid date in a recognized format. 
The reasoning is exactly what has been written here: There is no simpler way to validate input than to parse it, so if the parsing doesn't catch invalid input, nobody does.

The other `DateTime` constructors, then ones taking integers as input, should accept the input they are given, and be generous in the interpretation. That's actually useful, and it's simple to check if the input is valid (create the `DateTime`, then check that the year, month and date still match the constructor arguments, not something you can easily do with string inputs).

alan-knight:
There are some interesting edge cases with that kind of validation. e.g. validating February 18, 2018 by passing in integers when in Brazil.

hoylen:
alan-knight writes:
> interesting edge cases... February 18, 2018 by passing in integers when in Brazil

I assume you are talking about the general issue of dealing with daylight saving changes.

Daylight saving is a separate and complex issue in itself, so let's not try and solve it as a part of this issue - which is only about how `DateTime.parse` validates strings.

The `DateTime.parse` method currently does not allow the string to specify whether daylight saving is in use or not. The timezone component of the string (when present) only allows explicit offsets in hours and minutes from UTC, or "Z" - so there is no ambiguity there. That is, 30 minutes before the changeover the BRST time would be "2018-02-17 11:30-02:00" and 30 minutes after the changeover would be the BRT time of "2018-02-17 11:30-03:00" (i.e. the two times are 01:30 UTC and 02:30 UTC). Without the timezone, the string will be parsed differently depending on where/when the program was run (i.e. what the local timezone was at runtime). This is an example of why it is important to have the option of making the parser reject strings which don't have an explicit timezone in it.

lrhn writes:
> I agree that DateTime.parse should throw on invalid inputs...

[...]

PR #44521, opened 2020-12-21
Added documentation about out-of-range components to DateTime.parse.

hoylen:
The [DateTime.parse](https://api.dart.dev/stable/2.10.4/dart-core/DateTime/parse.html) method accepts out-of-range components. For example, parsing the string "2020-01-42" produces "2020-02-11".

This was raised as a bug in https://github.com/dart-lang/sdk/issues/11189, but after more than 7 years that issue was closed with the solution being "use parseStrict from the intl package" instead.

This pull request simply **updates the documentation of DateTime's _parse_ method** so that users know:

1. This is how the method behaves. Otherwise they might be surprised when it unexpectedly happens in their code.
2. This behaviour is a feature and not a bug. Otherwise they might raise another issue for it.
3. The _parseStrict_ method in the intl package exists. Otherwise they might reinvent the wheel.


kevmoo:
Tracked at https://dart-review.googlesource.com/c/sdk/+/176680

mraleph:
@hoylen see review comments on https://dart-review.googlesource.com/c/sdk/+/176680

once they are addressed just push a new commit into your branch and it will be picked up automatically.

sdk/lib/core/date_time.dart
```diff
@@ -275,8 +275,12 @@ class DateTime implements Comparable<DateTime> {
    * * `"2002-02-27T14:00:00-0500"`: Same as `"2002-02-27T19:00:00Z"`
    *
    * This method accepts out-of-range component values and interprets
-   * them as having overflows. For example, "2020-01-42" will be parsed
-   * as 11 February 2020. To detect and reject invalid component values, use
+   * them as overflows into the next larger component.
+   * For example, "2020-01-42" will be parsed as 2020-02-11, because
+   * the last valid date in that month is 2020-01-31, so 42 days is
+   * interprted as 31 days of that month plus 11 days into the next month.
```
MrBirb, line 281: nit: interpreted*

Jonas-Sander:
Would really love to see this. Is this still being discussed?

mraleph:
This is ready to land by requires rebase.

hoylen:
Rebase done. Ready "to land".

kevmoo:
@mraleph @lrhn – can you take a look and hit the button in gerrit?

PR #44634, opened 2021-01-10
Add insertInBetween to List

ghost:
## I did not manage to build the sdk myself even when following the steps in the build process

<details><summary>SHOW ERROR</summary>
<p>

#### I just ran this in my home folder

> mkdir dart-sdk
> cd dart-sdk
> fetch dart

#### And after what felt like an eternity of waiting

```
________ running 'cipd ensure -log-level error -root /home/dennis/dart-sdk -ensure-file /tmp/tmpt7c0fd8t.ensure' in '.'
[P24278 21:35:21.283 client.go:1498 E] cipd: failed to fetch flutter/android/ndk/linux-amd64:hZB1c9QLM86cSJMosTF5e1qhXQ-yoMKn7Pk5lf04_AEC - failed to download the package file after multiple attempts
[P24278 21:35:21.386 client.go:1711 E] Failed to install flutter/android/ndk/linux-amd64:hZB1c9QLM86cSJMosTF5e1qhXQ-yoMKn7Pk5lf04_AEC - failed to download the package file after multiple attempts
[P24278 21:38:12.634 client.go:1498 E] cipd: failed to fetch fuchsia/sdk/gn/linux-amd64:CghlNlbrZrJDAC1p5omkoqFd3eI_UADZsay7wJDZRAYC - failed to download the package file after multiple attempts
[P24278 21:38:12.662 client.go:1711 E] Failed to install fuchsia/sdk/gn/linux-amd64:CghlNlbrZrJDAC1p5omkoqFd3eI_UADZsay7wJDZRAYC - failed to download the package file after multiple attempts
[P24278 21:38:32.446 client.go:1711 E] Failed to install flutter/clang/win-amd64:8yiAFYUeCVlBkEiAdKb6qPFG0mcVnUGfKIKM_gRvatgC - write /home/dennis/dart-sdk/.cipd/pkgs/15/FcaYTjqL7JQG/bin/clang-cl.exe: no space left on device (and 167 other errors)
[P24278 21:39:09.653 client.go:1498 E] cipd: failed to fetch dart/dart-sdk/linux-amd64:QNGhSr8VF7eyfzzxDzv9yDxPtbCIhNlSyNXU2w2_ZgYC - failed to download the package file after multiple attempts
[P24278 21:39:09.677 client.go:1711 E] Failed to install dart/dart-sdk/linux-amd64:QNGhSr8VF7eyfzzxDzv9yDxPtbCIhNlSyNXU2w2_ZgYC - failed to download the package file after multiple attempts
Error: failed to update packages, see the log.
Error: Command 'cipd ensure -log-level error -root /home/dennis/dart-sdk -ensure-file /tmp/tmpt7c0fd8t.ensure' returned non-zero exit status 1
Subprocess failed with return code 1.
```

</p>
</details>

I cannot guarantee that it will work since I did not manage to get it running, but the point is that I think that something like this will be incredibly useful for [Flutter](https://github.com/flutter/flutter)

It started out in one of the projects in my company as a util function:

```dart
extension InbetweenInsert<T> on List<T> {
  void insertInBetween(T element) {
    for (int i = 1; i < length; i += 2) {
      insert(i, element);
    }
  }
}
```

But when I saw that we use it pretty much everywhere, I though that maybe other people might want to know about it as well.
Here is an example use case:

```dart
Column(
  children: [
    Button(
      variant: 'primary',
      buttonText: buttonText,
      onPressed: onConfirmButtonPressed,
    ),
    if (randomCondition)
      Button(
        variant: 'outline',
        buttonText: buttonTextDecline,
        onPressed: onDeclineButtonPressed,
      ),
  ]..insertInBetween(SizedBox(height: 20)),
),
```

This could be done with Flexbox as well, but that is usually not an option, since I know the spacing I want and not the final size. So the size depends on the spacing. 

I would love to know what you think.

### DIFF
--- a/sdk/lib/collection/list.dart
+++ b/sdk/lib/collection/list.dart
@@ -533,6 +533,12 @@ abstract class ListMixin<E> implements List<E> {
     }
   }
 
+  void insertInBetween(E element) {
+    for (int i = 1; i < length; i += 2) {
+      insert(i, element);
+    }
+  }
+
   E removeAt(int index) {
     E result = this[index];
     _closeGap(index, index + 1);
--- a/sdk/lib/core/list.dart
+++ b/sdk/lib/core/list.dart
@@ -493,6 +493,13 @@ abstract class List<E> implements EfficientLengthIterable<E> {
   void insert(int index, E element);
 
   /**
+   * Inserts [element] in between the elements of the collection.
+   * 
+   * For example, `[1, 1, 1].insertInBetween(2)` will result in 
+   * `[1, 2, 1, 2, 1]`
+   */
+  void insertInBetween(E element);
+  /**
    * Inserts all objects of [iterable] at position [index] in this list.
    *
    * This increases the length of the list by the length of [iterable] and
--- a/tests/lib/collection/list_test.dart
+++ b/tests/lib/collection/list_test.dart
@@ -150,6 +150,15 @@ void testRetainWhere() {
   }
 }
 
+void testInsertInBetween() {
+  List<int> list = [1, 1, 1];
+
+  list.insertInBetween(2);
+
+  Expect.equals(list, [1, 2, 1, 2, 1]);
+}
+
 void main() {
   testRetainWhere();
+  testInsertInBetween();
 }


mraleph:
Thanks for your contribution @DenisBarzanov! 

Unfortunately adding any methods to a core class like `List` is a non-starter at the moment: when you add a method you break every single class in the ecosystem that implements `List` (because it now needs to provide a valid override for this newly added method - and it does not), and there are a lot of classes implementing `List`... so adding a method becomes a huge breaking change. 

There might be an argument for adding such an extension to `package:collection` but I let @lrhn to comment on that. 

(btw, canonical name of this operation is _intersperse_. There is a package that implements this extension method on iterables and lists: https://pub.dev/packages/intersperse).

lrhn:
It's a function that I've personally never had any need for, so I'd be a little wary that it's not going to carry its own weight (read: not being used enough to warrant having to maintain it). Still, use cases differ, and you've obviously found use for it.

I'm also a little worried about the typing. You have to have a `List<T>` which can accept the value you use as separators. If you have a `List<MyWidet>` and want to insert a `SizedBox` between the widgets, it won't work because a `SizedBox` is not a `MyWidget`. That would work better as a constructor `List<T>.interspersed(List<T> values, T separator)`, because then you can create a new list. I can see uses for both modifying the list in-place and creating a new list. Since the result is always about twice as big as the original, and we know the size ahead of time, I think allocating a new list is actually *better* than modifying in-place. It's likely to be more efficient (the original list was going to grow anyway), and it's easier to avoid moving elements more than once.
Then it worries me to always use the same value as separator, and not have the option of creating a new one per position. It's nice that `SizedBox` is reusable, but I can imagine other settings where you need to allocate a new value for each separator.

So, all in all, the design space here is wider than it initially seems, and singling out a single use case is likely to leave us with a function which is too specialized to be generally useful. Making the function more general might make it less convenient, and it might still not be worth the wok needed to maintain it. I tend towards not adding a function to `package:collection`, at least until I have seen more use-cases where it would apply.

PR #44647, opened 2021-01-12
Add --no-sound-null-safety option to dart2native.

### LINKED ISSUE #44553: dart2native doesn't support --no-sound-null-safety
Using `dart2native` 2.12.0-133.2.beta on "mixed" null safe code I get the following error:

```
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:async
 - package:stream_transform
 - package:path
 - package:crypto
 - package:collection
 - package:convert
 - package:typed_data
 - package:charcode

For solutions, see https://dart.dev/go/unsound-null-safety
```

That page mentions `--no-sound-null-safety` which sounds like exactly what I want, however it isn't a valid parameter to `dart2native`. So there's no way to compile my code with `dart2native`.

slightfoot:
Currently running `dart compile exe` does not support `--no-sound-null-safety` meaning currently during the null safety transition period we cannot build binary files. The `dart compile exe` wrapper executes the `dart2native` tool.

This PR fixes this by adding: `--no-sound-null-safety`, `--sound-null-safety`.

We also match the `--extra-gen-snapshot-options`  hidden option by also adding `--extra-gen-kernel-options` (hidden) option so the tool is flexible in the future.

Fixes: #44553

/cc @mit-mit I guess you are the right person to initially review this.

### DIFF
--- a/pkg/dart2native/bin/dart2native.dart
+++ b/pkg/dart2native/bin/dart2native.dart
@@ -31,6 +31,12 @@ E.g.: dart2native -Da=1,b=2 main.dart''')
     ..addFlag('enable-asserts',
         negatable: false, help: 'Enable assert statements.')
     ..addMultiOption(
+      'extra-gen-kernel-options',
+      help: 'Pass additional options to gen_kernel.',
+      hide: true,
+      valueHelp: 'opt1,opt2,...',
+    )
+    ..addMultiOption(
       'extra-gen-snapshot-options',
       help: 'Pass additional options to gen_snapshot.',
       hide: true,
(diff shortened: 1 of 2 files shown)


mit-mit:
Hi Simon, thanks, but `dart compile` does not call `dart2native` (but it does use a library that they share). `dart2native` is about to get deprecated. Could we please add the flag to `dart compile` instead?

slightfoot:
@mit-mit I'll apply the patch to dart compile, if it still needs it.

mit-mit:
Thanks! Can you open a new PR for that, please?

PR #44758, opened 2021-01-23
Make compute assists async and fix message

### LINKED ISSUE #44730: AssistContributor.computeAssists is sync while ChangeBuilder.addGenericFileEdit needs to be awaited
A simple assist contributor looks like this. 
```dart
class MyAssistContributor extends AssistContributorMixin implements AssistContributor {
  DartAssistRequest request;

  @override
  AssistCollector collector;

  AnalysisSession get session => request.result.session;

  @override
  void computeAssists(DartAssistRequest request, AssistCollector collector) {
    this.request = request;
    this.collector = collector;

    final visitor = _MaterialImportsVisitor(this);
    request.result.unit.accept(visitor);

    final builder = ChangeBuilder(session: session);
    /* await */ builder.addGenericFileEdit(request.result.path, (builder) {
      builder.addSimpleInsertion(0, "Some new code");
    });

    addAssist(const AssistKind('material_to_widgets', 100, "Convert material import to widgets"), builder);
  }
}
```
The `ChangeBuilder` `builder` is immediately used by `addAssist`, however, the `builder.addGenericFileEdit(...)` is async and must be awaited before the `builder`'s `builder.sourceChange` property contains the change. With `computeAssists` being sync, that is not possible.

Found in analyzer_plugin versions 0.4.0 and 0.3.0.

### LINKED ISSUE #44731: AssistContributorMixin.addAssist ignores the message
When used as:
```dart
addAssist(const AssistKind('some_code', 100, "Some message"), builder);
```
the message "Some message" doesn't end up in the IDE.

I think the problem is that, because `builder.sourceChange` creates a new object on every invocation, the message is set on a local object, while a new object is created when the assist is added to the collector.
See: https://github.com/dart-lang/sdk/blob/5c174c8d8c7ccbbbd0678f3f4d73ea3654fafa5a/pkg/analyzer_plugin/lib/utilities/assist/assist_contributor_mixin.dart#L28-L30

spkersten:
Fixes #44730.
Fixes #44731.

For "fixes" there were similar bugs that I've fixed as well.

### DIFF
--- a/pkg/analyzer_plugin/doc/tutorial/assists.md
+++ b/pkg/analyzer_plugin/doc/tutorial/assists.md
@@ -62,7 +62,7 @@ class MyAssistContributor extends Object
     with AssistContributorMixin
     implements AssistContributor {
   static AssistKind wrapInIf =
-      new AssistKind('wrapInIf', 100, "Wrap in an 'if' statement");
+      AssistKind('wrapInIf', 100, "Wrap in an 'if' statement");
 
   DartAssistRequest request;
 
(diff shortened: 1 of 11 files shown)


bwilkerson:
Thanks for the PR! It look great.

As you might be aware, PRs for the sdk repository are automatically copied into a different system that we use named Gerrit, and that's where we do reviews. The link to the CL in Gerrit is https://dart-review.googlesource.com/c/sdk/+/180740.

PR #44808, opened 2021-01-29
adds current stack size getter to process info

chunhtai:
We would like to add a api to retrieve the stack size, This will be used in flutter to make sure we do not accidentally increase our stack size of rendering pipeline due to code refactoring

related: https://github.com/flutter/flutter/issues/73731


chunhtai:
Hi @zanderso , do you know who is the appropriate person to review this pr?

mraleph:
Any specific reason to make this a Dart API? If this is just for testing then you could just call a C function from a helper shared object through dart:ffi.

chunhtai:
we could do that, we were discussing whether it the right place to put such api. Alternative we can make it sit in dart:ui which is owned by flutter. @zanderso any thought?

mraleph:
I’d encourage you to not put this as API into any dart:* (including ui)  library because those API are not entirely treeshakable - even if Dart declaration is removed by Dart threeshaker, C++ code will stay behind. No reason to pollute the engine binary if it is test only API which can be implemented purely in framework code

chunhtai:
I am not too familiar with the dart ffi, but do I have access the dart sdk header like vm/os_thread.h in dart ffi?

chunhtai:
This usage: https://github.com/flutter/flutter/pull/75039/files#r567168897

mraleph:
@zanderso I am not sure if you actually need stack base. In @chunhtai's example we compute the difference so it should be enough to just get a hold of stack pointer itself. 

Strictly speaking if you want to support limited amount platforms (e.g. ARM64/ARM Android release build), then you can avoid adding any C++ code by writing something like this (tested to work on Android ARM64/ARM32 builds and desktop MacOS x64):

```dart
import 'dart:io';
import 'dart:ffi';

// void *mmap(void *addr, size_t len, int prot, int flags, int fd, off_t offset);
typedef c_mmap = Pointer<Void> Function(
    Pointer<Void>, IntPtr, Int32, Int32, Int32, IntPtr);
typedef dart_mmap = Pointer<Void> Function(
    Pointer<Void>, int, int, int, int, int);
final mmap = DynamicLibrary.process().lookupFunction<c_mmap, dart_mmap>("mmap");

// int mprotect(void *addr, size_t len, int prot);
typedef c_mprotect = Int32 Function(Pointer<Void>, IntPtr, Int32);
typedef dart_mprotect = int Function(Pointer<Void>, int, int);
final mprotect = DynamicLibrary.process()
    .lookupFunction<c_mprotect, dart_mprotect>("mprotect");

const protRead = 1;
const protWrite = 2;
const protExec = 4;

const mapPrivate = 0x02;
final mapJit = Platform.isMacOS ? 0x0800 : 0x0;
final mapAnon = Platform.isMacOS ? 0x1000 : 0x20;

String get currentArch {
  // Can probably use uname from libc for better detection logic. For now
  // just assume that we only expect to run this test on native hardware.
  if (Platform.isMacOS || Platform.isLinux || Platform.isWindows) {
    return 'x64'; // TODO: need to support Apple Silicon Macs as well
  } else if (Platform.isAndroid) {
    // Ignore x86 Android.
    return sizeOf<IntPtr>() == 4 ? 'arm' : 'arm64';
  }
  throw 'Failed to detect current arch';
}

final getStackPointer = () {
  final region = mmap(nullptr, 4096, protRead | protWrite,
      mapPrivate | mapAnon | mapJit, -1, 0);
  region.cast<Uint8>().asTypedList(4096).setAll(
      0,
      const {
        // X64: mov rax, rsp; ret;
        'x64': [0x48, 0x89, 0xe0, 0xc3],
        // ARM64: mov x0, sp; ret;
        'arm64': [0xe0, 0x03, 0x00, 0x91, 0xc0, 0x03, 0x5f, 0xd6],
        // ARM32: mov r0, sp; ret;
        'arm': [0x0d, 0x00, 0xa0, 0xe1, 0x1e, 0xff, 0x2f, 0xe1],
      }[currentArch]!);
  if (mprotect(region, 4096, protRead | protExec) != 0) {
    throw "Failed to mark code as executable";
  }
  return region
      .cast<NativeFunction<IntPtr Function()>>()
      .asFunction<int Function()>();
}();

@pragma('vm:never-inline')
int someOtherFunction([int level = 10]) {
  return level > 0 ? someOtherFunction(level - 1) : getStackPointer();
}

void main() {
  final base = getStackPointer();
  final inOtherFunction = someOtherFunction();
  print((base - inOtherFunction).toRadixString(16));
}
```

chunhtai:
Thanks @mraleph, I will give a try!

zanderso:
@mraleph neat!

I understand that the stack allocation/layout strategy has not changed for many years in the VM, but it occurs to me that the approach you've shared may not work if the stack used to handle an async message or event is ever different from the stack that was used to run `main()`. Is that something that we should be worried about?

mraleph:
@zanderso 

> I understand that the stack allocation/layout strategy has not changed for many years in the VM, but it occurs to me that the approach you've shared may not work if the stack used to handle an async message or event is ever different from the stack that was used to run main(). Is that something that we should be worried about?

This is an interesting and valid consideration, but I don't think we should worry about it, again, based on on @chunhtai example of how this API is going to be used: 


```
class ParentWidget extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    final int myStackSize = getStackPointer();//io.ProcessInfo.currentStackSize;
    return ChildWidget(parentStackSize: myStackSize);
  }
}

class ChildWidget extends StatelessWidget {
  const ChildWidget({this.parentStackSize, Key key}) : super(key: key);
  final int parentStackSize;

  @override
  Widget build(BuildContext context) {
    final int myStackSize = getStackPointer(); //io.ProcessInfo.currentStackSize;
    // Captures the stack size difference between parent widget and child widget
    // during the rendering pipeline, i.e. one layer of stateless widget.
    // print('parentStackSize: $parentStackSize, myStackSize $myStackSize');
    return Text(
      '${parentStackSize - myStackSize}',
      key: const ValueKey<String>(kStackSizeKey),
    );
  }
}
```

The stack is measured in two `build` methods (parent and child) which, I think, are expected to execute synchronously (and thus on the same stack), so no async should be involved. 

In any case if VM decides to shift to multiple thread stacks per isolate (or two scheduling UI isolate on different threads) we will have this communicated - and a benchmark will catch it because it will start reporting very strange numbers (huge negative or huge positive sizes).

chunhtai:
Hi @mraleph.
This is tangential. I have been looking at the code. The code is trying to add an assembly code block into the memory and runs it. Will this have security concern if developer can also insert random assembly codes through dart ffi and access the data in the vm from there?

mraleph:
@chunhtai yes, in the environments where Dart VM is expected to serve as a container/isolation boundary for untrusted third party code. We would disable `dart:ffi` in such environments. In general purpose Flutter development Dart VM however does not actually serve such purpose - user is already free to include arbitrary native code into their app and through that do whatever they want (e.g. meddle with VM internals if they want), so `dart:ffi` does not actually give them more power than they already have.

PR #45570, opened 2021-04-02
Make fields non-nullable in vm_service

xvrh:
Change the Dart generator of vm_service to generate all non-optional fields as non-nullable. This allow to remove a lot of `!` in the code.

### DIFF
--- a/pkg/vm_service/example/vm_service_assert.dart
+++ b/pkg/vm_service/example/vm_service_assert.dart
@@ -245,18 +245,18 @@ String assertStepOption(String obj) {
 
 vms.AllocationProfile assertAllocationProfile(vms.AllocationProfile obj) {
   assertNotNull(obj);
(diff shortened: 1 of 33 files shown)


mraleph:
Gerrit review https://dart-review.googlesource.com/c/sdk/+/193892 forwarded to @bkonyi

bkonyi:
Thanks for the PR @xvrh, but unfortunately we won't be able to accept this. We originally generated a similar interface to the one in this PR (see fb5ffc827580ae49cd4611d3ddf0ad3c7c592c54, which was undone in e3c0b0d3f7e4cae683c9417fbb2f944da5732e63), but `package:vm_service` isn't currently pinned to a particular VM service version, so it's possible for an older target VM service to respond with objects missing fields that are non-optional in a newer version of the specification.

PR #45651, opened 2021-04-10
Enable fix for always_use_package_imports

Kavantix:
Currently `dart fix` can do package to relative import but not the other way around.
This PR enabled the other direction by simply using the already available fix.


bwilkerson:
The review for this is on https://dart-review.googlesource.com/c/sdk/+/194940.

Kavantix:
As you pointed out this is not the correct code to change and there are some edgecases, I’m closing this for now

PR #45740, opened 2021-04-17
[CFE] add compile-time error when deriving a mixin from a class with a non-Object superclass

### LINKED ISSUE #45194: [cfe] The front end allows deriving a mixin from a class with a non-Object superclass
Cf. https://github.com/dart-lang/sdk/issues/45193.

lfkdsk:
resolve https://github.com/dart-lang/sdk/issues/45194

test result: 

```
../../a.dart:15:7: Error: The class 'Mixed' can't be used as a mixin because it extends a class other than Object.
class Mixed = Object with N, M;
      ^^^^^
../../a.dart:17:7: Error: The non-abstract class 'A' is missing implementations for these members:
 - N.text
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.

class A with Mixed {}
      ^
../../a.dart:12:14: Context: 'N.text' is defined here.
  String get text => 'Foo';
             ^^^^
```

### DIFF
--- a/pkg/_fe_analyzer_shared/lib/src/messages/codes_generated.dart
+++ b/pkg/_fe_analyzer_shared/lib/src/messages/codes_generated.dart
@@ -6804,6 +6804,31 @@ const MessageCode messageMixinFunction = const MessageCode("MixinFunction",
 // DO NOT EDIT. THIS FILE IS GENERATED. SEE TOP OF FILE.
 const Template<
     Message Function(
+        String
+            name)> templateMixinInheritsNotFromObject = const Template<
+        Message Function(String name)>(
+    messageTemplate:
+        r"""The class '#name' can't be used as a mixin because it extends a class other than Object.""",
+    withArguments: _withArgumentsMixinIneritsNotFromObject);
+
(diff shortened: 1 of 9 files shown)


mraleph:
I assume this is stale. Let me know if it is not.

sgrekhov:
@mraleph there is no error in CFE for the following code

```dart
class S {
  int get g1 => 1;
}

class M extends S {
  int get g1 => 2;
}

class A extends S {
  int get g1 => 3;
}

class C extends A with M {
//                     ^
// [analyzer] unspecified
// [cfe] unspecified
}

main() {
  new C();
}
```
There is expected compile error in Analyzer but no error in CFE

Tested on `Dart SDK version: 2.14.0-256.0.dev (dev) (Tue Jun 29 11:20:03 2021 -0700) on "windows_x64" `

mraleph:
@sgrekhov thanks for checking! the issue is still there (#45194 is still open), however PR seems stale unfortunately.

Issue #45672, opened 2021-04-13
[analyzer] Constant evaluation does not detect null related error


eernstg:
Consider the following program:

```dart
const dynamic d = null;
const xs = <int>[d];

void main() {
  print(xs);
}
```

This program is accepted by the analyzer (97742f827a207454102da154ec3befe29b624e80) with no errors (and no other diagnostic messages).

However, the program is purely null safe and should be assumed to run with sound null checking, and hence the `int` type argument should be considered to denote the type `int` (not `int*`). Cf. [this section](https://github.com/dart-lang/language/blob/master/accepted/2.12/nnbd/feature-specification.md#constant-instances):

> With sound null checking, all generic const constructors and generic const literals
> are evaluated using the type arguments provided

lfkdsk:
Hi, I try to fix this issue with https://github.com/dart-lang/sdk/pull/45750

scheglov:
https://dart-review.googlesource.com/c/sdk/+/196042

PR #45750, opened 2021-04-19
[analyzer] add constant evaluation check related collection type error.

lfkdsk:
try to fix this issue : https://github.com/dart-lang/sdk/issues/45672


mraleph:
I assume this is stale. Let me know if it is not.

PR #45752, opened 2021-04-19
[analyzer] add getter/setter conflict search in lexical scope lookup.

### LINKED ISSUE #45467: [analyzer] A getter/setter conflict is not detected
Consider the following program:

```dart
set m(_) {}

class A {
  int get m => 0;
}

class B extends A {
  void userOfM() => print(m);
}

void main() {}
```

This program is accepted by the analyzer (commit 5be4e20b16c15fcb6213fada575139e91f5efa11) with no errors (but the CFE reports `Getter not found: m`).

However, the error should be reported: The lookup for `m` should search enclosing lexical scopes (the syntax that actually encloses that occurrence of `m`), searching for the basename `m` (so it searches for both declarations named `m` and `m=`), and then it should find the top-level setter, and raise a compile-time error because there is no corresponding top-level getter.

lfkdsk:
resolve https://github.com/dart-lang/sdk/issues/45467


mraleph:
For now I assume this is stale. Let me know if it is not.

PR #45790, opened 2021-04-21
[analyzer] change type alias cycle focus on typedef name.

### LINKED ISSUE #45786: Analyzer detects error line number and position in the source code incorrectly
Dart SDK version: 2.14.0-3.0.dev (dev) (Tue Apr 13 22:57:19 2021 -0700) on "windows_x64"

The following source code example throws a compile time error for the line 5 because type alias cannot reference itself:
``` dart
/**
 * This is a test
 */

typedef FUNC1<T> = void Function<T1 extends FUNC1<T>>();

main() {}
```
The problems is that in this particular case analyzer ignores comments when calculate line number in the code.

So, dart prints error diagnostc for the line 5 whereas analyzer do the same for the line 1.

Sample output is:
```
$> dart test.dart
test.dart:5:9: Error: The typedef 'FUNC1' has a reference to itself.
typedef FUNC1<T> = void Function<T1 extends FUNC1<T>>();
        ^^^^^

$> dartanalyzer test.dart
Analyzing test.dart...
  error - Typedefs can't reference themselves directly or recursively via another typedef. - test.dart:1:1 - type_alias_cannot_reference_itself
1 error found.
```

Analyzer should refer to `test.dart:5:9` instead of `test.dart:1:1` here.

If I remove comments in the top of the source and just leave empty lines before the typedef definition, line number is calculated correctly.

lfkdsk:
try to fix : https://github.com/dart-lang/sdk/issues/45786

### DIFF
--- a/pkg/analyzer/lib/src/generated/error_verifier.dart
+++ b/pkg/analyzer/lib/src/generated/error_verifier.dart
@@ -3962,13 +3962,13 @@ class ErrorVerifier extends RecursiveAstVisitor<void>
   ///
   /// See [CompileTimeErrorCode.TYPE_ALIAS_CANNOT_REFERENCE_ITSELF].
   void _checkForTypeAliasCannotReferenceItself(
-    AstNode node,
+    TypeAlias node,
     TypeAliasElementImpl element,
   ) {
     if (element.hasSelfReference) {
       errorReporter.reportErrorForNode(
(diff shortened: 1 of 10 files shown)


mraleph:
I see that upstream review is +1 but can't land due to bot redness. 

For now I assume this is stale. Let me know if it is not.

PR #45808, opened 2021-04-23
[analyzer] bugfix: never's type should be an type-type, but not neverType or dynamicType.

### LINKED ISSUE #45399: Why is `Never` allowed by the analyzer pretty much anywhere?
Dartpad in null safe mode, and the Analyzer in Dart 2.13.0-137.0.dev, both allow code like `var x = 1 + 2 + Never;`.

When it runs, there's an error like "Error: A value of type 'Type' can't be assigned to a variable of type 'num'."

Consider this code:

```dart
void main() {
  1 + 2 + String;
  1 + 2 + Object;
  1 + 2 + dynamic;
  1 + 2 + void;
  1 + 2 + Null;
  1 + 2 + Never;
}
```

The analyzer catches each of these, except `Never`.

image

`Never` is a Type object, so why is it being analyzed like it's a value of type `Never` ?

I came across this with some code where the programmer was trying to use Never with streams where we care about the fact that a stream event occurred, but not about its value. In null safe Dart, adding `null` to a `Stream<String>` doesn't work. But, adding `Never` to it seems to... at least until you run it. This doesn't happen with any other Type as far as I can see.

Perhaps this is intentional for some esoteric reason? Or perhaps it's a mistake and `Never` is being treated like "instance of Never"?

image

lfkdsk:
try to fix: https://github.com/dart-lang/sdk/issues/45399

### DIFF
--- a/pkg/analyzer/lib/src/dart/resolver/simple_identifier_resolver.dart
+++ b/pkg/analyzer/lib/src/dart/resolver/simple_identifier_resolver.dart
@@ -254,7 +254,6 @@ class SimpleIdentifierResolver {
     } else if (element is DynamicElementImpl) {
       staticType = _typeProvider.typeType;
     } else if (element is NeverElementImpl) {
-      // NeverType's Type should be typeType, not neverType.
       staticType = _typeProvider.typeType;
     } else {
       staticType = DynamicTypeImpl.instance;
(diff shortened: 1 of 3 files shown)


vsmenon:
Thanks for the patch.  Note that we use gerrit instead of PRs for sdk work though:

https://github.com/dart-lang/sdk/blob/master/CONTRIBUTING.md

FYI - @srawlins @scheglov

PR #45937, opened 2021-05-06
Document same-library access for protected members

creativecreatorormaybenot:
I noticed that in the Flutter framework, `@protected` members are called from outside of instance methods. @Hixie responded:

> protected allows access from the same file

I think this should be documented.


creativecreatorormaybenot:
Gerrit URL: https://dart-review.googlesource.com/c/sdk/+/198480

PR #45947, opened 2021-05-07
Add access-control-allow-* headers

kevin-sakemaer:
cc @kevmoo if you know who can handle this.

### DIFF
--- a/sdk/lib/_http/http.dart
+++ b/sdk/lib/_http/http.dart
@@ -403,7 +403,6 @@ abstract class HttpHeaders {
       'access-control-request-headers';
   static const String accessControlRequestMethodHeader =
       'access-control-request-method';
-
   static const ageHeader = "age";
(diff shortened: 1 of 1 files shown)


kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/198780

@mit-mit – I'm ambivalent. Completeness is nice, but I feel it's a loosing game to try to add more of these to `dart:io`

bkonyi:
I've got no strong opinion one way or another, so LGTM, but let's get input from @mit-mit or @lrhn before landing?

lrhn:
It's essentially cost-free to add constants if they aren't used.
If someone needs them, I have no problem adding them. Completeness is not a goal, but adequacy is.

mraleph:
@lrhn @bkonyi can somebody plus +1 and +2 Gerrit CL?

PR #45976, opened 2021-05-10
Fix Socket.setOption failing for sockets with addresses of type InternetAddressType.unix

### LINKED ISSUE #45975: Socket.setOption fails for sockets with addresses of type InternetAddressType.unix
If you attempt to set a a socket option when using a Unix socket the following is printed on the console:
```
Value outside expected range
```

This is due to the following code, which was written when Unix sockets didn't exist.
```c
  int64_t protocol = DartUtils::GetInt64ValueCheckRange(
      Dart_GetNativeArgument(args, 2), SocketAddress::TYPE_IPV4,
      SocketAddress::TYPE_IPV6);
```

robert-ancell:
The existing code only allowed InternetAddressType.IPv4 and InternetAddressType.IPv6, which were the only two types at the time.

Don't limit the allowed values, which matches the same behaviour for Socket.getOption.

Fixes https://github.com/dart-lang/sdk/issues/45975

### DIFF
--- a/runtime/bin/socket.cc
+++ b/runtime/bin/socket.cc
@@ -1077,9 +1077,8 @@ void FUNCTION_NAME(Socket_SetOption)(Dart_NativeArguments args) {
   Socket* socket =
       Socket::GetSocketIdNativeField(Dart_GetNativeArgument(args, 0));
   int64_t option = DartUtils::GetIntegerValue(Dart_GetNativeArgument(args, 1));
-  int64_t protocol = DartUtils::GetInt64ValueCheckRange(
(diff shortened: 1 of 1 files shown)


mraleph:
Thanks for the contribution. Gerrit review at https://dart-review.googlesource.com/c/sdk/+/199121 

Please add some tests. Location would be `./tests/standalone/io/*_test.dart`. You can take a look at how other tests in the same folder are structured for guidance.

a-siva:
Thanks for your contribution @robert-ancell I have included your change along with some test cases and a small change in HttpServer to not set the TCP_NODELAY option for Unix sockets
https://dart-review.googlesource.com/c/sdk/+/199640

PR #45978, opened 2021-05-10
Don't set TCP_NO_DELAY socket option for Unix sockets.

### LINKED ISSUE #45977: HttpServer fails to work with Unix sockets
Creating an HttpServer with a Unix socket fails to accept connection, due to attempting to set the TCP_NO_DELAY socket option on the sockets, which is not supported for this socket type.

e.g.
```dart
var server = await HttpServer.bind(InternetAddress('/tmp/test-socket', type: InternetAddressType.unix), 0);
server.listen((request) {
  print(request.method);
});
```

Connecting using an HTTP client...
```shell
curl --unix-socket /tmp/test-socket http://localhost/
```

...causes the server to hit an exception:
```
Unhandled exception:
OS Error: Operation not supported, errno = 95
```

Note that this exception only occurs with a [fix to Socket.setOption](https://github.com/dart-lang/sdk/issues/45975), since it currently hits an assertion with Unix sockets.

robert-ancell:
This is not supported for this socket type and will generate a "Operation not supported" exception.

Fixes https://github.com/dart-lang/sdk/issues/45977

### DIFF
--- a/sdk/lib/_http/http_impl.dart
+++ b/sdk/lib/_http/http_impl.dart
@@ -2442,7 +2442,9 @@ class _ConnectionTarget {
       }
       return socketFuture.then((socket) {
         _connecting--;
-        socket.setOption(SocketOption.tcpNoDelay, true);
+        if (socket.address.type != InternetAddressType.unix) {
+          socket.setOption(SocketOption.tcpNoDelay, true);
+        }
         var connection =
             new _HttpClientConnection(key, socket, client, false, context);
         if (isSecure && !proxy.isDirect) {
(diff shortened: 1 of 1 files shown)


mraleph:
Thank you for the contribution. Gerrit review is at https://dart-review.googlesource.com/c/sdk/+/199122

Please add a regression test. Location for a test would be `./tests/standalone/io/http_*_test.dart`, you can look at how other tests in the same folder are structured.

a-siva:
Thank you for your contribution @robert-ancell I have included your changes along with some test cases here https://dart-review.googlesource.com/c/sdk/+/199640

Issue #34284, opened 2018-08-28
dart:io WebSocket client cannot connect to WebSocket server with pinned, self-signed certificate


CryptUser:
I have a WebSocket server at ip address 192.168.0.11 with external port 9000, and a self-signed certificate.
The server is tested to work ok with an ios client using SocketRocket.
When I try to connect with my flutter app using

socket = await WebSocket.connect('wss://192.168.0.11:9000');

I get the following error message:
[VERBOSE-2:dart_error.cc(16)] Unhandled exception:
HandshakeException: Handshake error in client (OS Error: 
	CERTIFICATE_VERIFY_FAILED: ok(handshake.cc:363))
#0      _WebsocketPlaygroundHomeState.connectWebsocket (package:quano_flutter/ui/debug/websocket_playground.dart:92:14)
<asynchronous suspension>
#1      _WebsocketPlaygroundHomeState.build.<anonymous closure> (package:quano_flutter/ui/debug/websocket_playground.dart:138:19)
#2      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:494:14)
#3      _InkResponseState.build.<anonymous closure> (package:flutter/src/material/ink_well.dart:549:30)
#4      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:102:24)
#5      TapGestureRecognizer._checkUp (package:flutter/src/gestures/tap.dart:161:9)
#6      TapGestureRecognizer.handlePrimaryPointer (package:flutter/src/gestures/tap.dart:94:7)
#7      PrimaryPointerGestureRecognizer.handleEvent (package:flutter/src/gestures/recognizer.dart:315:9)
#8      <…>

Note that I cannot just replace the server certificate with one signed by a CA: The reason I want to use a pinned, self-signed certificate is security. The public part of the certificate will be saved in the app, and only connections to a server with the matching certificate will be allowed, thus preventing man-in-the-middle attacks. Pinning a self-signed certificate has the advantage, that we as app developers do not need to trust any CA.

My idea then was to make the app aware of the certificate using a SecurityContext, but WebSocket.connect(...) does not seem to take a SecurityContext. Having a String "cert" that contains the certificate, I instead tried the following code that makes use of the fromUpgradedSocket constructor:

    var securityContext = SecurityContext();
    var bytesList = utf8.encode(cert);
    var bytes = bytesList is Uint8List ? bytesList : Uint8List.fromList(bytesList);
[...]

PR #46037, opened 2021-05-17
Exposing http client websocket

rnewquist:
Solution for certificate pinning for web sockets.
Solves [#34284](https://github.com/dart-lang/sdk/issues/34284)


rnewquist:
@googlebot I consent

PR #46040, opened 2021-05-17
[io/http] Add a HTTP Client parameter on WebSocket.connect to allow a custom HTTP Client for web socket connections.

rnewquist:
The WebSocket abstract class was changed to allow an optional parameter called customClient that takes in a HTTPClient and passes it to the WebSocket Implementation.
The WebSocket implementation takes the customClient, checks if its null, if its not null, it uses the customClient in place of the static HTTPClient that the WebSocket Implementation offers.
This custom client does not override the static HTTPClient, so all previous functionality remains the same when the customClient is not present.

TEST=testStaticClientUserAgentStaysTheSame() in web_socket_test.dart in standalone_2/standalone
TEST=new SecurityConfiguration(secure: true).runTests(); in web_socket_error_test.dart and web_socket_test.dart in standalone_2/standalone

Bug: https://github.com/dart-lang/sdk/issues/34284

### DIFF
--- a/sdk/lib/_http/websocket_impl.dart
+++ b/sdk/lib/_http/websocket_impl.dart
@@ -995,16 +995,12 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
   Timer? _closeTimer;
   _WebSocketPerMessageDeflate? _deflate;
 
-  static HttpClient _httpClient = new HttpClient();
+  static final HttpClient _httpClient = new HttpClient();
 
   static Future<WebSocket> connect(
       String url, Iterable<String>? protocols, Map<String, dynamic>? headers,
       {CompressionOptions compression = CompressionOptions.compressionDefault,
-      HttpClient customClient}) {
-    //overriding _httpClient if user specifies a custom http client.
-    if (customClient != null) {
-      _httpClient = customClient..userAgent ??= _httpClient.userAgent;
-    }
+      HttpClient? customClient}) {
     Uri uri = Uri.parse(url);
     if (uri.scheme != "ws" && uri.scheme != "wss") {
       throw new WebSocketException("Unsupported URL scheme '${uri.scheme}'");
@@ -1026,7 +1022,9 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
         path: uri.path,
         query: uri.query,
         fragment: uri.fragment);
-    return _httpClient.openUrl("GET", uri).then((request) {
+        // Use Custom HTTP Client first is it exists, else
+        // default to static client
+    return (customClient ?? _httpClient).openUrl("GET", uri).then((request) {
       if (uri.userInfo != null && !uri.userInfo.isEmpty) {
         // If the URL contains user information use that for basic
         // authorization.
--- a/sdk/lib/_http/websocket_impl.dart
+++ b/sdk/lib/_http/websocket_impl.dart
@@ -995,15 +995,15 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
   Timer? _closeTimer;
   _WebSocketPerMessageDeflate? _deflate;
 
-  static final HttpClient _httpClient = new HttpClient();
+  static HttpClient _httpClient = new HttpClient();
 
   static Future<WebSocket> connect(
       String url, Iterable<String>? protocols, Map<String, dynamic>? headers,
-      {CompressionOptions compression =
-          CompressionOptions.compressionDefault}) {
-    Uri uri = Uri.parse(url);
-    if (uri.scheme != "ws" && uri.scheme != "wss") {
-      throw new WebSocketException("Unsupported URL scheme '${uri.scheme}'");
+      {CompressionOptions compression = CompressionOptions.compressionDefault,
+      HttpClient customClient}) {
+    //overriding _httpClient if user specifies a custom http client.
+    if (customClient != null) {
+      _httpClient = customClient..userAgent ??= _httpClient.userAgent;
     }
 
     Random random = new Random();
--- a/sdk/lib/_http/websocket_impl.dart
+++ b/sdk/lib/_http/websocket_impl.dart
@@ -1022,8 +1022,6 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
         path: uri.path,
         query: uri.query,
         fragment: uri.fragment);
-        // Use Custom HTTP Client first is it exists, else
-        // default to static client
     return (customClient ?? _httpClient).openUrl("GET", uri).then((request) {
       if (uri.userInfo != null && !uri.userInfo.isEmpty) {
         // If the URL contains user information use that for basic
--- a/sdk/lib/_http/websocket_impl.dart
+++ b/sdk/lib/_http/websocket_impl.dart
@@ -1005,6 +1005,10 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
     if (customClient != null) {
       _httpClient = customClient..userAgent ??= _httpClient.userAgent;
     }
+    Uri uri = Uri.parse(url);
+    if (uri.scheme != "ws" && uri.scheme != "wss") {
+      throw new WebSocketException("Unsupported URL scheme '${uri.scheme}'");
+    }
 
     Random random = new Random();
     // Generate 16 random bytes.
--- a/tests/standalone/io/web_socket_error_test.dart
+++ b/tests/standalone/io/web_socket_error_test.dart
@@ -48,9 +48,9 @@ class SecurityConfiguration {
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
   Future<WebSocket> createClient(int port) => secure
-      ? WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/')
-      : WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
-          customClient: HttpClient(context: clientContext));
+      ? WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
+          customClient: HttpClient(context: clientContext))
+      : WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
 
   void testForceCloseServerEnd(int totalConnections) {
     createServer().then((server) {
--- a/tests/standalone/io/web_socket_error_test.dart
+++ b/tests/standalone/io/web_socket_error_test.dart
@@ -47,9 +47,10 @@ class SecurityConfiguration {
       ? HttpServer.bindSecure(HOST_NAME, 0, serverContext, backlog: backlog)
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
-  Future<WebSocket> createClient(int port) =>
-      // TODO(whesse): Add a client context argument to WebSocket.connect.
-      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
+  Future<WebSocket> createClient(int port) => secure
+      ? WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/')
+      : WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
+          customClient: HttpClient(context: clientContext));
 
   void testForceCloseServerEnd(int totalConnections) {
     createServer().then((server) {
@@ -94,7 +95,6 @@ class SecurityConfiguration {
 main() {
   asyncStart();
   new SecurityConfiguration(secure: false).runTests();
-  // TODO(whesse): WebSocket.connect needs an optional context: parameter
-  // new SecurityConfiguration(secure: true).runTests();
+  new SecurityConfiguration(secure: true).runTests();
   asyncEnd();
 }
--- a/tests/standalone/io/web_socket_error_test.dart
+++ b/tests/standalone/io/web_socket_error_test.dart
@@ -47,10 +47,9 @@ class SecurityConfiguration {
       ? HttpServer.bindSecure(HOST_NAME, 0, serverContext, backlog: backlog)
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
-  Future<WebSocket> createClient(int port) => secure
-      ? WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
-          customClient: HttpClient(context: clientContext))
-      : WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
+  Future<WebSocket> createClient(int port) =>
+      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
+          customClient: secure ? HttpClient(context: clientContext) : null);
 
   void testForceCloseServerEnd(int totalConnections) {
     createServer().then((server) {
--- a/tests/standalone_2/io/web_socket_test.dart
+++ b/tests/standalone_2/io/web_socket_test.dart
@@ -46,9 +46,10 @@ class SecurityConfiguration {
       ? HttpServer.bindSecure(HOST_NAME, 0, serverContext, backlog: backlog)
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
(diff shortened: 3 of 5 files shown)


rnewquist:
@athomas it was such minor code, I just created a new PR.

rnewquist:
This fixes #34284 by opening up the http client used in the web socket code, allowing users to add a custom Security Context and cert failure callback to the web socket.  Currently, the only solution is to create a global http override, which isn't an ideal solution.

mit-mit:
Code review is https://dart-review.googlesource.com/c/sdk/+/200262

a-siva:
The signature of `connect` has been changed in the implementation but this PR does not seem to make the corresponding change in the '`abstract class WebSocket'` code in websocket.dart.
Another question, the code could have set a userAgent using the '`set userAgent`' method, now with the static variable `_httpClient` being changed to this new value, the getter userAgent would return whatever the new client has set. Not sure if this would be a problem, should the userAgent set previously be transferred to the new object ?

rnewquist:
@a-siva 
The abstract webSocket shouldn't need changing, it's only the connection part that needs to have the changes.  The abstract webSocket has the appropriate connections, it's literally the httpClient being a private final that is preventing any sort of certificate pinning.  
As for the second question, the user agent would get replaced if the new httpClient contains a user agent, however if it doesn't, the user agent gets set to what ever the private http client was set to.  
_httpClient = customClient..userAgent ??= _httpClient.userAgent;
I suppose another approach would be to allow the httpClient to have exposed getters and setters for the webSocket as well.  I'm not sure if that would cause an issue though as I have not tested that, but I can verify, at least for my use case, this method does work for certificate pinning.

sdk/lib/_http/websocket_impl.dart
```diff
@@ -995,12 +995,16 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
...
 
   static Future<WebSocket> connect(
       String url, Iterable<String>? protocols, Map<String, dynamic>? headers,
-      {CompressionOptions compression =
-          CompressionOptions.compressionDefault}) {
+      {CompressionOptions compression = CompressionOptions.compressionDefault,
+      HttpClient customClient}) {
+    //overriding _httpClient if user specifies a custom http client.
+    if (customClient != null) {
+      _httpClient = customClient..userAgent ??= _httpClient.userAgent;
```
aam, line 1006: I think overriding static `_httpClient` here will be confusing because it means that if you call `connect` twice with different customClients the first customClient will be overridden with second one. Currently it will result in `userAgent` for the fist connect reporting the value from customClient of the last connect.

rnewquist: I agree, however, with how its set up, it would require the httpClient to be tied to the webSocket itself (which I think it should be).  However that problem exists in the current implementation does it not?  So it doesn't solve that problem, but it does solve having a custom httpClient for all webSocket connections.

aam: > I agree, however, with how its set up, it would require the httpClient to be tied to the webSocket connection itself (which I think it should be). 

Yeah, I think having httpClient as an instance field of _WebSocketImpl would be cleaner. It could default to static final HttpClient that can be used if user didn't provide customClient.

> However that problem exists in the current implementation does it not?

Well, currently there is only one httpClient used by all web socket connections, so it being static is okay.

rnewquist: In that case, would it be best to create a separate ticket for instancing the httpClient with the webSocket?  That said, I think that it would be much more clear to the user to have getters and setters for the static httpClient instead of having it passed as a method parameter.  If you guys agree, i'll update this PR to reflect that.

rnewquist: If it wasn't clear by my proposal, having two separate tasks, one having the existing framework unchanged, but allowing the user to change the httpClient for all webSocket connections, and the other instancing the httpClient on a per webSocket connection basis.  Since the first is literally making a private variable accessible, it shouldn't have much, if any, technical debt on the second task.

aam: Main issue with exposing global writeable httpClient that is used for all websocket clients is that it will be breaking change(hard) to get rid of it.
Now I think that httpClient actually should not be an instance property of the websocket as it's only used to establish connection, only used during connect static method runtime. Also websocket can be created without httpClient at all: from regular upgraded socket.

The only question is about `userAgent` static property, which configures static(global) httpClient that is used when no customClient is specified in `connect` call.
I think it should be fine if this property only reads/updates only that default static global httpClient, doesn't affect customHttpClient user can specify in `connect`. This way the change doesn't affect existing code that updates `userAgent` property of static global httpClient and is hopefully intuitive that it won't affect custom httpClient that was provided in `connect` call.

a-siva: I still don't get how this change helps.
The class _WebSocketImpl is a private class and would be visible to code only within dart:io library, how will the
user code invoke the connect method with this new parameter.
Can we add a test case that demonstrates how this change will be effective.

aam: > The class _WebSocketImpl is a private class and would be visible to code only within dart:io library, how will the user code invoke the connect method with this new parameter.

Yeah, we definitely have to align `WebSocket` `connect` with `_WebSocketImpl`'s one.
With that one can fix existing test that has corresponding use case commented out:
```
diff --git a/tests/standalone/io/web_socket_error_test.dart b/tests/standalone/io/web_socket_error_test.dart
index a7fdd0ef7ca..75a8ea0c24b 100644
--- a/tests/standalone/io/web_socket_error_test.dart
+++ b/tests/standalone/io/web_socket_error_test.dart
@@ -47,9 +47,10 @@ class SecurityConfiguration {
       ? HttpServer.bindSecure(HOST_NAME, 0, serverContext, backlog: backlog)
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
-  Future<WebSocket> createClient(int port) =>
-      // TODO(whesse): Add a client context argument to WebSocket.connect.
-      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
+  Future<WebSocket> createClient(int port) {
+    return WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
+        customHttpClient: secure ? HttpClient(context: clientContext) : null);
+  }
 
   void testForceCloseServerEnd(int totalConnections) {
     createServer().then((server) {
@@ -94,7 +95,6 @@ class SecurityConfiguration {
 main() {
   asyncStart();
   new SecurityConfiguration(secure: false).runTests();
-  // TODO(whesse): WebSocket.connect needs an optional context: parameter
-  // new SecurityConfiguration(secure: true).runTests();
+  new SecurityConfiguration(secure: true).runTests();
   asyncEnd();
 }
```

aam: @rnewquist do you still have time to make the changes(use `customHttpClient` if it is specified, otherwise use static `httpClient_`, don't override `httpClient_` with `customHttpClient` so that static `userAgent` getter/setter only affect static `httpClient_`, add a test)?

rnewquist: Yeah, I can this weekend.

aam: thank you, @rnewquist. Have you looked at patching tests/standalone/io/web_socket_error_test.dart along the lines of the diff in my comment above?

aam: you want to keep tests in standalone/io and standalone_2/io in sync so if you updated tests/standalone/io/web_socket_test.dart, we should also update tests/standalone_2/io/web_socket_test.dart.

aam: also take a look at try-run failures on https://dart-review.googlesource.com/c/sdk/+/200262. It is automatically synchronized with this PR, hopefully once you fix those you can request another dry run with `CQ DRY RUN` button at top right corner of that webpage.

rnewquist: I fixed the standalone 2 issue, however I'm a little bit confused what you would like me to do on the web_socket_error_test.

rnewquist: @aam never mind, after re-reading the code, it was really clear what you wanted me to do.  it has since been fixed

aam: https://github.com/dart-lang/sdk/blob/main/tests/standalone/io/web_socket_error_test.dart#L97 has a TODO to enable that test so it works with secure connection. The quote in https://github.com/dart-lang/sdk/pull/46040#discussion_r682156221 is a change that can be applied to that test(lines prefixed with '-' are deleted, with '+' - added) so that is accomplished.

aam: Awesome, thanks

rnewquist: I do not see the option for CQ DRY RUN in the googlesource webpage unfortunately.

aam: @athomas do you know if this is by design where PR authors are not able to trigger dry runs in Gerrit?

athomas:
Yes, only Googlers can trigger tryjobs at this time.

sdk/lib/_http/websocket_impl.dart
aam: @rnewquist, I triggered the dry run.

Can you please update the commit message so it roughly follows the outline we try to adhere to for vm changes:

```[io/http] <One-liner that summarizes the change>

<Paragraph/few lines that describe what was changed, what was accomplished>

<Line that starts with TEST=... that lists tests that were used to verify correct behavior>

<Line that starts with Bug:... that references existing dart issue in github>```

Comment (rnewquist) [reply sdk/lib/_http/websocket_impl.dart]: On github, gerrit, or both?

Comment (rnewquist) [reply sdk/lib/_http/websocket_impl.dart]: @aam I am not authorized to change the message in gerrit, so I updated the github message.

Comment (aam) [reply sdk/lib/_http/websocket_impl.dart]: Thank you! Looking at how this ends up being copied into Gerrit by Copybara, I think `[io/http] <One-liner...` should be put into the subject of this PR(replacing `exposed http client for web sockets` subject) leaving only `<Paragarph... etc` in the description. 
Also word `TEST` in `Test=...` lines should be all capital and there should be no space around equal sign.

You can see how we attempt to follow this convention in https://github.com/dart-lang/sdk/commits/main

Comment (rnewquist) [reply sdk/lib/_http/websocket_impl.dart]: @aam fixed, also added secure tests to web_socket_test.dart in both standalones.

File: tests/standalone/io/web_socket_error_test.dart [hunk 2]
```diff
@@ -47,9 +47,10 @@ class SecurityConfiguration {
       ? HttpServer.bindSecure(HOST_NAME, 0, serverContext, backlog: backlog)
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
-  Future<WebSocket> createClient(int port) =>
-      // TODO(whesse): Add a client context argument to WebSocket.connect.
-      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
+  Future<WebSocket> createClient(int port) => secure
+      ? WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
+          customClient: HttpClient(context: clientContext))
+      : WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
```
aam, line 53: perhaps this can be simplified by passing `null` `customClient` if `secure` is `false`?
```
Future<WebSocket> createClient(int port) => 
      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
          customClient: secure ? HttpClient(context: clientContext): null);

Comment (rnewquist) [reply tests/standalone/io/web_socket_error_test.dart:53]: completely agree, i wanted to keep the same logic as the server but didn't notice they were calling different functions.  I have the customUserAgent there for the added test.

File: sdk/lib/_http/websocket_impl.dart (hunk 1 above)
Comment (athomas) [reply sdk/lib/_http/websocket_impl.dart]: FYI, if you start with a PR it's best to keep modifying the PR and just treat Gerrit as read-only (if you modify the review, that will likely break the automatic syncing from GitHub because of the conflicting edits).

Comment (rnewquist) [reply sdk/lib/_http/websocket_impl.dart]: @athomas I did not know that, sorry about that.  I don't think I have any permission to edit the gerrit, so I "think" it should good, however I will keep that in mind for any future changes that I do.

File: tests/standalone_2/io/web_socket_test.dart [hunk 3]
```diff
@@ -46,9 +46,11 @@ class SecurityConfiguration {
       ? HttpServer.bindSecure(HOST_NAME, 0, serverContext, backlog: backlog)
       : HttpServer.bind(HOST_NAME, 0, backlog: backlog);
 
-  Future<WebSocket> createClient(int port) =>
-      // TODO(whesse): Add client context argument to WebSocket.connect
-      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/');
+  Future<WebSocket> createClient(int port, {String customUserAgent}) =>
+      WebSocket.connect('${secure ? "wss" : "ws"}://$HOST_NAME:$port/',
+          customClient: secure
```
aam, line 51: Not sure if you are able to run the web_socket_test to troubleshoot the failure on Gerritt, hopefully you are.
Just to confirm - the issue here is that `testShouldSetUserAgent` when run with `SecurityConfiguration` `secure`:`true` fails due to the fact that global `WebSocket.userAgent` is not used in that setting - the connection is made using `customClient` with its `userAgent`.

rnewquist: I am not able to actually test it.  However, WebSocket.userAgent comes from a final static, unless something else is setting the user agent, it shouldn't it never get changed?  Would be a lot easier if I could debug it though.

aam: You could try following https://github.com/dart-lang/sdk/wiki/Building to get the sdk built locally, test running.
https://github.com/dart-lang/sdk/blob/main/tests/standalone/io/web_socket_test.dart#L550 from the test that [fails](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8836546572039956609/+/u/test_results/new_test_failures__logs_) is setting WebSocket.userAgent directly.

rnewquist: Took a while to get it up and running, but it was a simple fix.  Because of how the tests are ran and because WebSocket.userAgent comes from a static final, testShouldSetUserAgent and testStaticClientUserAgentStaysTheSame were setting the userAgent to different values at roughly the same time.  By just making sure that they are the same value, the static final variable doesn't get changed.

aam: Thank you, it looks like there are more failures due to use of WebSocket.connect() when `secure` is `true` but without customClient that is needed.

aam: I'm going to push a fix to the test to this PR if you don't mind @rnewquist.
It's basically about several places in the test where `WebSocket.connect` is invoked without provided `customClient` which is needed when running in secure configuration

sdk/lib/_http/websocket_impl.dart
```diff
@@ -1022,7 +1022,9 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
         path: uri.path,
         query: uri.query,
         fragment: uri.fragment);
-    return _httpClient.openUrl("GET", uri).then((request) {
+        // Use Custom HTTP Client first is it exists, else
```
aam, line 1025: I think this comment can be omitted as it just describes what is happening on the next line.
Ideally comments are answering the question "why?", rather than "what?", because normally code answers "what?" question already.

tests/standalone_2/io/web_socket_test.dart
rnewquist: @aam that would be perfect, I am a little busy this week, sorry about that.

aam:
thank you @rnewquist once again!

a-siva:
I noticed that the status files has a skip
io/web_socket_error_test: Skip # Flaky
for [ $compiler == dartkp && $runtime == dart_precompiled ]
The new tests added here do not run in precompiled mode, is that intended ?
It appears the fix for setting up the resources done in 
https://dart-review.googlesource.com/c/sdk/+/213524
does not work for configurations that have a two step process,
see https://github.com/dart-lang/sdk/issues/47239

PR #46132, opened 2021-05-25
expose onUploadProgress on HttpRequest dart:html

phamnhuvu-dev:
Context:
I am using HttpRequest in `dart:html` to upload file. 
All is ok when I use HttpRequest with `onProgress` on Flutter `2.0.5` but when I upgraded Flutter `2.2.0`, the `onProgress` is not called then I read the document and know the way use [xhr.upload.onprogress](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/upload) to get the upload progress and it works.

### DIFF
--- a/sdk/lib/html/dart2js/html_dart2js.dart
+++ b/sdk/lib/html/dart2js/html_dart2js.dart
@@ -18134,14 +18134,17 @@ class HttpRequest extends HttpRequestEventTarget {
    *
    * See also: [authorization headers](http://en.wikipedia.org/wiki/Basic_access_authentication).
    */
-  static Future<HttpRequest> request(String url,
-      {String? method,
-      bool? withCredentials,
-      String? responseType,
-      String? mimeType,
-      Map<String, String>? requestHeaders,
-      sendData,
-      void onProgress(ProgressEvent e)?}) {
+  static Future<HttpRequest> request(
+    String url, {
+    String? method,
+    bool? withCredentials,
+    String? responseType,
+    String? mimeType,
+    Map<String, String>? requestHeaders,
+    sendData,
+    void onProgress(ProgressEvent e)?,
+    void onUploadProgress(ProgressEvent e)?,
+  }) {
     var completer = new Completer<HttpRequest>();
 
     var xhr = new HttpRequest();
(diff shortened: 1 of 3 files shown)


sdk/lib/html/dart2js/html_dart2js.dart
```diff
@@ -18134,14 +18134,17 @@ class HttpRequest extends HttpRequestEventTarget {
...
    */
-  static Future<HttpRequest> request(String url,
-      {String? method,
-      bool? withCredentials,
-      String? responseType,
-      String? mimeType,
-      Map<String, String>? requestHeaders,
-      sendData,
-      void onProgress(ProgressEvent e)?}) {
+  static Future<HttpRequest> request(
```
srujzs, line 18137: Please modify `tools/dom/templates/html/impl/impl_XMLHttpRequest.darttemplate` instead of the file directly. You'll need to use `void onUploadProgress(ProgressEvent e)$NULLABLE` there instead of with `?`. These files were used to generate both legacy and null-safe versions of this library, hence the `$NULLABLE`.

`dart:html` is generated by using these templates, source files, and codegen scripts, so any edits in the file directly without modifying one of those will remove this code on regeneration. Once you've made your edits there you can run `go.sh` in `tools/dom/scripts` to generate the library.

srujzs (changes requested):
Thanks for contributing! I'm OK with this pull-request, it just needs to be added to the proper file.

sdk/lib/html/dart2js/html_dart2js.dart
phamnhuvu-dev: I run `./go.sh` then facing this error.

```
+ python2.7 ./dartdomgenerator.py --systems=htmldart2js --logging=40 --update-dom-metadata --gen-interop ''
Traceback (most recent call last):
  File "./dartdomgenerator.py", line 32, in <module>
    assert (os.path.exists(third_party_dir))
AssertionError
```

mraleph: Make sure you get SDK source code by following https://github.com/dart-lang/sdk/wiki/Building#getting-the-source - just cloning GitHub repository is not enough to make a proper development checkout.

mit-mit:
@phamnhuvu-dev is this still being worked on?

phamnhuvu-dev:
@mit-mit Sorry, I forgot this, I will do it tomorrow. Thanks for reminding me.

kevmoo:
Friendly ping!

mit-mit:
@phamnhuvu-dev any news? We'll close this as assumed stale if we don't hear back soon

mit-mit:
Closing as assumed stale.

PR #46153, opened 2021-05-26
Update bigint.dart

BerikaSuna:


mraleph:
This change does not make any sense and looks rather random. Could you clarify why you are doing it?

PR #46185, opened 2021-05-29
Develop

Rupesh-Darimisetti:


mraleph:
I kindly ask you to stop sending meaningless changes. 

Your change to `PRESUBMIT.py` breaks it - which you would have known if you followed proper contribution workflow and not taken a shortcut of sending a PR on GitHub (which are left open as a courtesy and not as a main way to develop Dart SDK). Changing code in legacy newsletters archive also does not make much sense. 

If you wish to make some contributions on consider instead *asking* first (eg look at the issue tracker)  - there might be places where help is wanted, rather then just doing changes that do not make sense.

Rupesh-Darimisetti:
@googlebot I fixed it.

PR #46321, opened 2021-06-10
added Null safety to jsonDecode for source var

yashodeepk:


yashodeepk:
needed CLA

sdk/lib/convert/json.dart
```diff
@@ -91,7 +91,7 @@ String jsonEncode(Object? object,
 ///
 /// Shorthand for `json.decode`. Useful if a local variable shadows the global
 /// [json] constant.
-dynamic jsonDecode(String source,
+dynamic jsonDecode(String? source,
```
mraleph, line 94: Why do you make this change? What is the purpose of it?

It is not making anything "null safe" (it makes it less null safe if anything) and it will not even compile because underlying `json.decode` expects `String`, not `String?`

yashodeepk: what if the source type is String ?

mraleph: then you need to check it for `null` one way or another. In JS you can `JSON.parse(null)` and get `null` back (which only works because `JSON.parse` starts by converting its argument to string), but in Dart you actually have to give `json.decode` a `String`. 

So if you want you can do `jsonDecode(input ?? "null")` in your code if your input is `String?`.

yashodeepk: Thanks for the help, i will use this

yashodeepk: One more thing why we have the same in jsonEncoder?

PR #46372, opened 2021-06-15
[doc] setTrustedCertificates doesn't replace certs

ktosiek:
With the previous wording, "Sets the set of ... certificates", I've assumed it always replaces everything in the store (at least on non-iOS).
This change should make it clear that the certificates are added.

(See also https://github.com/dart-lang/sdk/blob/bc2cf88bac44d02a44b9c315a8fd05eca2ecbaa4/runtime/bin/security_context.cc#L181 for the actual implementation).


ktosiek:
@googlebot I signed it!

PR #46481, opened 2021-06-28
Implement RawSocket.readMessage/writeMessage.

### LINKED ISSUE #46328: Support sending/receiving file descriptors over sockets
On Linux you are able to [send and receive file descriptors using Unix domain sockets](https://blog.cloudflare.com/know-your-scm_rights/). This is done using `sendmsg`/`recvmsg` instead of `sendto`/`recvfrom`. This is desirable when interfacing with services that make use of this functionality, for example DBus.

A proposed API would be something like:
```dart
var socketAddress = InternetAddress('/var/run/dbus/system_bus_socket', type: InternetAddressType.unix);
var socket = await RawSocket.connect(socketAddress, 0);
var data = [0];
var fd = getFileDescriptorFromSomewhere();
var message = SocketMessage([data], controlMessages: [SocketFileDescriptorMessage([fd])]);
socket.writeMessage(message);

var message = socket.readMessage();
for (var data in message.vectors) {
  processData(data);
}
for (var controlMessage in message.controlMessages) {
  if (message is SocketFileDescriptorMessage) {
    processFileDescriptors(message.fileDescriptors);
  }
}
```

This API is similar to the one [used in GLib](https://developer.gnome.org/gio/stable/GSocket.html#g-socket-send-message).

This new API would throw an exception on platforms that didn't support it.

robert-ancell:
Fixes https://github.com/dart-lang/sdk/issues/46328

Some notes on the implementation:
- A specific case where this feature will be used is in [D-Bus](https://github.com/canonical/dbus.dart/issues/245).
- This on `RawSocket` and not `Socket`, since when receiving a message you need to decide if using `recv` or `recvmsg`.
- I haven't done any additional wrapping of the file descriptors - they are just left as integers for the API user to interpret. This might either involve passing them to some FFI code, or using some [new Dart constructors that can take file descriptors](https://github.com/canonical/dbus.dart/issues/245).
- I've only done an implementation for Linux, though other platforms (Android?) may also make use of these.


mraleph:
Review will happen on the corresponding Gerrit CL https://dart-review.googlesource.com/c/sdk/+/205067

mit-mit:
Looks like the gerrit review for this is now https://dart-review.googlesource.com/c/sdk/+/212036

PR #46667, opened 2021-07-20
Add fuchsia.net.name.Lookup

tamird:
fuchsia.net.NameLookup is transitioning to fuchsia.net.name.Lookup.

@zanderso


zanderso:
@arbreng

zanderso:
Dart uses depot_tools/Gerrit, so this change might have to use that flow.

tamird:
I think copybara did it. https://dart-review.googlesource.com/c/sdk/+/207561

PR #46846, opened 2021-08-07
Upgrade fromEnvironment() docs to make clearer that constructor is only guaranteed to work when called as const

irvine5k:
This PR provides additional info on the `fromEnvironment()` constructors, as it will not work when the constructor is not constant on AOT-compiled platforms

Since `fromEnvironment()` constructors don't throw a warning when called as non-constant on AOT-compiled platforms, it makes confusing the use of dart define and lead doubts like https://twitter.com/prasadsunny1/status/1423721300149760000

### DIFF
--- a/sdk/lib/core/bool.dart
+++ b/sdk/lib/core/bool.dart
@@ -19,8 +19,6 @@ class bool {
   /// In all other cases, including when there is no declaration for `name`,
   /// the result is the [defaultValue].
   ///
-  /// The value must be assigned to a constant.
-  ///
   /// The result is the same as would be returned by:
   /// ```dart
   /// (const String.fromEnvironment(name) == "true")
@@ -43,6 +41,8 @@ class bool {
   /// must be consistent across all calls to [String.fromEnvironment],
   /// [int.fromEnvironment], `bool.fromEnvironment` and [bool.hasEnvironment]
   /// in a single program.
+  /// 
+  /// The bool.fromEnvironment() constructor must be constant.
   // The .fromEnvironment() constructors are special in that we do not want
   // users to call them using "new". We prohibit that by giving them bodies
   // that throw, even though const constructors are not allowed to have bodies.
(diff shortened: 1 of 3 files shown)


irvine5k:
@googlebot I signed it!

sdk/lib/core/bool.dart
```diff
@@ -19,6 +19,8 @@ class bool {
   /// In all other cases, including when there is no declaration for `name`,
   /// the result is the [defaultValue].
   ///
+  /// The value must be assigned to a constant.
```
lrhn, line 22: That's actually not correct. You can do:
```dart
 var davar = const String.fromEnvironment("daname", defaultValue: "dafault");
```

What is true is that the constructor invocation must be constant when used on the web (perhaps other ahead-of-time compiled targets too).
That's a restriction which was added by the web compilers (and semi-documented in line 44 below).
We should figure out whether we want that to be a general restriction (affecting the VM too), and then document whichever choice we make.

irvine5k: Oh, cool, I didn’t know that, thanks for the information, so, do you think that still worth updating the docs with this information?

irvine5k: Something like The .fromEnvironment() constructors must be constant.

irvine5k: @lrhn I updated the docs, do you think that's good enough now?

lrhn: It's correct and complete. 
I'm a little worried that it lacks context (AoT is an undefined abbreviation, but explaining it also seems like overkill. So, a matter of trade-off between readability and detail.) 
The documentation should apply to all four `Environment` functions.

I'll try writing something up.

irvine5k:
@lrhn continued my work, so I close this PR, if anyone want to see the result:
https://github.com/dart-lang/sdk/commit/020f21261ab531f56f6fddb5d2fe42a0f6928bbb

PR #46847, opened 2021-08-07
Fix isAfter method code example on DateTime class

niku:
The code example looks similar as [isBefore method](https://github.com/dart-lang/sdk/blob/b8d040e5239e02b945b529347d0b980fc8a26db2/sdk/lib/core/date_time.dart#L405-L417) except in this change.


lrhn:
Good catch. Thanks. 
I've accepted it in https://dart-review.googlesource.com/c/sdk/+/209460

PR #46859, opened 2021-08-09
Add FutureOrExtensions on FutureOr<T>

Alekseylv:
Adds a convenience method on FutureOr<T> to transform it's values similar to Future.then

### DIFF
--- a/sdk/lib/async/future.dart
+++ b/sdk/lib/async/future.dart
@@ -1037,3 +1037,22 @@ void _asyncCompleteWithErrorCallback(
   }
   result._asyncCompleteError(error, stackTrace);
 }
+
+
+/// Convenience methods on FutureOr<T>.
+///
+/// Adds functionality makes it easier to
+/// write well-typed asynchronous code.
+extension FutureOrExtensions<T> on FutureOr<T> {
+
+  /// Convenience method to transform FutureOr<T> values.
+  ///
+  /// Similar to [Future.then], but also handles values not
+  /// wrapped in a future.
+  FutureOr<R> then<R>(R Function(T) fun) {
+    if (this is Future<T>) {
+      return (this as Future<T>).then(fun);
+    }
+    return fun(this as T);
+  }
+}
--- a/sdk/lib/async/future.dart
+++ b/sdk/lib/async/future.dart
@@ -1041,7 +1041,7 @@ void _asyncCompleteWithErrorCallback(
 
 /// Convenience methods on FutureOr<T>.
 ///
-/// Adds functionality makes it easier to
+/// Adds functionality that makes it easier to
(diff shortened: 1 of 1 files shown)


Alekseylv:
@googlebot I signed it!

lrhn:
I'm not sure this is something we'd want to have, or recommend using. The operation to use for `FutureOr<T>` is `await`, or possibly `Future.value` to turn it into a certain `Future`.
Propagating the synchronicity of the `or` part is something we generally try to avoid. Having both a synchronous and asynchronous track in the same computation is risky and tends to encourage people to optimize for the synchronous path at the cost of readability and maintainability. So, for a function like this, I'd *want* it to return `Future<R>`.

In fact, it seems to assume that `R` cannot be a future, because if it was `Future<X>`, you'd return `FutureOr<Future<X>>` instead of just `Future<X>`. I'd rather prefer something more like:
```dart
Future<R> then<R>(FutureOr<R> Function(T) fun) async => await fun(await this);
```
but then it's fairly close to just doing `Future.value(futureOr).then(fun)`.

All in all, I think we'll reject this (otherwise fine) code because it's encouraging behavior that we don't actually want to encourage, and alternatives (`await` and `Future.value`) are readily available.

Alekseylv:
I think there are still benefits to have it here either as

```dart
Future<R> then<R>(FutureOr<R> Function(T) fun) async => await fun(await this);
```

or

```dart
Future<R> then<R>(FutureOr<R> Function(T) fun) => Future.value(this).then(fun);
```

because
a) Discoverability - it took me a while to figure out how to handle a FutureOr in a nice way, `sdk/lib/async/future.dart` was one of the places I looked at and I didn't find anything useful. Guiding users directly to the _right_ way of doing things by a simple method like this would aid a lot.
b) Documentation - the reasoning you just mentioned in the comment could be added as doc comments for future generations as to why FutureOr should be handled exactly like this so users avoid reinventing the wheel wrongly (like I did).

lrhn:
The other problem, which is why I'm going to say no to this, is that putting an extension method on `FutureOr<T>` means putting it on *every type*. Since `int` is a subtype of `FutureOr<int>`, an extension like `extension <T> on FutureOr<T> { ... }` applies to the raw `int` type.
It does make `then` discoverable on `FutureOr<int>`, but it also makes it discoverable on `int` and `Null` and `Future<int>?` (with the last one creating a `Future<Future<int>?>` as written).

In general, extensions on `FutureOr` should be avoided because they apply to the element type by itself. Generic extensions on `FutureOr<T>` apply to *all* types in the universe.

mit-mit:
Thanks for the contribution, but I'm afraid per the comments from Lasse, we're not able to take this.

PR #46945, opened 2021-08-19
Update comparable.dart

gnudles:
fix typo in documentation


gnudles:
I think it is better to have a large commit of typos, than have commit for each letter fixed. so don't merge it, but collect it

mraleph:
It's not a typo. `iff` means `if and only if` (https://en.wikipedia.org/wiki/If_and_only_if#:~:text=In%20logic%20and%20related%20fields,true%20or%20both%20are%20false.)

gnudles:
Sorry, learned something new today.

PR #46950, opened 2021-08-19
fix NamedType doc typos

grafovdenis:


bwilkerson:
Thanks, but those aren't actually typos. This text was taken from the language specification where 'iff' is used as a shortened form of "if and only if" (https://www.dictionary.com/browse/iff). But I'm guessing you figured that out, given that you closed the PR.

grafovdenis:
> Thanks, but those aren't actually typos. This text was taken from the language specification where 'iff' is used as a shortened form of "if and only if" (https://www.dictionary.com/browse/iff). But I'm guessing you figured that out, given that you closed the PR.

Yep, sorry)

bwilkerson:
No problem. We're always happy to have contributions! Community involvement is valuable at all levels.

Issue #46764, opened 2021-07-30
Field "exports" has not been initialized - Analyzer


Felcks:
### flutter pub run build_runner build

```
LateInitializationError: Field 'exports' has not been initialized.
#0      LibraryReader.exports (package:analyzer/src/summary2/bundle_reader.dart)
#1      LinkedElementFactory.exportsOfLibrary (package:analyzer/src/summary2/linked_element_factory.dart:157:20)
#2      LinkedElementFactory.buildExportNamespace (package:analyzer/src/summary2/linked_element_factory.dart:51:42)
#3      LibraryElementImpl.exportNamespace (package:analyzer/src/dart/element/element.dart:3628:42)
#4      NamespaceBuilder._getExportMapping (package:analyzer/src/dart/resolver/scope.dart:271:20)
#5      NamespaceBuilder.getImportedElements (package:analyzer/src/dart/resolver/scope.dart:211:15)
#6      new PrefixScope (package:analyzer/src/dart/element/scope.dart:181:48)
#7      new _LibraryImportScope (package:analyzer/src/dart/element/scope.dart:295:28)
#8      new LibraryScope (package:analyzer/src/dart/element/scope.dart:106:15)
#9      LibraryElementImpl.scope (package:analyzer/src/dart/element/element.dart:3840:23)
#10     LibraryBuilder.buildScope (package:analyzer/src/summary2/library_builder.dart:134:21)
#11     Linker._computeLibraryScopes (package:analyzer/src/summary2/link.dart:170:15)
#12     Linker._buildOutlines (package:analyzer/src/summary2/link.dart:98:5)
#13     Linker.link (package:analyzer/src/summary2/link.dart:78:5)
#14     link (package:analyzer/src/summary2/link.dart:35:10)
#15     LibraryContext.load2.loadBundle (package:analyzer/src/dart/analysis/library_context.dart:172:24)
#16     SetMixin.forEach (dart:collection/set.dart:142:30)
#17     LibraryContext.load2.loadBundle (package:analyzer/src/dart/analysis/library_context.dart:110:32)
#18     LibraryContext.load2.<anonymous closure> (package:analyzer/src/dart/analysis/library_context.dart:203:17)
#19     PerformanceLog.run (package:analyzer/src/dart/analysis/performance_logger.dart:32:15)
#20     LibraryContext.load2 (package:analyzer/src/dart/analysis/library_context.dart:201:12)
[...]

PR #47230, opened 2021-09-16
fix https://github.com/dart-lang/sdk/issues/46764

quango2304:
https://github.com/dart-lang/sdk/issues/46764

Late Initialization Error: Field 'exports' has not been initialized.#0 LibraryReader.exports


mraleph:
This likely just hides the problem, rather than fixes it. As @scheglov has indicated in his comments there is an expectation that this field is initialized at the point where it is used, which implies that there is some sort of logical error somewhere else.

PR #47359, opened 2021-10-02
Incorrect wording

mnordine:
I think "contains" was meant here.


mit-mit:
Code review in https://dart-review.googlesource.com/c/sdk/+/215340

Thanks @mnordine !

PR #47430, opened 2021-10-11
Documentation update for JsonEncoder and JsonDecoder class

jpsiitonen:
Added example codes for classes: JsonEncoder and JsonDecoder.
Added example codes for functions: jsonEncode and jsonDecode.


jpsiitonen:
@googlebot I signed it!

PR #47537, opened 2021-10-24
Mention Object.hashCode in the note about LinkedHashSet in the Set documentation

hacker1024:
The note about `LinkedHashSet` in the `Set` documentation is a little misleading, and it can lead developers to implement `==` without `hashCode`, only to be confused when their set doesn't work correctly.

See this [StackOverflow question](https://stackoverflow.com/questions/69664073/dart-comparison-of-objects-within-a-list/69694646#69694646), for example.

This PR mentions `Object.hashCode` along with `Object.==` in the `Set` documentation explaining `LinkedHashSet`.

### DIFF
--- a/sdk/lib/core/set.dart
+++ b/sdk/lib/core/set.dart
@@ -13,8 +13,8 @@ part of dart.core;
 /// elements are treated as being the same for any operation on the set.
 ///
 /// The default [Set] implementation, [LinkedHashSet], considers objects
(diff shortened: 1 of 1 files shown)


mit-mit:
@hacker1024 thanks for the contribution! For the Dart SDK we do the review in a separate tool. Can you please review the comments here? https://dart-review.googlesource.com/c/sdk/+/217861

PR #47613, opened 2021-11-03
Update CSS lib version dep

TedSander:
This is needed internally to get a change to global-context rename which was requested by the CSS team at Google.


mit-mit:
Code review in https://dart-review.googlesource.com/c/sdk/+/219121

PR #47693, opened 2021-11-14
Fix `ZoneBinaryCallback` documentation

uxduck:
There was a small mistake in the documentation, because it was copy-pasted from `ZoneUnaryCallback` and not fixed.


kevmoo:
Great find and thanks for the fix, @ben-xD !!! 🙏  👍

PR #47716, opened 2021-11-17
Documentation update for HashMap

jpsiitonen:
Added example codes for HashMap - small updates to description text.


sdk/lib/collection/hash_map.dart
```diff
@@ -17,6 +17,7 @@ typedef _Equality<K> = bool Function(K a, K b);
 typedef _Hasher<K> = int Function(K object);
 
 /// A hash-table based implementation of [Map].
+/// The [HashMap] is unordered (the order of iteration is not guaranteed).
 ///
```
bernaferrari, line 21: You could add. For ordered, use [whateverMap].

PR #47749, opened 2021-11-22
[fuchsia] use the binary tools in the fuchsia SDK

omerlevran46:
This commit moves from using the bash scripts to using the binary tools.


mit-mit:
Review is https://dart-review.googlesource.com/c/sdk/+/221024

PR #47947, opened 2021-12-16
Expand the range of trace line numbers

Yogaflre:
00~999.
Solve the problems that more than 100 lines cannot be translated.

### DIFF
--- a/pkg/native_stack_traces/lib/src/convert.dart
+++ b/pkg/native_stack_traces/lib/src/convert.dart
@@ -70,7 +70,7 @@ const _symbolOffsetREString = r'(?<symbol>' +
     r')\+(?<offset>(?:0x)?[\da-f]+)';
 final _symbolOffsetRE = RegExp(_symbolOffsetREString);
 final _traceLineRE = RegExp(
-    r'    #(\d{2}) abs (?<absolute>[\da-f]+)(?: virt (?<virtual>[\da-f]+))? '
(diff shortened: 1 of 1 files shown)


Yogaflre:
CLA signed.

sstrickl (changes requested):
I'd suggest going ahead and changing it from `{2}` to `+` instead to allow any length number here, because the chance that any non-stacktrace line in the parsed input matches this regexp is minimal as it is.

While I expect stack traces with 4 or more digit line numbers will be... unlikely, might as well allow them while we're making the change so we don't have to extend it again in the future. :)

mraleph:
Gerrit CL: https://dart-review.googlesource.com/c/sdk/+/224560

Yogaflre:
> I'd suggest going ahead and changing it from `{2}` to `+` instead to allow any length number here, because the chance that any non-stacktrace line in the parsed input matches this regexp is minimal as it is.
> 
> While I expect stack traces with 4 or more digit line numbers will be... unlikely, might as well allow them while we're making the change so we don't have to extend it again in the future. :)

I am not familiar enough with this part of the code. Thanks for your suggestions.

sstrickl:
Thanks for the fix! Have approved the Gerrit CL and put in the commit queue, so should land shortly.

sstrickl:
I've published version 0.4.5 of the `native_stack_traces` package, which includes this fiix.

PR #48250, opened 2022-01-28
Update precompiler_trace.dart

polina-c:
`CallGraphNode.dominator` should be nullable, not late, as it is expected for root node of a CallGraph to have a null dominator.

When it is non-nullable, devTools to fails convert to null-safe version of vm_snapshot_analysis: https://github.com/flutter/devtools/runs/4974867863?check_suite_focus=true


polina-c:
Will resubmit with gerrit

PR #48561, opened 2022-03-15
Be more explicit about when Futures are used

Hixie:


Hixie:
Cloned to https://dart-review.googlesource.com/c/sdk/+/237241

bwilkerson:
Just FYI, the PR will automatically be closed when the Gerrit CL is committed.

PR #48572, opened 2022-03-16
Fix build script host arch detection on M1

osa1:
This fixes host architecture guessing on M1.

Since arm64 and aarch64 are synonymous branches for those are merged.


osa1:
Didn't realize there's a patch in Gerrit for this. Closing.

PR #48665, opened 2022-03-25
Fallback to uname for detecting ARM CPU architecture

ntkme:
Currently running official dart `linux/arm/v7` container image on x86_64 with QEMU results in `Unrecognized ARM CPU architecture`.

This is due to QEMU reporting host `/proc/cpuinfo` instead of the emulated architecture.

Checking `uname` in addition to `/proc/cpuinfo` would make `linux/arm/v7` build work consistently regardless of whether `--use-qemu` is supplied during SDK build time.

TEST=running the qemu bot
TEST=ran the ffi/hardfp_test.dart locally both in AOT and JIT mode.

### DIFF
--- a/build/config/linux/BUILD.gn
+++ b/build/config/linux/BUILD.gn
@@ -35,10 +35,6 @@ config("sdk") {
       ]
     }
   }
-
-  if (is_qemu) {
-    cflags += [ "-DDART_RUN_IN_QEMU_ARMv7" ]
-  }
 }
 
 config("executable_config") {
--- a/runtime/vm/cpu_arm.cc
+++ b/runtime/vm/cpu_arm.cc
@@ -20,6 +20,10 @@
 #elif defined(DART_HOST_OS_WINDOWS)
 #include <processthreadsapi.h>
 #endif
+#if !defined(DART_HOST_OS_WINDOWS)
+#include <string.h>      /* NOLINT */
+#include <sys/utsname.h> /* NOLINT */
+#endif
 #endif
 
 // ARM version differences.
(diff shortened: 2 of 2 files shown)


parlough:
Just for reference, this will be reviewed on Gerrit: https://dart-review.googlesource.com/c/sdk/+/238841

dcharkes:
Did you double check that FFI calls with floating points work? And that they work both in AOT and in JIT?

As far as I can remember, we needed this flag to make the precompiler and precompiled runtime agree on whether it is running in hardfp or softfp for ARMv7.

e.g. One might precompile a Dart program on your x64 host for either armv7 hardfp or armv7 softfp, and you need to somehow tell that to the Dart precompiler.

That said, linux/arm/v7 should be both on a physical device and on qemu most often be hardfp I believe.

ntkme:
https://github.com/dart-lang/sdk/blob/f6dd348f555c02d1cbbbaaba1b5cb43b82e9e504/runtime/vm/cpu_arm.cc#L207-L213

This is the only other place `DART_RUN_IN_QEMU_ARMv7` (`--use-qemu`) is used.

`hardfp_supported_` here is statically defined. For the docker-qemu use case, the official linux/arm/v7 image is built with hardfp, and qemu is also hardfp so that it should work.

In my opinion, if a user compiles the SDK with softfp and then run it in qemu hardfp, that is a user problem.

dcharkes:
> In my opinion, if a user compiles the SDK with softfp and then run it in qemu hardfp, that is a user problem.

Well, that is what we're trying to prevent here with this check.

Maybe it would be a better solution to build the `linux/arm/v7` image with the `DART_RUN_IN_QEMU_ARMv7` passed. @athomas do you know where the image is built?

dcharkes:
> official dart `linux/arm/v7` container image

@ntkme  does this refer to the Linux armv7 SDK download [from here](https://dart.dev/get-dart/archive)? or something else?

athomas:
Now that I re-read this I think @ntkme refers to the docker image, e.g.: 
https://hub.docker.com/layers/dart/library/dart/latest/images/sha256-b972de46eb24f117322cfa6ecedfa998a90be7a96b66d0e81ad9f4794f947009?context=explore

athomas:
@dcharkes the image is built from the docker files here (spoiler: it downloads the zip from dart.dev):
https://github.com/dart-lang/dart-docker/blob/main/stable/bullseye/Dockerfile

dcharkes:
@ntkme if we build that image with `DART_RUN_IN_QEMU_ARMv7` defined, it would also solve your problem, correct?

athomas:
Attempt at implementing @dcharkes' suggestion: https://dart-review.googlesource.com/c/sdk/+/240080

ntkme:
@dcharkes Yes, I think that would work.

dcharkes:
I've spent some time tinkering with different solutions. I think we can get away with removing `DART_RUN_IN_QEMU_ARMv7` and simplifying the code base. See https://dart-review.googlesource.com/c/sdk/+/240400.

Please incorporate those changes, or alternatively I can merge that CL.

ntkme:
@dcharkes Added your patch set to this PR.

ntkme:
Updated to resolve conflict with recent windows arm related changes.

ntkme:
@dcharkes Would you mind take another look on the updated commit and see if any further changes will be needed?

Issue #43612, opened 2020-09-29
Support symbolicating native call stacks with --split-debug-info


xster:
Downstream feature request: https://github.com/flutter/flutter/issues/60189

https://github.com/dart-lang/sdk/issues/35851 now supports translating obfuscated Dart stacks into the original stack, but as far as I understand, after applying --split-debug-info, we no longer have enough symbols left in the binary to then generate Xcode Debugging Symbols (dSym) from the binary/framework. This means users can no longer symbolize native call stacks into Dart call stacks. 

Would it be possible for --split-debug-info to output both a obfuscated Dart -> Dart map as well as a native -> Dart map? 

cc @sstrickl and @mraleph

sstrickl:
So the separate debugging information (which is in ELF format) contains both the symbols and the DWARF information to symbolicate non-symbolic Dart stack traces. Currently, one way to do so is to use [pkg/native_stack_traces](https://pub.dev/packages/native_stack_traces).

However, I'm not sure if an easy way to convert ELF files containing static symbol tables and DWARF information to a dSym file (which I'm guessing is the Mach-O equivalent of having a separate file with just debugging information similar to our ELF one) already exists. I'll look into this.

(I'd be happy to just make a Mach-O output instead, but it seems like [documentation of the Mach-O format](https://web.archive.org/web/20090901205800/http://developer.apple.com/mac/library/documentation/DeveloperTools/Conceptual/MachORuntime/Reference/reference.html) was pulled at some point, so I'm not sure how stable it is outside of "what XCode produces" nowadays.)

jan-auer:
@sstrickl are you saying that you're generating ELF containers even for iOS / macOS builds?

File formats like MachO and ELF are usually meant to be platform-dependent, so I'm not sure if it makes sense to even convert an ELF file into a MachO file. Many debuggers and symbolizers, including `llvm-symbolizer` and `addr2line` are able to read both file formats. The DWARF information contained in the debug sections is equal in its format.

Usually, on iOS builds you would output the entire debugging information into the `__debug_*` sections of the binary and then have the build toolchain run `dsymutil` to split it out into the debug companion. This takes care of creating a reference in the binary and moves all relevant sections and load commands over. Note that the convention for section naming on macOS is with two leading underscores instead of a dot.

Another important point is to create a `LC_UUID` load command before invoking dsymutil. It is essentially the same as the `.gnu.build-id` section in ELF and allows debuggers to identify a debug companion.

[...]

PR #48767, opened 2022-04-08
feat: support dSYM symbolication in native_stack_traces

vaind:
This is part of #43612 - enabling Flutter users to `split-debug-info` into `dSYM` and symbolize it back, using `flutter symbolize` or directly using the CLI tool in `native_stack_traces`.

The general approach taken was proposed by @mraleph in this comment: https://github.com/dart-lang/sdk/issues/43612#issuecomment-842164854

There are quite a few open questions here as well as in the Flutter tooling, but I'd like to get the feedback on this. There are two main pain points in this PR
1. the current implementation in `dwarf.dart` only supports ELF format - that doesn't match the dwarf file contained in a dSYM container and it is beyond me to implement the dwarf parsing so the proposed solution uses a CLI tool `atos` get symbol names for individual addresses. This, however, loses the option to load the dwarf from a Uint8List, which is currently done by the `flutter` tool to allow unit testing.
2. So far I couldn't figure out a way to extract the following addresses from dSYM and they're required to properly call `atos`... any ideas? Currently they're parsed from the given stack trace (the second line already was before this PR).
    ```
    isolate_dso_base: 116904000, vm_dso_base: 116904000
    isolate_instructions: 11690f6d0, vm_instructions: 11690aee0
    ```
    
Additionally, I haven't seen any tests for the existing CLI tool. Hopefully, the CI will give me some feedback here.

### DIFF
--- a/pkg/native_stack_traces/bin/decode.dart
+++ b/pkg/native_stack_traces/bin/decode.dart
@@ -162,7 +162,9 @@ Dwarf? _loadFromFile(String? original, Function(String) usageError) {
     return null;
   }
   final filename = path.canonicalize(path.normalize(original));
-  if (!io.File(filename).existsSync()) {
+  if (filename.endsWith('dSYM')
+      ? !io.Directory(filename).existsSync()
+      : !io.File(filename).existsSync()) {
     usageError('debug file "$original" does not exist');
     return null;
   }
@@ -207,7 +209,7 @@ void find(ArgResults options) {
     return usageError('need both VM start and isolate start');
   }
 
-  var vmStart = dwarf.vmStartAddress;
+  late final int vmStart;
   if (options['vm_start'] != null) {
     final address = tryParseIntAddress(options['vm_start']);
     if (address == null) {
@@ -215,9 +217,11 @@ void find(ArgResults options) {
           '${options['vm_start']}');
     }
     vmStart = address;
+  } else {
+    vmStart = dwarf.vmStartAddress;
   }
 
-  var isolateStart = dwarf.isolateStartAddress;
+  late final int isolateStart;
   if (options['isolate_start'] != null) {
     final address = tryParseIntAddress(options['isolate_start']);
     if (address == null) {
@@ -225,6 +229,8 @@ void find(ArgResults options) {
           '${options['isolate_start']}');
     }
     isolateStart = address;
+  } else {
+    isolateStart = dwarf.isolateStartAddress;
   }
 
   final header = StackTraceHeader(isolateStart, vmStart);
--- a/pkg/native_stack_traces/lib/src/convert.dart
+++ b/pkg/native_stack_traces/lib/src/convert.dart
@@ -13,26 +13,49 @@ String _stackTracePiece(CallInfo call, int depth) =>
 
 // A pattern matching the last line of the non-symbolic stack trace header.
 //
-// Currently, this happens to include the only pieces of information from the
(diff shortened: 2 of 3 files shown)


mraleph:
I started to take an initial look at this. 

> 1. the current implementation in dwarf.dart only supports ELF format - that doesn't match the dwarf file contained in a dSYM container and it is beyond me to implement the dwarf parsing so the proposed solution uses a CLI tool atos get symbol names for individual addresses.

We are in luck here. We have actually recently added a [Mach-O parser][1] which might be good enough for our purposes here. It has somewhat different API from Elf/Dwarf but I think we can contort things enough to make them work. 

I am trying to prototype something up. 

> 2. So far I couldn't figure out a way to extract the following addresses from dSYM and they're required to properly call atos... any ideas? Currently they're parsed from the given stack trace (the second line already was before this PR).

You can't extract these values from `dSYM` because they are _runtime values_. (The same Mach-O file can be loaded into different places in memory). 

Though I think this question is moot if we start reading DWARF natively from `dSYM` instead of relying on `atos`.

[1]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2native/lib/macho_parser.dart

vaind:
> We are in luck here. We have actually recently added a [Mach-O parser](https://github.com/dart-lang/sdk/blob/main/pkg/dart2native/lib/macho_parser.dart) which might be good enough for our purposes here. It has somewhat different API from Elf/Dwarf but I think we can contort things enough to make them work.

Cool, I'll have a look.

> You can't extract these values from `dSYM` because they are _runtime values_. (The same Mach-O file can be loaded into different places in memory).

Yeah, that was confusing to me because the original dwarf symbolization somehow reads them, e.g. https://github.com/dart-lang/sdk/blob/84497e41ca4421251215faffe889283e319fdfad/pkg/native_stack_traces/lib/src/dwarf.dart#L1265

fzyzcjy:
Looking forward to the PR!

mraleph:
I hacked some stuff together to verify that we can actually read DWARF from Mach-O. See this CL: https://dart-review.googlesource.com/c/sdk/+/242108

Seems to work like a charm in my limited testing. 

This also uncovered some bugs in our implementation of DWARF reading. 

I am planning to take some time next week to clean this CL up.

mkustermann:
@sstrickl has kindly agreed to take this work over from @mraleph .

vaind:
> @sstrickl has kindly agreed to take this work over from @mraleph .

awesome. Please let me know if there's anything I can do to help land it so those changes.

sstrickl:
Hi @vaind, sorry about the delay, but if you could take a look at https://dart-review.googlesource.com/c/sdk/+/240645 and address the comments there (it's just about the TODOs), I think otherwise we can go with this initial cut as is and then I'll work on addressing any remaining open questions, for example, the TODOs you've written plus those issues noticed by Slava during his work on his CL.

sstrickl:
Ah. Gotcha, I'll take a look at working on your CL specifically instead then. Sorry for the misunderstanding.

vaind:
> Ah. Gotcha, I'll take a look at working on your CL specifically instead then. Sorry for the misunderstanding.

Great, let me know if there's anything I should do @sstrickl

scott-the-brewer:
@sstrickl is there a CL I can follow for this feature? Is it https://dart-review.googlesource.com/c/sdk/+/242108? Just noticed there hasn't been any update on it for weeks.

PR #48798, opened 2022-04-12
Fix documentation for operators `<<` and `>>`.

halildurmus:
I couldn't understand what `shiftIndex` is referring to while reading the docs for operators `<<` and `>>`. 
I think `shiftIndex` should be replaced with `shiftAmount`.


kevmoo:
See https://dart-review.googlesource.com/c/sdk/+/241045

PR #48983, opened 2022-05-09
Fix the server name cannot be set when SNI SSL/TLS request causes handshake failure or request error

chansLam:
BUG:
The following error occurs when using a server without domain name resolution to make an SNI request:
Unhandled exception:
HandshakeException: Connection terminated during handshake

### DIFF
--- a/sdk/lib/_http/http.dart
+++ b/sdk/lib/_http/http.dart
@@ -1688,6 +1688,10 @@ abstract class HttpClient {
   void set badCertificateCallback(
       bool Function(X509Certificate cert, String host, int port)? callback);
 
+  /// Sets the server name used for SSL/TLS secure connections.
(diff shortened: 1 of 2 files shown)


mit-mit:
Review CL is https://dart-review.googlesource.com/c/sdk/+/243980

devoncarew:
@chansLam - there is some review feedback in https://dart-review.googlesource.com/c/sdk/+/243980 (we don't use github reviews directly for the sdk repo, but can copy the PRs into our review system).

brianquinlan:
Hi, I made some comments on https://dart-review.googlesource.com/c/sdk/+/243980 if you are still looking at this.

mit-mit:
No reply, closing as assumed stale

PR #49015, opened 2022-05-14
Fix several typos in sdk/lib/internal/internal.dart

johnslavik:
I've fixed a few typos in the mentioned module while flicking through the dart built-in libraries.


mit-mit:
Review CL is https://dart-review.googlesource.com/c/sdk/+/244940

mit-mit:
Thanks for the contribution; this will land shortly!

PR #49016, opened 2022-05-14
Fix several typos in sdk/lib/internal/internal.dart

johnslavik:
I've fixed a few typos in the [sdk/lib/internal/internal.dart](https://github.com/dart-lang/sdk/blob/e7590a9bb4169ffca98a7c5999684867d2aee628/sdk/lib/internal/internal.dart) module while flicking through the built-in dart libraries.


mraleph:
Thanks for the PR! Could you please resubmit it against the `main` rather than `dev` branch?

`dev` branch is for `dev` channel releases rather than for development. It's a bit of a confusing name.

PR #49078, opened 2022-05-22
Replace `null` with `_TIMEOUT_EVENT`.

nmtigor:


sdk/lib/_internal/vm/lib/timer_impl.dart
```diff
@@ -437,14 +437,14 @@ class _Timer implements Timer {
     if (!_receivePortActive) {
       _createTimerHandler();
     }
-    VMLibraryHooks.eventHandlerSendData(null, _sendPort!, wakeupTime);
```
mraleph, line 440: The first argument is the sender, not the message. `null` means timer.

If you want to make more readable you could make 

```dart
const Object? timerSender = null;
```

and use it here:

```dart
VMLibraryHooks.eventHandlerSendData(timerSender, ...);
```

nmtigor: All right.

PR #49164, opened 2022-06-02
allow rename on extension via fix_data.yaml

a14n:


a14n:
https://dart-review.googlesource.com/c/sdk/+/246983

PR #49179, opened 2022-06-03
Enable Large File Support extension (LFS)

ntkme:
This PR adds flags to enable LFS to make 32-bit builds to correctly handle 64-bit inode from the ext4/xfs file systems.

Specifically, many APIs internally uses `readdir()` for listing directory, which may crash with an `EOVERFLOW`, `Value too large for defined data type`.

See https://bugs.gentoo.org/681790#c0 for some background about this issue.

### DIFF
--- a/build/config/linux/BUILD.gn
+++ b/build/config/linux/BUILD.gn
@@ -35,6 +35,13 @@ config("sdk") {
       ]
     }
   }
+
+  # Enable Large File Support extension (LFS)
+  cflags += [
+    "-D_FILE_OFFSET_BITS=64",
+    "-D_LARGEFILE_SOURCE",
+    "-D_LARGEFILE64_SOURCE",
+  ]
 }
 
 config("executable_config") {


mit-mit:
Review CL is https://dart-review.googlesource.com/c/sdk/+/247245

ntkme:
Two CI tests failed:
- https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8812086779767926721/+/u/ffi_tests/stdout
  - This appears to be a bug in the qemu 2.11, which has been fixed a long time ago: https://github.com/qemu/qemu/commit/73a988d957b9142e0a005f4dc87944574f02de51
- https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8812086779767926737/+/u/vm_unit_tests/stdout
  - It is unclear to me what is happening.

mraleph:
@ntkme you could wrap your code into `if (!is_qemu)` to workaround the issue for now. AFAIK we are blocked from updating to newer qemu 

/cc @athomas is there a timeline for when we will be able to use newer qemu?

mraleph:
@ntkme I'd assume that your build is not setting `is_qemu` or does it? You are probably building SDK with default flags. `is_qemu` is probably only set when building for our `qemu` bots.

athomas:
@mraleph The qemu flake already exists on [main]. So it shouldn't block this PR. About getting a newer qemu: we're not planning to replace the bionic images we currently use until next year (when bionic LTS exits Ubuntu standard support). We can move the qemu builder to newer images earlier but I don't have an ETA, yet.

[main]: https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8813904858121678753/+/u/ffi_tests/stdout

brianquinlan:
I made a comment about tests at https://dart-review.googlesource.com/c/sdk/+/247245 but I'm happy to commit this if writing them isn't practical.

ntkme:
@brianquinlan I'm not familiar with the test structure in this project and I will leave it to the dart team for adding tests for this.

Issue #49372, opened 2022-06-30
Compile deferred component error. Seem like a compile error in gen_snapshot.


lawangle:
## Environment:
flutter tools:  2.10.4(https://github.com/flutter/flutter/commit/c860cba910319332564e1e9d470a17074c1f2dfd)
dart version: 2.16.2 (https://github.com/dart-lang/sdk/commit/6fbb9236c65a54c26830f047b88709ac11bdf2ae)

## Replay steps: 
1. git clone flutter gallery: https://github.com/flutter/gallery/tree/fully-deferred-gallery . (I tried other branch, the bug is the same.) 
2. run: flutter build appbundle --verbose.
3. gen_snapthot .... command start
4. waiting and gen_snapthot fail

The problem log in my local environment is:
`[+822599 ms] [+822565 ms] Unresolved call to Code([Optimized] ___TabsNonScrollableDemoState&State&SingleTickerProviderStateMixin&RestorationMixin._updateProperty) from Code([Optimized]
__HomePageState&State&SingleTickerProviderStateMixin&RestorationMixin.registerForRestoration.listener)
[        ] [   +2 ms] Unresolved call to Code([Optimized] __CupertinoAlertDemoState&State&RestorationMixin._updateProperty) from Code([Optimized]
__LoginPageState&State&RestorationMixin.registerForRestoration.listener)
[        ] [        ] Unresolved call to Code([Optimized] __CupertinoAlertDemoState&State&RestorationMixin._updateProperty) from Code([Optimized]
__LoginPageState&State&RestorationMixin.registerForRestoration)
[        ] [        ] Unresolved call to Code([Optimized] __CupertinoAlertDemoState&State&RestorationMixin._updateProperty) from Code([Optimized] [tear-off]
__LoginPageState&State&RestorationMixin._updateProperty)
[        ] [        ] Unresolved call to Code([Optimized] __ProgressIndicatorDemoState&State&SingleTickerProviderStateMixin._updateTicker) from Code([Optimized]
__RallyPieChartState&State&SingleTickerProviderStateMixin.createTicker)
[        ] [        ] Unresolved call to Code([Optimized] __ProgressIndicatorDemoState&State&SingleTickerProviderStateMixin._updateTicker) from Code([Optimized] [tear-off]
__RallyPieChartState&State&SingleTickerProviderStateMixin._updateTicker)
[...]

PR #49393, opened 2022-07-04
[vm] Bug fix in dedup optimization when build deferred components.

lawangle:
`Dedup optimization` will remove duplicated Code object and keep one instance.
When building deferred components, the only Code object (e.x. as callee B) maybe
exist in different loading unit from the caller A. This will cause a rellocation
bug if A->B was a relative call.

Add a flag to indicate building with deferred components.
When working with dedup and deferred components, assignUnit first and
Code object from different unit is unequal.

Issues: https://github.com/dart-lang/sdk/issues/49372

Signed-off-by: fangzheyuan <fangzheyuan@bytedance.com>

### DIFF
--- a/runtime/tests/vm/dart/regress_49372_deferred.dart
+++ b/runtime/tests/vm/dart/regress_49372_deferred.dart
@@ -1,0 +1,12 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Regression test for https://github.com/dart-lang/sdk/issues/49372.
+// Verifies that dedup optimization doesn't merge Code from different unit.
+
+@pragma('vm:never-inline')
+int foo() => 10;
+
+@pragma('vm:never-inline')
+int bar() => foo() + 2;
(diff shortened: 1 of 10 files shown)


lawangle:
This Pr fix one bug in building deferred component. I first found this error in https://github.com/dart-lang/sdk/issues/49372.
At first, i thought this bug only existed in older versions. But when i found out the real cause of the issue, i tnink this bug still exist in current master branch.

I will explain this bug.
`___TabsNonScrollableDemoState&State&SingleTickerProviderStateMixin&RestorationMixin._updateProperty) from Code([Optimized] __HomePageState&State&SingleTickerProviderStateMixin&RestorationMixin.registerForRestoration.listener)`
This error log shows than xxx.registerForRestoration.listener(caller) called xxx._updateProperty(callee) but callee could be
found in relocation phase.

__HomePageState is in unit4 and ___TabsNonScrollableDemoState is in another unit and this call entry's kind is relative call, that means caller and callee shouid in the same unit. But they are in the different units, so crash.

The cause is the dedup optimization. Actually 2.16.2's dedup. In 2.16.2, the dedup try to merge Code Object instead of instruciton. 
Before dedup, this call entry is from __HomePageState.xxx.registerForRestoration.listener to __HomePageState.xxx._updateProperty, relative call is ok. 
After dedup, __HomePageState.xxx._updateProperty is replaced with ___TabsNonScrollableDemoState.xxx and ___TabsNonScrollableDemoState is not exist in unit4.

mraleph:
Thanks for the contribution! Great catch. 

Initial code review comments are here: https://dart-review.googlesource.com/c/sdk/+/250491

lawangle:
Change to  https://dart-review.googlesource.com/c/sdk/+/250720/

PR #49496, opened 2022-07-21
Fix documentation for `Object.hashAllUnordered`

kaboc:
The method name in the document seems wrong.
I also found that the last bit is only slightly different from that of `hashAll`, but don't see any reason it should be.


mit-mit:
Associated review will be https://dart-review.googlesource.com/c/sdk/+/252400

PR #49549, opened 2022-07-28
Build fixes for Windows on Arm

p-b-o:
As part of Linaro [Windows on Arm project](https://www.linaro.org/windows-on-arm/), we are helping to enable some projects for this new platform.
Dart/Flutter is on our list, and we are willing to make it work for this platform. Luckily, a lot of work has already been done by Dart community (thanks!), and only a few bits are missing.

This PR fixes build of dart-sdk for windows on arm (natively) when using a "vanilla" visual studio installation.
After contacting @rmacnak-google (who ported for this platform), it seems build was not tried out of google infrastructure, which seems to have some custom setup.

```python3 ./tools/build.py --arch arm64 --mode release```, now results in a working build.

Please see individual commits for details.


pbo-linaro:
To be upstream directly with chromium build files.

PR #49597, opened 2022-08-04
Remove "grain" from AudioBufferSourceNode.start()

mnordine:
I have no idea why "grain" was in the parameters. There is no reference to "grain" in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode/start), for example.


mraleph:
These files are autogenerated from IDLs and IDLs has this name: https://chromium.googlesource.com/chromium/blink/+/refs/heads/main/Source/modules/webaudio/AudioBufferSourceNode.idl#39

mnordine:
I guess in that case, it's better to just leave it alone.

PR #49877, opened 2022-08-31
Change the contract of the method getAllocationProfile.

polina-c:


polina-c:
As discussed offline vm team prefer not to change contract to have flexibility for other implementations of garbage collector.

PR #50257, opened 2022-10-20
Deleted repetition "/// (We return the string)" from the Documentation of DartObject.toSymbolValue

Alexander-Aue-Johr:
…n of DartObject.toSymbolValue

Deleted "/// (We return the string)" from the documentation of `DartObject.toSymbolValue` because it is a repetition of the first sentence and it seems to be a residue of a internal temporary description. If the "/// (We return the string" was meant to stay here, then the closing ")" is missing.

### DIFF
--- a/pkg/analyzer/lib/dart/constant/value.dart
+++ b/pkg/analyzer/lib/dart/constant/value.dart
@@ -136,7 +136,6 @@ abstract class DartObject {
   /// or `null` if
   /// * this object is not of type 'Symbol', or
   /// * the value of the object being represented is `null`.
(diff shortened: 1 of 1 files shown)


asashour:
Can we have a review, please?

bwilkerson:
Sorry for the delay, but thanks for bringing this to our attention. Unfortunately there's absolutely no signal when a PR is converted to a CL, so there's no automatic way for us to know when there's something waiting for a review. I have approved the CL and it is currently running on the bots. Assuming that the bots test it will be landed.

PR #50395, opened 2022-11-06
Typo in \param doc name

jhsware:
This typo causes lots of warnings during compilation so it is worth fixing.


mkustermann:
@jhsware Thank you for the PR. The corresponding gerrit CL is [cl/269601](https://dart-review.git.corp.google.com/c/sdk/+/269601)

PR #50427, opened 2022-11-10
docs(Duration): fix typo in `compareTo` method

Zekfad:
Fixes small typo in  `compareTo` method of `Duration`.


mraleph:
Thank you for the contribution. 

It's not a typo. `iff` means `if and only if`. See https://en.wikipedia.org/wiki/Iff

Zekfad:
Thanks for clarification, misunderstood the docs then.

PR #50488, opened 2022-11-16
Removed `Finalizer`s from the unsupported features by dart:ffi

Haidar0096:
As dart:ffi now supports `Finalizer`s, they should be removed from the unsupported features list.


dcharkes:
Thanks for pointing this out @Haidar0096 !

I see this whole section is out of date. I will make a follow up PR to clean up the whole section.

Haidar0096:
@dcharkes 
Since you'll do a full cleanup already, please consider adding examples about efficient returning of arrays from C to dart if you can, or if you want I may be able to write something about it in a PR.

For example, because the `Finalizer`s are now supported, we can do zero-copying an array from C to dart while still cleaning up the backing-up array automatically using `asTypedList` and using `NativeFinalizers`.

Thank you :)

PR #50620, opened 2022-12-03
Leave `dependencies:` node uncommented

timsneath:
Ideally we'd leave the `dependencies:` node itself uncommented, even if the content beneath it is commented out. If the node is commented, then VSCode's "Dart: Add Dependency" command gets confused, and adds a new node at the bottom, rather than uncommenting the `dependencies:` node. 

This change should be benign, and removes the need for Dart-Code to introduce logic to uncomment the parent node. For finesse, this also adds an explanatory comment to indicate that `path:` is an example.

### DIFF
--- a/pkg/dartdev/lib/src/templates/console.dart
+++ b/pkg/dartdev/lib/src/templates/console.dart
@@ -44,7 +44,8 @@ version: 1.0.0
 environment:
   ${common.sdkConstraint}
 
-# dependencies:
+dependencies:
+# Add regular dependencies here, for example:
 #   path: ^1.8.0
 
 dev_dependencies:


pkg/dartdev/lib/src/templates/console.dart
```diff
@@ -44,7 +44,8 @@ version: 1.0.0
 environment:
   ${common.sdkConstraint}
 
-# dependencies:
+dependencies:
```
devoncarew, line 47: 👍  - this is a good change; this will probably also work better w/ the `dart pub add` command.

Two nits; the comment between `dependencies` and the example entry looks odd - it may look better before the dependencies section. And, we have a few more of these commented out dependencies sections in the other templates.

devoncarew: Perhaps https://dart-review.git.corp.google.com/c/sdk/+/273543 instead?

DanTup:
Yeah, this makes sense to me too. I recall a user once sent me a pubspec that had added dependencies but not uncommented `dependencies` but hadn't noticed. I'm not sure why changing this didn't occur to me at the time.

https://dart-review.googlesource.com/c/sdk/+/273543 includes the other project types too so may be better.

pkg/dartdev/lib/src/templates/console.dart
whesse: That Gerrit CL has landed, and I think it replaces this PR, so I am tentatively closing this PR.  If it isn't reopened, then I will remove the branch for this PR as well.

whesse:
Included in  https://dart-review.git.corp.google.com/c/sdk/+/273543 , which has landed.  Closing the PR.

PR #50752, opened 2022-12-16
[dart2js] Correct comment

jsoref:
I have a much larger change that includes changes to this file, but I figure I'd start with this change.

### DIFF
--- a/pkg/compiler/test/model/token_naming_test.dart
+++ b/pkg/compiler/test/model/token_naming_test.dart
@@ -77,7 +77,7 @@ void main() {
   {
     TokenScope scope = new TokenScope(initialChar: $A + 1);
 
-    // We start with 'A'.
+    // We start with 'B'.
     Expect.equals("B", scope.getNextName());
   }
 }


pkg/compiler/test/model/token_naming_test.dart
```diff
@@ -77,7 +77,7 @@ void main() {
   {
     TokenScope scope = new TokenScope(initialChar: $A + 1);
 
-    // We start with 'A'.
+    // We start with 'B'.
```
mraleph, line 80: Thank you for the contribution, but I think this comment is correct. We start with A so we naturally get B when we ask for the next name.

jsoref: I don't think so, line 78 starts with $A+1 which is $B

jsoref: To be slightly more explicit, if your explanation was correct, then the comment in the previous block would be wrong, but it really can't be:
https://github.com/dart-lang/sdk/blob/94f20d77314a91489e2c32fd144b81f506822561/pkg/compiler/test/model/token_naming_test.dart#L69-L72

mraleph: Fair point. It seems that `getNextName` actually returns the current name and advances. Makes sense.

mraleph:
+2ed the CL (https://dart-review.googlesource.com/c/sdk/+/275985), waiting for it to land.

Issue #50754, opened 2022-12-16
Spelling


jsoref:
The [contribution guidelines](https://github.com/dart-lang/sdk/blob/94f20d77314a91489e2c32fd144b81f506822561/CONTRIBUTING.md) suggest using the [Dart Issue Tracker](https://dartbug.com/) which appears to just link to [dart-lang/sdk/issues](https://github.com/dart-lang/sdk/issues), so, I'm opening this ticket...

Here's the rough list of items that I've found:
https://github.com/jsoref/dart-lang-sdk/actions/runs/3711536404

I have a (medium size by my standards) set of changes that I'd like to offer:

Notes:
1. I generally annotate my PRs, but the guidelines suggested I file a ticket before I make a PR, so, I'm not going to do that until later.
2. I generally prefer not to squash my PRs until their content is deemed acceptable because it makes rebasing fairly risky/painful
3. IME, gerrit hates my style of commits.
4. I'm happy to split my PRs by directory or file type.
5. It's generally quite impractical for me to split by change type (comment, api, locals) -- it's quite likely that generally doing this will result in build failures / problematic merge conflicts and incomplete changes
6. I haven't tried to build things, it's quite possible that some of the changes relate to external modules.
7. I do try to identify external files/modules and avoid touching them, but I don't always succeed. There were a couple of files that appeared to be lists of tests that are probably from external modules, I left some of those changes in but, dropping them is trivial (as is dropping changes to any other file/directory).
8. Some projects don't like changing changelogs. I've _tried_ to avoid breaking changelog entries, so if a changelog entry said it was correcting a typo, I tried to leave the type alone, but if a changelog entry included its own errant typo, I tried to correct that
9. A number of names of contributors, authors, and publications had typos.
10. Some items in license blocks have typos, historically lawyers would get antsy about such changes.
11. I do try to correct typos in file names/directories, but it's certainly possible that I haven't done that completely.
12. There are a couple of categories of preference-ish changes, esp. `in to`/`into` or `opt in`/`opt-in`, `work around`/`workaround`, `re-entrant`/`reentrant`. It's trivial for me to drop any individual changeset, I can explain each of my choices (`work around` is a verb-phrase / `workaround` is a noun, `in to` is a preposition and the beginning of a verb, / `into` is a preposition).
13. There are a number of cases where words were doubled, often the correction is dropping the duplicate, but in a number of cases I changed one of the two (this is very manual).

[...]

PR #50764, opened 2022-12-19
Spelling tests

jsoref:
This is the remainder of tests noted in #50754, it's _around_ 50 files and < 100 changes, which is hopefully digestible.

I'm not going to send any other PRs at this time as I'd like some input on how to slice the rest of the bits.

### DIFF
--- a/tests/lib/mirrors/mirrors_reader.dart
+++ b/tests/lib/mirrors/mirrors_reader.dart
@@ -1,6 +1,6 @@
 // Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICESNE file.
+// BSD-style license that can be found in the LICENSE file.
 
 library mirrors.reader;
 
--- a/tests/standalone/io/link_test.dart
+++ b/tests/standalone/io/link_test.dart
@@ -239,7 +239,7 @@ testRenameSync() {
 
 void testLinkErrorSync() {
   Expect.throws(
-      () => new Link('some-dir-that-doent exist/some link file/bla/fisk')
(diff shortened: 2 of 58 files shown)


tests/lib/mirrors/mirrors_reader.dart
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICESNE file.
+// BSD-style license that can be found in the LICENSE file.
```
jsoref, line 3: License changes historically spook lawyers, but this shouldn't be a problem...

tests/standalone/io/link_test.dart
```diff
@@ -239,7 +239,7 @@ testRenameSync() {
 
 void testLinkErrorSync() {
   Expect.throws(
-      () => new Link('some-dir-that-doent exist/some link file/bla/fisk')
+      () => new Link('some-dir-that-does-not exist/some link file/bla/fisk')
```
jsoref, line 242: I personally prefer `nonexistent`, but this seems like the simplest change matching the spirit of the test,

eernstg: Should be fine.

tests/lib/mirrors/mirrors_reader.dart
eernstg: Indeed.

eernstg:
Got this one: https://dart-review.googlesource.com/c/sdk/+/276360. Currently checking that the changes do not cause any new test failures. Several identifiers were modified, but they look like every modified name has been changed consistently, and the name is declared in the same library (so we aren't breaking code because it refers to a misspelled imported name).

PR #50806, opened 2022-12-21
spelling: favorites

jsoref:
This will require a release note and other bits, split off from https://dart-review.googlesource.com/c/sdk/+/276560/1..3/tools/dom/src/KeyName.dart#b44


mit-mit:
CL: https://dart-review.googlesource.com/c/sdk/+/276773

PR #50808, opened 2022-12-21
spelling: is_signed

jsoref:
This change is scoped per https://github.com/dart-lang/sdk/issues/50754

I'm happy to drop this change, but I'm hoping that people will consider that there's value in spelling things out.

I think that `signd` means `signed`, although it's possible that it's `sign`+`d` for something else -- e.g. significant+digits. Given that it's a boolean, this feels like the most likely answer.

The only hunk of this change that gave me confidence in my guess about the meaning of the token was this:
```c++
      const bool signd = instr->Bit(23) == 1;
      // Write the W register for signed values when size < 2.
      // Write the W register for unsigned values when size == 2.
```

I believe the reason that it was written this was is that `signed` is a reserved word. This code uses `_` in various places; including `classes.dart`, `types.dart`, and `dartutils.cc` which specifically use `is_` as a prefix.

### DIFF
--- a/runtime/vm/simulator_arm.cc
+++ b/runtime/vm/simulator_arm.cc
@@ -3040,7 +3040,7 @@ void Simulator::DecodeSIMDDataProcessing(Instr* instr) {
                (instr->Bit(23) == 0) && (instr->Bits(25, 3) == 1)) {
       // Format(instr, "vshlqu'sz 'qd, 'qm, 'qn");
       // Format(instr, "vshlqi'sz 'qd, 'qm, 'qn");
-      const bool signd = instr->Bit(24) == 0;
+      const bool is_signed = instr->Bit(24) == 0;
       const int size = instr->Bits(20, 2);
       if (size == 0) {
         for (int i = 0; i < 16; i++) {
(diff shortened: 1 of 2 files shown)


runtime/vm/simulator_arm.cc
```diff
@@ -3040,15 +3040,15 @@ void Simulator::DecodeSIMDDataProcessing(Instr* instr) {
                (instr->Bit(23) == 0) && (instr->Bits(25, 3) == 1)) {
       // Format(instr, "vshlqu'sz 'qd, 'qm, 'qn");
       // Format(instr, "vshlqi'sz 'qd, 'qm, 'qn");
-      const bool signd = instr->Bit(24) == 0;
+      const bool is_signed = instr->Bit(24) == 0;
```
asashour, line 3043: According to [Effective dart](https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-a-boolean-variable-or-property-comment-with-whether-followed-by-a-noun-or-gerund-phrase), it should be `isSigned`.

However, the owner my prefer `signed`.

mit-mit:
CL: https://dart-review.googlesource.com/c/sdk/+/276774

PR #50866, opened 2023-01-01
Update to add `const DateTime.epoch`

KKimj:
[WIP]

https://github.com/dart-lang/sdk/issues/17014

motivated by
- https://github.com/dotnet/corefx/pull/25235
- https://learn.microsoft.com/en-us/dotnet/api/system.datetime.unixepoch?view=net-7.0

### DIFF
--- a/sdk/lib/core/date_time.dart
+++ b/sdk/lib/core/date_time.dart
@@ -361,15 +361,11 @@ class DateTime implements Comparable<DateTime> {
     }
   }
 
-  /// Constructs a new [DateTime] instance
-  /// with Unix time epoch.
+  /// Epoch of the `DateTime` calendar.
   ///
-  /// The constructed [DateTime] represents 1970-01-01T00:00:00Z
-  /// ```dart
-  /// const unixEpochDay = DateTime.unixEpoch();
-  /// print(unixEpochDay); // 1970-01-01 00:00:00.000Z
-  /// ```
-  external const DateTime.unixEpoch();
+  /// The epoch is the start of January 1st 1970, as UTC time,
+  /// where, e.g., [millisecondsSinceEpoch] is zero.
+  external const DateTime.epoch();
 
   static const int _maxMillisecondsSinceEpoch = 8640000000000000;
 
--- a/sdk/lib/core/date_time.dart
+++ b/sdk/lib/core/date_time.dart
@@ -365,7 +365,11 @@ class DateTime implements Comparable<DateTime> {
   ///
   /// The epoch is the start of January 1st 1970, as UTC time,
   /// where, e.g., [millisecondsSinceEpoch] is zero.
-  external const DateTime.epoch();
+  static const DateTime epoch = DateTime._epoch();
+
+  /// Constructs a new [DateTime] instance
+  /// where, e.g., [millisecondsSinceEpoch] is zero.
+  external const DateTime._epoch();
 
   static const int _maxMillisecondsSinceEpoch = 8640000000000000;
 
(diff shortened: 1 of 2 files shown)


sdk/lib/core/date_time.dart
```diff
@@ -361,6 +361,16 @@ class DateTime implements Comparable<DateTime> {
...
 
+  /// Constructs a new [DateTime] instance
+  /// with Unix time epoch.
+  ///
+  /// The constructed [DateTime] represents 1970-01-01T00:00:00Z
+  /// ```dart
+  /// const unixEpochDay = DateTime.unixEpoch();
+  /// print(unixEpochDay); // 1970-01-01 00:00:00.000Z
+  /// ```
+  external const DateTime.unixEpoch();
```
lrhn, line 372: This should probably just be a constant (in UTC time).

mit-mit:
CL: https://dart-review.googlesource.com/c/sdk/+/277920

sdk/lib/core/date_time.dart
KKimj: Thanks, Updated to be a constant!

sdk/lib/core/date_time.dart
```diff
@@ -361,6 +361,16 @@ class DateTime implements Comparable<DateTime> {
...
 
+  /// Epoch of the `DateTime` calendar.
+  ///
+  /// The epoch is the start of January 1st 1970, as UTC time,
+  /// where, e.g., [millisecondsSinceEpoch] is zero.
+  static const DateTime epoch = DateTime._epoch();
+
+  /// Constructs a new [DateTime] instance
+  /// where, e.g., [millisecondsSinceEpoch] is zero.
+  external const DateTime._epoch();
```
KKimj, line 372: To make _epoch() non-external, a different const constructor must exist in different DateTime.

For example, make DateTime.fromMillisecondsSinceEpoch to const constructor..

So the options I can offer are as follows.
1. Add external const DateTime._epoch();
2. Put it off until another const constructor is added.

KKimj: In my view, the perspective to consider is as follows.
1. Developers using `DateTime.epoch`
2. Developer to maintain `DateTime._epoch()`

For No. 1, there is no risk as to whether `_epoch()` is an extenal.

For No. 2, when `DateTime.fromMillisecondsSinceEpoch` is refactored to a const constructor later, `_epoch` should be refactored as non-extenal.

I think it is worth considering `external _epoch`,
if the risk cost for number 2 is not high,
and it is a big thing to refactor from `DateTime.fromMillisecondsSinceEpoch` to const constructor right away.

lrhn: Assuming another `const` constructor is reasonable. All current implementations use an internal `final int` field to store milliseconds or microseconds since epoch. We can definitely add a `const DateTime._epoch() : _time = 0;` constructor if needed.

My biggest worry is that "epoch" is not a particularly useful time for anything, so I'm not sure what the use-case is.
There is absolutely nothing special about 1970-01-01 00:00:00.000, so why is *that* constant more interesting than, say, 2000-01-01 00:00:00.000?

The fact that it's the zero value for our representation is an implementation detail.

KKimj: Okay sir,
I fully agree with you.

I don't think it's reasonable for `DateTime._epoch()` to become a special one.

I think it is right to wait for `DateTime.fromMillisecondsSinceEpoch` to become a const constructor.

Close this pull-request.
Thanks.

PR #50878, opened 2023-01-02
Update dart_native_api.h

Haidar0096:
Make `as_string` field of `_Dart_CObject` a `const char*` instead of `char*` to avoid forcing users to maintain a separate copy of the string passed to this field if they don't want to get a warning from the compiler


mraleph:
@Haidar0096 if you are still interested in landing this you can cherry pick build fixes from my CL: https://dart-review.googlesource.com/c/sdk/+/277985

Haidar0096:
reopened in #50884

PR #50884, opened 2023-01-03
[vm/api] Make CObject as_string `const char*` instead of `const`

Haidar0096:
Change-Id: I0c8342174391e7f075f0317f6323da9e07f06165

### DIFF
--- a/runtime/bin/dartutils.cc
+++ b/runtime/bin/dartutils.cc
@@ -884,7 +884,7 @@ Dart_CObject* CObject::NewString(intptr_t length) {
 Dart_CObject* CObject::NewString(const char* str) {
   intptr_t length = strlen(str);
   Dart_CObject* cobject = NewString(length);
-  memmove(cobject->value.as_string, str, length + 1);
+  memmove(const_cast<char*>(cobject->value.as_string), str, length + 1);
   return cobject;
 }
 
--- a/runtime/bin/dartutils.cc
+++ b/runtime/bin/dartutils.cc
@@ -875,16 +875,12 @@ Dart_CObject* CObject::NewDouble(double value) {
   return cobject;
 }
 
-Dart_CObject* CObject::NewString(intptr_t length) {
-  Dart_CObject* cobject = New(Dart_CObject_kString, length + 1);
-  cobject->value.as_string = reinterpret_cast<char*>(cobject + 1);
-  return cobject;
-}
-
 Dart_CObject* CObject::NewString(const char* str) {
   intptr_t length = strlen(str);
-  Dart_CObject* cobject = NewString(length);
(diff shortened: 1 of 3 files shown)


mraleph:
Could you address the comment from @mkustermann given here: https://dart-review.googlesource.com/c/sdk/+/278140 ?

Haidar0096:
@mkustermann 
Just to make sure I follow correctly, do you mean I should delete `Dart_CObject* NewString(intptr_t length)` function and change `Dart_CObject *NewString(const char *str)` into 
```
Dart_CObject *NewString(const char *str) {
        return New(Dart_CObject_kString, strlen(str) + 1);
    }
```
?

mraleph:
@Haidar0096 what @mkustermann means is that you should inline the function into its callsite. The code that you have suggested does not do what the original function did (you are missing `memmove`, etc).

```cpp
// Take the body of this function
Dart_CObject* CObject::NewString(intptr_t length) {
  Dart_CObject* cobject = New(Dart_CObject_kString, length + 1);
  cobject->value.as_string = reinterpret_cast<char*>(cobject + 1);
  return cobject;
}

Dart_CObject* CObject::NewString(const char* str) {
  intptr_t length = strlen(str);
  Dart_CObject* cobject = NewString(length);  // <- and replace this callsite with the body.
  memmove(const_cast<char*>(cobject->value.as_string), str, length + 1);
  return cobject;
}
```

So you get this: 

```cpp
Dart_CObject* CObject::NewString(const char* str) {
  intptr_t length = strlen(str);
  Dart_CObject* cobject = New(Dart_CObject_kString, length + 1);
  cobject->value.as_string = reinterpret_cast<char*>(cobject + 1);
  memmove(const_cast<char*>(cobject->value.as_string), str, length + 1);
  return cobject;
}
```

Now you can make it a bit prettier by doing something like:

```cpp
Dart_CObject* CObject::NewString(const char* str) {
  intptr_t length = strlen(str);
  Dart_CObject* cobject = New(Dart_CObject_kString, length + 1);
  char* payload = reinterpret_cast<char*>(cobject + 1);
  memmove(payload, str, length + 1);
  cobject->value.as_string = payload;
  return cobject;
}
```

mkustermann:
Apologies for late reply & thanks for explaining it, @mraleph !

(As background: Using `const_cast<char*>(pointerToConst)` is generally dangerous because it assumes that the `pointerToConst` is actually pointing to mutable memory and not read-only memory. It happens to be here because the memory is `malloc()`ed in another function. But it's good to avoid `const_cast`s if we can - which we can here)

mkustermann:
@Haidar0096 One last comment: Could you also remove the declaration of `NewString(intptr_t)` in the `.h` file?

PR #50890, opened 2023-01-04
Use the function arguments instead of static value

xiaoxiaowesley:
The function `PostAndWait` inputs two arguments which first `port` but never used. It use the argument`port`   instead of the static value `send_port_` .

```
  void PostAndWait(Dart_Port port, Dart_CObject* object) {
    std::unique_lock<std::mutex> lock(mutex);
    const bool success = Dart_PostCObject_DL(send_port_, object);
    if (!success) FATAL("Failed to send message, invalid port or isolate died");

    printf("C   :  Waiting for result.\n");
    while (!notified) {
      cv.wait(lock);
    }
  }
```


dcharkes:
Good spotting!

It is the same port in how the test is set up the only call to the function is referring to the same global variable.

dcharkes:
Merging through https://dart-review.googlesource.com/c/sdk/+/278099.

Issue #40906, opened 2020-03-06
Build SDK with musl


JasMich:
This tracker is for issues related to:

* Common Front End (CFE) and kernel
* Dart VM
* dev_compiler

* Whether you are using: Void Linux x64 musl

It it somehow possible to build Dart using `musl` compiler instead of `glibc`?

mraleph:
Right now we don't provide a way to build Dart with `musl` compiler.

PureTryOut:
I've looked into this a bit and the initial blocker is the fact that the build process requires a pre-build dart binary dynamically linked to glibc. Thus you'd need to bootstrap Dart without using pre-built binaries, and so far I've not found instructions or any efforts to do so.

onodera-punpun:
Any update on this?

PureTryOut:
Well, Alpine Linux has been shipping Dart for quite a bit.
https://pkgs.alpinelinux.org/package/edge/testing/x86_64/dart

Bootstrapped from the Windows version of Dart through Wine...

a4z:
The wine/dart does not help really if you run docker on Mac M1 
(yes I know that I can change the arch of a container)
also, adding wine might remove the size advantage, and it's a super ugly workaround

It would be really nice if dart-SDK would build on alpine. There are quite some other projects that require it and have now real inconvenient situations for users.
One example is static blog site generators that used the old ruby sass and now switch to dart sass. And ups, the container you had does not work anymore. Sory user. That is kind of really bad

PureTryOut:
I agree, it's really, really bad. But it's a way to get _something_ running. Musl itself isn't really a problem, Dart only needs a few small patches to get it working which should be easy enough to upstream.

The issue is the bootstrapping. Dart requires Dart to compile, but when Dart doesn't run on the platform... well you need to have a workaround. In Alpine's case this is Wine right now, but I rather do a full from-source bootstrap, starting from the very last version that doesn't require itself to compile (which is as far as I can see Dart 2.0.0-dev.8.0) and then going up version by version till you're on the most recent one. It's a better chain of trust but also makes it easier to compile it for new architectures, as Wine on anything non-x86 is not that great.
[...]

PR #50998, opened 2023-01-12
Add argument --no-use-tcmalloc

ntkme:
This PR allows the Linux SDK to be built without tcmalloc. It would partially address https://github.com/dart-lang/sdk/issues/40906

By building Linux SDK without tcmalloc, the SDK would run on musl-libc with gcompat. It’s not perfect but still much better than no support at all.


ntkme:
Just learned that this can be done with `export DART_GN_ARGS="dart_use_tcmalloc=false"`, so it is not necessary to add an other option.

PR #51057, opened 2023-01-18
[infra] Create a GIT_TIMESTAMP file for tarball

ntkme:
When building from debian tarball generated from `tools/linux_dist_support/create_tarball.py`, we get `Unknown timestamp` when running `dart --version`.

This PR addresses that by adding a `GIT_TIMESTAMP` file similar to how `GIT_REVISION` file works today.

This also removes a `.decode('utf-8')` call on string which was for Python 2 and no longer works in Python 3.


ntkme:
@athomas @whesse Would you mind take a look at this as you worked on this area recently?

mit-mit:
Review expected in https://dart-review.googlesource.com/c/sdk/+/279278

PR #51171, opened 2023-01-30
Remove unneeded null check after promotion

jarrodcolburn:
Remove unneeded null check after promotion

### DIFF
--- a/sdk/lib/collection/list.dart
+++ b/sdk/lib/collection/list.dart
@@ -363,7 +363,6 @@ abstract class ListMixin<E> implements List<E> {
 
   void shuffle([Random? random]) {
     random ??= Random();
-    if (random == null) throw "!"; // TODO(38493): The `??=` should promote.
 
     int length = this.length;
     while (length > 1) {
@@ -384,7 +383,6 @@ abstract class ListMixin<E> implements List<E> {
   List<E> sublist(int start, [int? end]) {
     int listLength = this.length;
     end ??= listLength;
-    if (end == null) throw "!"; // TODO(38493): The `??=` should promote.
 
     RangeError.checkValidRange(start, end, listLength);
     return List.from(getRange(start, end));


jarrodcolburn (commented):
^

jarrodcolburn:
Question about list Mixin...
The `ListMxin` documentation states that all Mixin methods really only on  `length`,`operator[]` ,`add`, `length=` and `operator[]=`.
https://github.com/dart-lang/sdk/blob/df6ea6326fb51e1f1deb8b5e25fd53afa9f1d605/sdk/lib/collection/list.dart#L48-L54
 My ((in)correct?) understand of this is... no public method defined on the mixin calls another (public method). This ensures later classes that  `@override` a mixin's method will be limited to just that method, not effecting any of the other functions defined on the mixin. 

However in the following snippet, notice `insert` (public method) calls `setRange` (public method). As such `@override`ing of `setRange` would effect `insert`. 
https://github.com/dart-lang/sdk/blob/df6ea6326fb51e1f1deb8b5e25fd53afa9f1d605/sdk/lib/collection/list.dart#L283-L290
Also, some methods seem to be using the iterator to loop. Like the `insert` method (example below). While I prefer this syntax, is it in keeping with the description to read using only `length` and `operator`? 
https://github.com/dart-lang/sdk/blob/df6ea6326fb51e1f1deb8b5e25fd53afa9f1d605/sdk/lib/collection/list.dart#L525-L534
I actually much prefer the looping using iterator, and prefer it if the methods were written using it's declarative style (as opposed to the imperative style required by limiting to using `operator[]` and `length`. But just want to make sure it's in keeping with the documentation.

mit-mit:
Review expected in https://dart-review.googlesource.com/c/sdk/+/280075

asashour:
@jarrodcolburn there are test failures, e.g. [here](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8790438494412655633/+/u/test_results/new_test_failures__logs_).

Please update the expectation files in the PR.

Another way is to use Gerrit to popose changes as described [here](https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md#setting-up-the-environment).

asashour:
https://dart-review.googlesource.com/c/sdk/+/283682

which includes additional places than the ones in the PR, and updates the `front_end` expectations.

PR #51191, opened 2023-01-31
Suppress clang warning for musl libc inline

ntkme:
Previously in https://github.com/dart-lang/sdk/pull/51044 `-Werror` was removed when building with musl libc. It turns out `CMSG_NXTHDR` from musl is the only thing that causes warnings for comparing signed with unsigned, therefore it seems better to suppress the warning inline rather than disable `-Werror`.

cc @mraleph

### DIFF
--- a/runtime/bin/socket_base_posix.cc
+++ b/runtime/bin/socket_base_posix.cc
@@ -151,13 +151,27 @@ intptr_t SocketBase::ReceiveMessage(intptr_t fd,
   size_t num_messages = 0;
   while (cmsg != nullptr) {
     num_messages++;
+#ifndef __GLIBC__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wsign-compare"
+#endif
     cmsg = CMSG_NXTHDR(&msg, cmsg);
+#ifndef __GLIBC__
+#pragma clang diagnostic pop
+#endif
   }
   (*p_messages) = reinterpret_cast<SocketControlMessage*>(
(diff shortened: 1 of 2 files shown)


runtime/bin/socket_base_posix.cc
```diff
@@ -151,13 +151,27 @@ intptr_t SocketBase::ReceiveMessage(intptr_t fd,
   size_t num_messages = 0;
   while (cmsg != nullptr) {
     num_messages++;
+#ifndef __GLIBC__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wsign-compare"
+#endif
     cmsg = CMSG_NXTHDR(&msg, cmsg);
```
mraleph, line 158: Consider introducing a macro that wraps `CMSG_NXTHDR`, e.g. 

```cpp
// We wrap CMSG_NXTHDR to suppress sign-compare warnings which occur on musl.
#define NEXT_HEADER(mhdr, cmsg)                     \ 
  _Pragma("clang diagnostic push")                   \
  _Pragma("clang diagnostic ignored -Wsign-compare") \
  CMSG_NXTHDR(mhdr, cmsg)                            \
  _Pragma("clang diagnostic pop")
```

Instead of suppressing the error around each call site.

mraleph (commented):
small suggestion

runtime/bin/socket_base_posix.cc
ntkme: @mraleph Good point. Updated PR.

mraleph:
LGTM. CQ is running. 

Thanks for the contribution!

Issue #51163, opened 2023-01-30
Support for more default URI ports in `Uri` class


JKRhb:
The `Uri` class currently only supports the default ports for the `http` and `https` schemes:

https://github.com/dart-lang/sdk/blob/50b94ef9fc24302401e0511fcb5e73e50f85e25c/sdk/lib/core/uri.dart#L1718-L1723

For a library I am contributing to, it would be handy if more URI schemes could be added here. Would you be fine with turning the checks in this method into a switch statement with the possibility to add more URI schemes? If so, I would open a PR/hand in a patch with an initial proposal.

mit-mit:
cc @brianquinlan @lrhn

lrhn:
We try not to have to recognize too may schemes, currently just `package:`, `data:`, `http:`, `https:` and `file:`.

It's easy to add more default ports, but I don't want to have to support any other special-cases of other schemes (like special canonicalization rules or path interpretations). They have to be completely standard hierarchical URIs with host names.
Adding some support for a scheme, but not fully supporting the scheme's special cases, can be deceptive.

So, which that in mind, which schemes do you propose to add default ports for?

JKRhb:
Thank you, @Irhn, for the quick response and your explanations!

> So, which that in mind, which schemes do you propose to add default ports for?

My use-cache is probably a bit niche and concerns CoAP ([RFC 7252](https://www.rfc-editor.org/rfc/rfc7252)), a web transfer protocol for IoT scenarios, which is quite similar to HTTP, but more resource efficient. If I see it correctly, in order to support this protocol and its different URI schemes, adding the default ports would probably be sufficient (see [section 6 of RFC 7252 for more information](https://www.rfc-editor.org/rfc/rfc7252#section-6)). However, I could understand if you want to focus on more commonly used protocols to keep things as maintainable as possible.

> In practice, I'd probably just support ws and wss, which have ports 80 and 443 too, and not get into too many "old" protocols.

Adding additional support for these two URI schemes definitely sounds good :)

JKRhb:
I created #51203 (https://dart-review.googlesource.com/c/sdk/+/280215) with an initial proposal which could be used for further discussing this issue.

lrhn:
The implementation looks fine.

My main worry is that it feels arbitrary. Why these? What is the protocol for adding more in the future?
Should we just take the entire "official" list of default numbers once and for all? (And again, when and how do we update?)

[...]

PR #51203, opened 2023-02-01
Add support for more default URI ports

JKRhb:
This PR contains an initial proposal for the default URI port additions discussed in #51163.

As mentioned by me in https://github.com/dart-lang/sdk/issues/51163#issuecomment-1410183342, the proposal also contains the default ports for CoAP (see [RFC 7252](https://www.rfc-editor.org/rfc/rfc7252) for CoAP over UDP and [RFC 8323](https://www.rfc-editor.org/rfc/rfc8323) for CoAP over TCP, TLS, and WebSockets). Since CoAP is designed to be very closely aligned with HTTP, adding the corresponding port numbers should not cause any harm and should be compatible right away. However, let me know if you would prefer to see these additions removed.

### DIFF
--- a/sdk/lib/core/uri.dart
+++ b/sdk/lib/core/uri.dart
@@ -1720,30 +1720,10 @@ class _Uri implements Uri {
     // TODO: Magic numbers could be replaced with constants
     switch (scheme) {
       case "http":
-        return 80;
-      case "https":
-        return 443;
       case "ws":
         return 80;
+      case "https":
       case "wss":
-        return 443;
-      case "nntp":
-        return 119;
-      case "sftp":
-        return 22;
-      case "gopher":
-        return 70;
-      case "coap":
-        return 5683;
-      case "coaps":
-        return 5684;
-      case "coap+tcp":
-        return 5683;
-      case "coaps+tcp":
-        return 5684;
-      case "coap+ws":
-        return 80;
-      case "coaps+ws":
         return 443;
     }
     return 0;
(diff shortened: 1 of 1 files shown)


mit-mit:
CL is https://dart-review.googlesource.com/c/sdk/+/280215

sdk/lib/core/uri.dart
```diff
@@ -1717,8 +1718,16 @@ class _Uri implements Uri {
 
   /// The default port for the scheme of this Uri.
   static int _defaultPort(String scheme) {
-    if (scheme == "http") return 80;
-    if (scheme == "https") return 443;
+    switch (scheme) {
+      case "http":
+        return 80;
```
lrhn, line 1723: I'd combine these into one case:
```dart
  switch (scheme) {
    case "http":
    case "ws":
      return 80;
    case "https":
    case "wss":
       return 443;
  }
```

sdk/lib/core/uri.dart
```diff
@@ -1717,8 +1718,14 @@ class _Uri implements Uri {
...
   /// The default port for the scheme of this Uri.
   static int _defaultPort(String scheme) {
-    if (scheme == "http") return 80;
-    if (scheme == "https") return 443;
+    switch (scheme) {
+      case "http":
+      case "ws":
+        return 80;
+      case "https":
+      case "wss":
```
kevmoo, line 1726: Points for adding links to relevant RFCs here

JKRhb:
@lrhn Since this approach won't work, should we close this PR and the corresponding CL? Or is there anything we could still do here?

lrhn:
Let's close the pull request, and keep the issue open as a reminder to consider adding ways to add default ports for other protocols, locally or globally.

Even if it's just an extra optional `Map<String, int> defaultPorts` parameter to `Uri.parse` and other methods where it'd matter.

PR #51333, opened 2023-02-09
`identical`: document `null` cases

gmpassos:


lrhn:
There are many examples we can add. I'm not sure these are essential.

gmpassos:
> There are many examples we can add. I'm not sure these are essential.

There's NO documentation of the behavior with `null`. Until it's properly documented the behavior with `null` is undefined and undocumented (the developer should guess/test the behavior).

mraleph:
> There's NO documentation of the behavior with null.

`null` is just an instance of `class Null`, which is documented [here](https://api.dart.dev/dev/3.0.0-365.0.dev/dart-core/Null-class.html). It's not a special value and consequently it does not really need any special documentation in `identical(...)`.

gmpassos:
Almost every developer ignores that 'null' is an object in Dart, specially because in other languages null have many different implementations.

It's not clear for an average developer the real behavior of 'identical' for null values. Also other functions/classes have special treatments for primitives and null (like Expando).

Note that documentation should work very well to guide developers that are below average. 

IMHO the main point of view of the documentation is to help the developers. I don't think that is productive to remove relevant information just because is defined in another location not clear for an average developer.

I really hope to see some documentation for null values ​​added to the 'identical' function.

PR #51370, opened 2023-02-11
Add problemJson, problemXml content types

bytes7bytes7:
Add "application/problem+json" and "application/problem+xml" as static fields to ContentType.
These content types are described here https://www.rfc-editor.org/rfc/rfc7807


kevmoo:
We want to minimize the number of content types on this class. Thanks for your contribution.

PR #51494, opened 2023-02-22
Replace flutter/plugins analysis

stuartmorgan-g:
Replaces the flutter/plugns with analysis of flutter/packages, now that flutter/plugins has been merged into flutter/packages and will no longer be updated.

Also switches to calling the tool script directly, since the wrapper uses branch-based behavior that is no longer correct for this usage. This version matches the invocation used in flutter/flutter, so breakage due to repo changes is unlikely.

### DIFF
--- a/tools/bots/flutter/analyze_flutter_packages.sh
+++ b/tools/bots/flutter/analyze_flutter_packages.sh
@@ -4,7 +4,7 @@
 # for details. All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
 
-# Analyze Dart code in the flutter/plugins repo.
+# Analyze Dart code in the flutter/packages repo.
 
 set -e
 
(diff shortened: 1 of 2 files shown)


tools/bots/flutter/analyze_flutter_packages.sh
```diff
@@ -26,15 +26,15 @@ git clone --single-branch -vv https://github.com/flutter/flutter
...
+git clone --single-branch -vv https://github.com/flutter/packages
+cd packages
 
 # validate the tool's source
 (cd script/tool; dart pub get)
 (cd script/tool; dart analyze --fatal-infos)
 
 # Invoke the repo's analysis script.
-./script/tool_runner.sh analyze \
+dart run script/tool/bin/flutter_plugin_tools.dart analyze \
```
stuartmorgan, line 38: If we have any issues with this in the future, I'll create a custom wrapper called something like `script/out_of_repo_analysis.sh`, and update here and flutter/flutter to use it.

PR #51555, opened 2023-02-28
Added Iterable<Comparable> Extensions: .min and .max

ChristianKleineidam:
I created .min and .max Iterable<Comparable> extensions. I used the fact that NaN != NaN to make the functions return the same results as the existing min() and max() functions.

See previous discussion at https://github.com/dart-lang/sdk/pull/42000 and https://github.com/dart-lang/sdk/issues/41902


kevmoo:
@devoncarew @natebosch @lrhn – are we keeping these in pkg:collection – I could go either way here.

CC @mit-mit too

natebosch:
Thanks for the PR.
I think we should keep these in `package:collection`.

https://pub.dev/documentation/collection/latest/collection/IterableComparableExtension.html

Issue #51535, opened 2023-02-26
unbundling dependencies (distro packaging)


selfisekai:
port analogical solutions from chromium:
- tool for replacing normal gn build files with unbundle ones, for use in such environments (engine already had it for gyp, but it doesn't use gyp anymore)
- shim headers

I'm working on the same issue in flutter engine as well (flutter/flutter#121106, flutter/buildroot#687) - should at least some of the replacement build files be in a repository shared with flutter? (some are specific to flutter engine, some are shared with dart)

sortie:
Hi, I build the release engineering infrastructure for Dart and I'd love for Dart to better fit distros, I'd be interested in this work and potentially helping out. I'll look a bit more into your links. What sort of specific changes did you have in mind?

sortie:
This sounds doable yeah. We have a high level tools/build.py entry point that invokes gn so I'm sure it could have a new option that makes it unbundle the dependencies, perhaps by passing a parameter down through the gn logic. I think that's a cleaner solution than invoking a script to rewrite the tree :)

sortie:
Do you want to give this work an initial try? I'm happy to do the code reviews to make this happen :)

Something like this is also on our roadmap to improve the build-from-source journey although it's a little down the road

selfisekai:
what about flutter engine in this case? also the build workflows for dependencies are coming from other projects, as I see: zlib, icu and tcmalloc from chromium, protobuf from fuchsia

sortie:
I'm not sure about Flutter. Their infrastructure is a bit ever-changing and we constantly roll Dart into it via some complicated multi-headed solutions. If we at least solve the problem for Dart, we can see what it would take to do the equivalent work for Flutter. We're happy to try to collaborate with our friends in Chromium and Fuchsia if we need to amend or fork shared code :)

selfisekai:
@sortie submitted #51612 and #51614 so far.

I've got changes for icu done as well, but:
1. `use_system_icu=true` without removing some files from `//third_party/icu/` ends up with linker errors. chromium has a tool for removing those - should I do the same here?
2. I've been working around the need for depot_tools by teapot_tools, but there's no teapot git-cl, so I must figure this out one way or another

selfisekai:
@sortie submitted icu changes: https://chromium-review.googlesource.com/c/chromium/deps/icu/+/4307370

selfisekai:
@sortie I've tried to submit zlib changes, but I'm failing with this:
```
[...]

PR #51614, opened 2023-03-02
fix //build/config/linux/pkg_config.gni

selfisekai:
part of #51535

subprocess.check_output returns bytes, not str

it doesn't seem to be used yet in the project, so I assume it doesn't work for anyone. not sure how did it land here if it doesn't work (returned type changed in python 3?)

### DIFF
--- a/build/config/linux/pkg-config.py
+++ b/build/config/linux/pkg-config.py
@@ -161,7 +161,7 @@ try:
     # For now just split on spaces to get the args out. This will break if
     # pkgconfig returns quoted things with spaces in them, but that doesn't seem
     # to happen in practice.
-    all_flags = flag_string.strip().split(' ')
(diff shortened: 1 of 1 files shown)


sortie:
Thanks :)

Do add me as a reviewer if you're able to rather than merely a cc :)

If you're able to use gerrit and `depot_tools`, consider going through `git cl upload` directly instead of using pull requests since it's the proper workflow

selfisekai:
I got git-cl working after submitting these 2 PRs. should I also add you as reviewer on chromium changes? what if I can't submit changes on some repositories? (namely, `chromium/src/third_party/zlib`)

sortie:
I don't own the chrome repositories so I can't review other there but feel free to cc me if I have an user account there.

PR #51742, opened 2023-03-15
[ANTLR] Move optional trailing question mark up.

### LINKED ISSUE #51706: [ANTLR] Some nullable types are inheritable.
There are several places where `typeNotVoidNotFunction` makes it possible for nullable types to be extended/mixed-in/implemented or be part of a `mixinApplication`.

It looks like extending nullable types is not intended to be supported and is a compile-time error because the following does not compile:

```dart
//   A class can't extend a nullable type.
//                vvvvvvvvv
class Foo extends Function? {}
```

I think there is a simple adjustment that would allow the grammar to specify that some places don't expect nullable types by making them a syntax error i.e. by replacing the [`typeNotVoidNotFunction`](https://github.com/dart-lang/sdk/blob/6c7f3ab27d68268dcbe34cbb805b324094fe508d/tools/spec_parser/Dart.g#L1409-L1413) where nullable types are not expected with one that does not have trailing questionmarks.

I found the following places that look like they are not intended to expect nullable types:

https://github.com/dart-lang/sdk/blob/6c7f3ab27d68268dcbe34cbb805b324094fe508d/tools/spec_parser/Dart.g#L557-L559

https://github.com/dart-lang/sdk/blob/6c7f3ab27d68268dcbe34cbb805b324094fe508d/tools/spec_parser/Dart.g#L411-L415

and via `typeNotVoidNotFunctionList`

https://github.com/dart-lang/sdk/blob/6c7f3ab27d68268dcbe34cbb805b324094fe508d/tools/spec_parser/Dart.g#L398-L400

https://github.com/dart-lang/sdk/blob/6c7f3ab27d68268dcbe34cbb805b324094fe508d/tools/spec_parser/Dart.g#L394-L396

https://github.com/dart-lang/sdk/blob/6c7f3ab27d68268dcbe34cbb805b324094fe508d/tools/spec_parser/Dart.g#L389-L392

The LaTeX-based spec seems to mirror the behavior of the ANTLR-based spec (unless it specifies that this is not allowed in natural language). See: [typeNotVoidNotFunction](https://github.com/dart-lang/language/blob/2a3a253ca81d3cf0b78ad88e34aaaf6e185b5345/specification/dartLangSpec.tex#L20189-L20190) and e.g. [superclass](https://github.com/dart-lang/language/blob/2a3a253ca81d3cf0b78ad88e34aaaf6e185b5345/specification/dartLangSpec.tex#L4502-L4504)

modulovalue:
Move optional trailing question mark in typeNotVoidNotFunction to where typeNotVoidNotFunction is being referenced from except for places that are not expected to take nullable types.

Closes #51706

For reference: [here's](https://github.com/dart-lang/language/blob/abca38c39e9bea0af2ff47fce55d6686207178e7/specification/dartLangSpec.tex#L20180-L20187) the matching grammar snippet in the language spec.

### DIFF
--- a/tools/spec_parser/Dart.g
+++ b/tools/spec_parser/Dart.g
@@ -1397,18 +1397,18 @@ type
 typeNotVoid
     :    functionType '?'?
     |    recordType '?'?
-    |    typeNotVoidNotFunction
+    |    typeNotVoidNotFunction '?'?
     ;
 
 typeNotFunction
-    :    typeNotVoidNotFunction
+    :    typeNotVoidNotFunction '?'?
     |    recordType '?'?
     |    VOID
     ;
 
 typeNotVoidNotFunction
-    :    typeName typeArguments? '?'?
-    |    FUNCTION '?'?
+    :    typeName typeArguments?
+    |    FUNCTION
     ;
 
 typeName


modulovalue:
@copybara-service

PR #51744, opened 2023-03-15
[ANTLR] Transform recursive definition of onParts to use iteration ...

modulovalue:
... to match the [language spec](https://github.com/dart-lang/language/blob/master/specification/dartLangSpec.tex#L18224-L18228).

### DIFF
--- a/tools/spec_parser/Dart.g
+++ b/tools/spec_parser/Dart.g
@@ -1274,17 +1274,12 @@ rethrowStatement
     ;
 
 tryStatement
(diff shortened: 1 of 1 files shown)


eernstg:
Oops, the comments here aren't transferred to the Gerrit CL, and I didn't see the reference to the language specification [here](https://github.com/dart-lang/sdk/pull/51744#issue-1625711088), so there was no need for me to worry about making a lot of small changes to the language specification.

Anyway, it's done now. Thanks for noticing this discrepancy!

PR #51934, opened 2023-04-03
[SDK docs] update the SDK's API doc readme file

devoncarew:


devoncarew:
cc @sortie - it looks like copybara accidentally created two CLs from this PR? I don't know it this was a one time hiccup or something we've seen before.

sortie:
@devoncarew This is unfortunately a known issue. It's a problem on copybara's end.

Issue #49183, opened 2022-06-05
HttpClient sends invalid extension during TLS Client Hello causing failed connection to server


Programatic:
I originally wrote this issue on the flutter issue tracker because I did not know where to put it. I do believe this is here, so I will just copy it over here; https://github.com/flutter/flutter/issues/105381.

## Steps to Reproduce

1. Execute `dart run example.dart` on the code sample 
2. Press Button
3. See error in debug output

**Expected results:**
Request successfully goes through

**Actual results:**
Error fails with error 
```
I/flutter (29677): Error HandshakeException: Handshake error in client (OS Error:
I/flutter (29677): 	TLSV1_ALERT_DECODE_ERROR(tls_record.cc:594))
```

### Code sample

```dart
import 'dart:io';

void main() async {
  HttpClient client = HttpClient();
  Uri url = Uri.parse("https://10.0.0.224:8888/");
  try {
    await client.postUrl(url);
  } catch (error) {
    print("Error $error");
  }
}
```

### Logs

```
Error HandshakeException: Handshake error in client (OS Error:
	TLSV1_ALERT_DECODE_ERROR(tls_record.cc:594))
```

```
Dart SDK version: 2.18.0-edge.09967d8c255e6d996bbc419fe3a2148aaff5e9e4 (be) (Sun Jun 5 03:11:37 2022 +0000) on "linux_x64"
```

Currently, my server is using axum-server with rustls in order to create a backend for my application. In doing so, I only have an ip to connect with. However, using wireshark we can decode the TLSv1.2 Client Hello packet that the flutter application is sending.
![image](https://user-images.githubusercontent.com/41210901/172034650-79078c85-fd00-48a3-83cd-ce93426d7b60.png)
Specifically what we see here is that the HttpClient is sending a server_name extension. While this traditionally would serve to increase security, the way it is represented with the uri is a bug. 10.0.0.224 is clearly not a hostname, but instead an ip. Furthermore, according [to the ietf rfc, an ip is a distinct thing from a hostname and "Literal IPv4 and IPv6 addresses are not permitted in 'HostName'"](https://datatracker.ietf.org/doc/html/rfc6066). As a result, [when the server goes to decode the host name](https://github.com/rustls/rustls/blob/main/rustls/src/msgs/handshake.rs#L268), it fails to get a DnsNameRef. If we replay this packet with only the hostname changed to the hex of a string, such as 'google.com', then the connection succeeds. 

[...]

Issue #35462, opened 2018-12-20
setTrustedCertificates doesn't work. Causes HandshakeException all the time


bdshadow:
* Dart VM version: 2.1.0 (Tue Nov 13 18:22:02 2018 +0100) on "linux_x64"
* Fedora 28
----
https://gist.github.com/bdshadow/ad64d6f3660422e8ff06572f4bf6e7fd

```dart
import 'dart:convert';
import 'dart:io';

void main() {
  var url = "https://192.168.39.219:8443/version";
  SecurityContext context = SecurityContext();
  context.setTrustedCertificates("/home/dbocharo/.minikube/ca.crt");
  context.usePrivateKey("/home/dbocharo/.minikube/ca.key");
  HttpClient client = new HttpClient(context: context);
  client.getUrl(Uri.parse(url))
      .then((HttpClientRequest request) => request.close())
      .then((HttpClientResponse response) {
    response.transform(utf8.decoder).listen((contents) {
      print(contents);
    });
  });
}
```

 As you can see i try to connect to kubernetes run by minikube. But i always have "CERTIFICATE_VERIFY_FAILED: Hostname mismatch" error. The certificates are correct because:
- "curl --cert ~/.minikube/ca.crt --key ~/.minikube/ca.key https://192.168.39.219:8443/version" runs fine from the terminal
-  When adding this "ca.crt" to android i'm able to do the similar stuff from android using java without getting handshake errors. 

I also tried experimenting with apiserver.crt, client.crt and their keys (though it doesn't make much sense because of the above items). I found that people have the same problem https://stackoverflow.com/questions/46135757/ssl-handshake-error-on-self-signed-cert-in-flutter?rq=1 and there is no a proper reply. I also tried using IOClient from `package:http/io_client.dart` - the error is the same. I googled everything i could imagine :) Also i found here https://github.com/dart-lang/sdk/issues/27246 that @whesse  is your ssl expert :) It would be great if William or anybody else could help me! Thank you in advance!

bdshadow:
@whesse thank you very much for your reply.
I don't think that i need to set the certificate chain. I don't think so, because of the reason i described:
- `curl --cert ~/.minikube/ca.crt --key ~/.minikube/ca.key https://192.168.39.219:8443/version` works fine. Nothing else is needed. The key and certificate are fine.
- and these certificate is ok for them same app in android (java). 
So it must be no difference with dart. At least i don't see a reason why would dart client need more than other clients

[...]

PR #52118, opened 2023-04-20
Fix SSL certificate check when hostname is an IP address

vially:
The previous logic in `SSLFilter::Connect` was assuming the host is always a DNS name so it was unconditionally calling [`X509_VERIFY_PARAM_set1_host`](https://github.com/dart-lang/sdk/blob/9bb4402abdb5784281d43b99f4b45ab852f9eb47/runtime/bin/secure_socket_filter.cc#L543-L544) even for hosts that were actually IP addresses. This in turn was causing a `CERTIFICATE_VERIFY_FAILED: Hostname mismatch(handshake.cc:393)` exception when a client attempted to connect to an IP address (rather than a DNS name) host.

This CL fixes that by checking whether the host is an IP address or a DNS name and calling either `X509_VERIFY_PARAM_set1_ip_asc` or `X509_VERIFY_PARAM_set1_host` depending on the result.

PS: ~The certificates had to be regenerated because the previous configuration in [`sample_certificate_v3_extensions`](https://github.com/dart-lang/sdk/blob/9bb4402abdb5784281d43b99f4b45ab852f9eb47/tests/standalone/io/sample_certificate_v3_extensions#L23-L24) had the `127.0.0.1` and `::1` IP addresses stored as *DNS* subject alternative names. This meant that the newly introduced tests would still pass *even with the previous logic*, but for the wrong reasons.~ EDIT: Regenerating the certificates has been submitted as a separate [change](https://dart-review.googlesource.com/c/sdk/+/300000).

Bug: #35462, #49183

### DIFF
--- a/runtime/bin/secure_socket_filter.cc
+++ b/runtime/bin/secure_socket_filter.cc
@@ -13,6 +13,7 @@
 #include "bin/lockers.h"
 #include "bin/secure_socket_utils.h"
 #include "bin/security_context.h"
+#include "bin/socket_base.h"
 #include "platform/syslog.h"
 #include "platform/text_buffer.h"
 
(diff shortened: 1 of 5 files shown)


brianquinlan:
I've left a comment on the gerrit review: https://dart-review.googlesource.com/c/sdk/+/296643

mraleph:
The comment is actually on https://dart-review.googlesource.com/c/sdk/+/296722. This is rather puzzling - we seem to import the same PR twice under different CL numbers. (/cc @athomas)

mraleph:
Regarding Mac OS X failures. Among other things please make sure that you adhere to https://support.apple.com/en-us/HT210176 when generating test certificates - that's the most common source of issues.

vially:
@brianquinlan @mraleph Thanks for the prompt review.

I've dropped the certificate regeneration from this initial set of changes. Unfortunately, as I've mentioned above, this means the newly introduced tests in `secure_server_client_certificate_test.dart` would still pass even without the logic fixes from this pull-request.

I'm still planning to submit a separate pull-request which removes the IP addresses from the [DNS alt-names](https://github.com/dart-lang/sdk/blob/9bb4402abdb5784281d43b99f4b45ab852f9eb47/tests/standalone/io/sample_certificate_v3_extensions#L23-L24) and regenerate all the test certificates once I figure out how to that without breaking the tests.

However, I don't think it's worth blocking *this* pull-request until the certificates are regenerated, because they are arguably two different bugs:
- SSL check is broken when hostname is an IP address (fixed by this pull-request)
- testing SSL certificates are misconfigured to contain IP addresses in the DNS field (hopefully fixed in the next future pull-request)

vially:
I have submitted a separate change for regenerating the test certificates at: https://dart-review.googlesource.com/c/sdk/+/300000.

PR #52293, opened 2023-05-07
[ANTLR] Various changes.

### LINKED ISSUE #3039: [grammar] `selector`/`typeArgument`-related ambiguity.
A recent change made it possible to give function invocations type arguments without having to invoke the function with value arguments.

```dart
void main() {
  final a = foo<int>;
  print(a); 
}

void foo<T>() {}
```
It looks like this change has lead to a new production `typeArguments` in the selector production rule in the ANTLR-based grammar spec.

https://github.com/dart-lang/sdk/blob/b06b994a4eea9eb88ef5b40060bd8ff16b848868/tools/spec_parser/Dart.g#L971-L982

It looks like the language spec grammar snippets have not yet been adjusted to reflect this change i.e. the `selector` grammar snippet does not contain a `typeArguments` production.

https://github.com/dart-lang/language/blob/a6209672d4bc11442ef64561dc612edb7accfe5f/specification/dartLangSpec.tex#L16270-L16277

If we were to add a `typeArguments` production to the `selector` production rule, then I think this would give `id<int>()` 2 parse trees:

```
                ...
1. primary              selector*
   identifier           selector      selector
   identifier           typeArguments argumentPart
   identifier           typeArguments typeArguments? arguments
   identifier           typeArguments      null      arguments
      'id'                  '<int>'                     '()'


                ...
2. primary              selector*
   identifier           selector
   identifier           argumentPart
   identifier           typeArguments arguments
      'id'                  '<int>'      '()'
```

I'm wondering what a grammar that does not have this ambiguity would look like?

modulovalue:
This PR proposes the following changes:

### Convert left-recursion in `cascade` to iteration.

Since one of the goals of the grammar is to be a reference for recursive-descent-style parsers, I believe this is a desirable change.

### Convert right-recursion in `unaryExpression` to iteration.

In the spirit of the PR that added the previous changelog entry, this simplifies the grammar by making it explicit that unary expressions have a regular structure.
 
### Remove the dependency on order for disambiguating `<typeArguments> <arguments>` and `<arguments>` in selectors.

Closes https://github.com/dart-lang/language/issues/3039

### Fix inconsistent order-dependent disambiguation between methods and external constructors.

The following program
```dart
class Bar {
  Bar();
  external Bar();
}
```

has the following parse tree:
```
<startSymbol>
┗━ <libraryDefinition>
  ┣━ <metadata>
  ┣━ <topLevelDefinition>
  ┃  ┗━ <classDeclaration>
  ┃    ┣━ <classModifiers>
  ┃    ┣━ 'class'
  ┃    ┣━ <typeWithParameters>
  ┃    ┃  ┗━ <typeIdentifier>
  ┃    ┃    ┗━ 'Bar'
  ┃    ┣━ '{'
  ┃    ┣━ <metadata>
  ┃    ┣━ <classMemberDeclaration>
  ┃    ┃  ┣━ <declaration>
  ┃    ┃  ┃  ┗━ <functionSignature>
  ┃    ┃  ┃    ┣━ <identifier>
  ┃    ┃  ┃    ┃  ┗━ 'Bar'
  ┃    ┃  ┃    ┗━ <formalParameterPart>
  ┃    ┃  ┃      ┗━ <formalParameterList>
  ┃    ┃  ┃        ┣━ '('
  ┃    ┃  ┃        ┗━ ')'
  ┃    ┃  ┗━ ';'
  ┃    ┣━ <metadata>
  ┃    ┣━ <classMemberDeclaration>
  ┃    ┃  ┣━ <declaration>
  ┃    ┃  ┃  ┣━ 'external'
  ┃    ┃  ┃  ┗━ <constructorSignature>
  ┃    ┃  ┃    ┣━ <constructorName>
  ┃    ┃  ┃    ┃  ┗━ <typeIdentifier>
  ┃    ┃  ┃    ┃    ┗━ 'Bar'
  ┃    ┃  ┃    ┗━ <formalParameterList>
  ┃    ┃  ┃      ┣━ '('
  ┃    ┃  ┃      ┗━ ')'
  ┃    ┃  ┗━ ';'
  ┃    ┗━ '}'
  ┗━ '<EOF>'
```

Notice how the external method is recognized as being a constructor. This is inconsistent with how non-external constructors are being recognized. This PR fixes this to treat external methods and constructors the way that non-external methods and constructors are being treated.

### Minor formatting changes.

I think the changes related to formatting are self-explanatory. They are meant to make those parts match how the rest of the file is formatted.

### DIFF
--- a/tools/spec_parser/Dart.g
+++ b/tools/spec_parser/Dart.g
@@ -3,6 +3,16 @@
 // BSD-style license that can be found in the LICENSE file.
 
 // CHANGES:
+// v0.34
+// - Convert left-recursion in `cascade` to iteration.
+// - Convert right-recursion in `unaryExpression` to iteration.
+// - Add await and negation as prefixes to prefixed super unaryExpressions
+//   to match the implementation.
+// - Remove the dependency on order for disambiguating
+//   `<typeArguments> <arguments>` and `<arguments>` in selectors.
+// - Fix inconsistent order-dependent disambiguation between external
+//   methods and external constructors.
+// - Minor formatting changes.
 //
 // v0.33 This commit does not change the derived language at all. It just
 // changes several rules to use the regexp-like grammar operators to simplify
@@ -71,7 +81,7 @@
 // `builtinIdentifier` and `reservedWord`; update `typeAlias` to enable
 // non-function type aliases; add missing `metadata` to formal parameter
 // declarations; correct `symbolLiteral` to allow `VOID`;
-
+//
 // v0.12 (82403371ac00ddf004be60fa7b705474d2864509) Cf. language issue #1341:
 // correct `metadata`. Change `qualifiedName` such that it only includes the
 // cases with a '.'; the remaining case is added where `qualifiedName` is used.
@@ -118,11 +128,11 @@
 
 grammar Dart;
 
-@parser::header{
+@parser::header {
 import java.util.Stack;
 }
 
-@lexer::header{
+@lexer::header {
 import java.util.Stack;
 }
 
@@ -170,7 +180,7 @@ import java.util.Stack;
   }
 }
 
-@lexer::members{
+@lexer::members {
   public static final int BRACE_NORMAL = 1;
   public static final int BRACE_SINGLE = 2;
   public static final int BRACE_DOUBLE = 3;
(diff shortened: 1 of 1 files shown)


eernstg:
Hi @modulovalue, thanks for the input!

This PR makes a lot of changes to the grammar, and this is usually something that involves discussions for quite a while. For instance, the `<cascade>` rules were adjusted at some point such that they would allow for a specification of the semantics of null shorting (the fact that `a?.b.c` evaluates `a` and then evaluates to null, rather than evaluating to null and then throwing because `null.c` fails). This means that it's going to be a rather involved discussion about the consequences (e.g., for the ability to keep track of the connection to the language specification, and to the implementation) if we're going to change the structure of the grammar, even in the case where it is rather easy to prove that it doesn't change the language which is derivable from that grammar.

Could you say something about your motivation for proposing these changes?

devoncarew:
(assigning a reviewer to remove this from the triage queue)

mraleph:
@eernstg ping?

modulovalue:
Quote:

> This PR makes a lot of changes to the grammar [...]

I was planning to revisit this. I will close it until then.

Issue #40614, opened 2020-02-13
Clean up API boundaries when null-safety is non-optional.


lrhn:
Current null-safe platform libraries have seemingly redundant null-checks in many places.

A parameter of type `int` should only contain integers, but if the function is called from a non-null-safe library, it can get away with passing `null` values anyway. 
To improve the error message in that case, we have inserted test which check that the value is indeed not `null` (mostly using `ArgumentError.checkNotNull` calls).

When null safety becomes non-optional (when language versions before the introduction of null safety stops being supported), those extra checks should be removed again.

This issue should be kept open until this has happened.

asashour:
I think this should be part of one of Dart 3 milestones.

lrhn:
Most of the uses that I've touched are using the internal `checkNotNullable`, which should make them easy to remove.
We may need to keep supporting non-null-safe code a little past the 3.0 release, for technical reasons, but removing unnecessary checks should be on the list for when we turn it off for good.

eseidel:
Ran into one of these today.  I presume this is possible now?

PR #52396, opened 2023-05-14
Remove unnecessary null checks, now that null-safety is mandatory

mnordine:
See https://github.com/dart-lang/sdk/issues/40614

Seems this should have been part of Dart 3 release.

*Thanks for your contribution! Please replace this paragraph with a description of what this PR is changing or adding and why. List any relevant issues, and review the contribution guidelines below.*

---

#### Contribution guidelines:

---

### DIFF
--- a/pkg/analyzer/lib/src/dart/analysis/context_builder.dart
+++ b/pkg/analyzer/lib/src/dart/analysis/context_builder.dart
@@ -75,7 +75,6 @@ class ContextBuilderImpl implements ContextBuilder {
   }) {
     // TODO(scheglov) Remove this, and make `sdkPath` required.
     sdkPath ??= getSdkPath();
-    ArgumentError.checkNotNull(sdkPath, 'sdkPath');
     if (updateAnalysisOptions != null && updateAnalysisOptions2 != null) {
       throw ArgumentError(
           'Either updateAnalysisOptions or updateAnalysisOptions2 must be '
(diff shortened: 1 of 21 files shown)


devoncarew:
@lrhn for review (note that this PR should be reviewed / landed using the associated gerrit CL).

lrhn:
We still have a few cases of pre-2.12 code being run against the current SDK code, for annoying technical reasons.
I guess removing the checks should be safe enough. The existing code is presumably safe, so as long as no new pre-2.12 code is written, we should not have any needs for catching newly developed bugs.

mraleph:
@lrhn we should either land it or close it.

lrhn:
Or wait until we've gotten rid of all internal non-sound-null-safe code, which is what I'm leaning towards.
This is exactly the kind of error which will show up even in null-safe code if it's running in non-sound-null-safe mode.

(The alternate point is that the throw has been there for ages now, so no code is currently running which would hit it. Nobody will write *new* non-null-safe code, or add new non-sound null safe projects, so it *is* safe to remove the checks now. In either case, it needs to be merged to head.)

vsmenon:
@lrhn - can we either land or close this?

mit-mit:
I'm afraid this has gone stale. We were unfortunately not in a position to take this when it was submitted, and not that we're ready, the code has "drifted" too far in the meantime.

PR #52509, opened 2023-05-25
HttpOverrides support for WebSocket

tauu:
This PR fixes WebSocket.connect not using any HttpClient provided by HttpOverrides. Currently the HttpClient used for WebSocket connections is initialised in a static field. Consequently any HttpClient set in HttpOverrides after the first initialisation is not taken into account for WebSocket connections. This is changed by initialising a new HttpClient for each WebSocket connection.

As it is currently possible to set the user agent for the HttpClient used for the websocket request, the current behaviour is retained:
  - If a custom HttpClient is provided in the connect call, the userAgent field is not used.
  - If no custom HttpClient is provided in the connect call and the userAgent field was not set, the userAgent of the default HttpClient is not set.
  - If no custom HttpClient is provided in the connect call and the userAgent field was set, the userAgent of the default HttpClient is set.
This is accomplished by adding the private field _overrideUserAgent and setting it the first time the userAgent field ist set. This extra field is necessary, as null is a valid value for the userAgent property of a HttpClient and it is not possible to distinguish between _userAgent being initialised to null or it being set to null otherwise.

Two tests are included which provide two different overrides and verify that these lead to results for the same WebSocket.connect call.

---

### DIFF
--- a/sdk/lib/_http/websocket_impl.dart
+++ b/sdk/lib/_http/websocket_impl.dart
@@ -989,8 +989,6 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
   Timer? _closeTimer;
   _WebSocketPerMessageDeflate? _deflate;
 
-  static final HttpClient _httpClient = HttpClient();
-
   static Future<WebSocket> connect(
       String url, Iterable<String>? protocols, Map<String, dynamic>? headers,
       {CompressionOptions compression = CompressionOptions.compressionDefault,
@@ -1018,7 +1016,12 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
         path: uri.path,
         query: uri.query,
         fragment: uri.fragment);
-    return (customClient ?? _httpClient).openUrl("GET", uri).then((request) {
+    final client = customClient ?? HttpClient();
+    // The default userAgent is only overriden, if no custom HttpClient is set.
+    if (customClient == null && _overrideUserAgent) {
+      client.userAgent = _userAgent;
+    }
+    return client.openUrl("GET", uri).then((request) {
       if (uri.userInfo != null && uri.userInfo.isNotEmpty) {
         // If the URL contains user information use that for basic
         // authorization.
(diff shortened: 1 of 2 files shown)


brianquinlan:
I left some feedback here:
https://dart-review.googlesource.com/c/sdk/+/305620/2/sdk/lib/_http/websocket_impl.dart#b992

This is a potentially breaking change which needs to follow the [Dart SDK breaking change process](https://github.com/dart-lang/sdk/blob/master/docs/process/breaking-changes.md).

Would you like for me to deal with that bureaucracy?

tauu:
First of all, sorry for the slow response. I wrongly assumed, that I would get a notification from google source / copybara would relay comments to github. But apparently that did not happen 😅 and I was on vacation once the github comment appeared. Thank you for the review. I will address the comments in dart-review.

> Would you like for me to deal with that bureaucracy?
That would be amazing!!

mraleph:
@brianquinlan are you taking over this? Should this PR be closed?

brianquinlan:
@tauu 

I have a proposed set of simplified tests: https://dart-review.googlesource.com/c/sdk/+/305620/7/tests/standalone/io/http_override_websocket_test.dart

Let me know what you think.

sdk/lib/_http/websocket_impl.dart
```diff
@@ -1263,11 +1269,12 @@ class _WebSocketImpl extends Stream with _ServiceObject implements WebSocket {
     }
     return _sink.close();
   }
-
-  static String? get userAgent => _httpClient.userAgent;
+
```
albertms10, line 1272: Unwanted whitespace? 👀

brianquinlan:
Unfortunately, I don't think that this change is viable because it will break existing code. For example, `flutter test` uses `HttpOverrides` to replace `HttpClient` with a mock but using that mock in `WebSocket` will take away the ability to talk to VM services. See https://github.com/dart-lang/sdk/commit/1b8382b14d5cf24d6d5ebe3a2d74fba286c45a12

mraleph:
@brianquinlan could not we just fix the tests by undoing the override? These tests starting to fail is in spirit of the change / flutter test - we have discovered tests which are using actual network connection instead of a mocked connection. Test should be explicit about that.

brianquinlan:
@mraleph I'm not sure that I understand your question.

[`flutter_test` defines its own `HttpOverrides`](https://github.com/flutter/flutter/blob/0dc066b717a459f3544b2b074ffc35ba7c351e30/packages/flutter_test/lib/src/_binding_io.dart#L71) that disables all HTTP requests and [installs it globally](https://github.com/flutter/flutter/blob/0dc066b717a459f3544b2b074ffc35ba7c351e30/packages/flutter_test/lib/src/_binding_io.dart#L27)

 The tests were failing because the test infrastructure itself uses `WebSocket`, which will end up using the `_MockHttpClient`.

I can probably fix `flutter_test` but I'm worried about other code out there that consumes VM services and will be broken by this change.

PR #52515, opened 2023-05-25
Fix grammar docs for type aliases in AST

### LINKED ISSUE #35485: Doc comment on TypeAlias and subclasses regarding syntax is wrong.
Current syntax comment for TypeAlias:

> typeAlias ::= 'typedef' typeAliasBody
>
> typeAliasBody ::= classTypeAlias | functionTypeAlias

And for ClassTypeAlias:

> classTypeAlias ::= SimpleIdentifier TypeParameterList? '=' 'abstract'? mixinApplication
>
> mixinApplication ::= TypeName WithClause ImplementsClause? ';'

Which would suggest that a class type alias is declared with `typedef Foo = Bar with Baz;`. Which is incorrect; the keyword `class` is used, not `typedef`.

I think the best solution is to not specify any keyword in from of typeAliasBody; It should just be:

> typeAlias ::= classTypeAlias | functionTypeAlias

Then GenericTypeAlias specifies `metadata` as the first element, while the others do not...

shilangyu:

Noticed issues when researching this mystical "class alias" feature. There are a few issues still:

- The AST node `TypeAlias` has a getter called `typedefKeyword` which says it will return the token for the 'typedef' token. This, however, is not true. For a class type alias this will return the token for 'class'. Is that something that should be fixed (at least the docs)?
- The language spec calls this feature "mixin classes" or "mixin application class" https://github.com/dart-lang/language/blob/cc2c933b1ae5a2c837f89ca4490040ff158a676f/specification/dartLangSpec.tex#L5428. In the AST this is called "ClassTypeAlias". This is somewhat confusing, especially when trying to find any documentation for this feature. Which, to my knowledge is very scares. I could not find any mention of it on dart.dev. The lang spec is also outdated, since it does not mention dart 3 class modifiers.

Maybe I am nitpicking here, but is this really a "type alias"? It introduces a new nominal type which exists at runtime, unlike standard `typedef`s. TBH it is quite unclear what this feature is for exactly.

Let me know if I should convert any of the above to an issue.

closes #35485

### DIFF
--- a/pkg/analyzer/lib/src/dart/ast/ast.dart
+++ b/pkg/analyzer/lib/src/dart/ast/ast.dart
@@ -3152,8 +3152,7 @@ abstract final class ClassTypeAlias implements TypeAlias {
 /// A class type alias.
 ///
 ///    classTypeAlias ::=
-///        [SimpleIdentifier] [TypeParameterList]? '=' classModifiers
-///        mixinApplication
+///        classModifiers 'class' [SimpleIdentifier] [TypeParameterList]? '=' mixinApplication
 ///
 ///    classModifiers ::= 'sealed'
 ///      | 'abstract'? ('base' | 'interface' | 'final')?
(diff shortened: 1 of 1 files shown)


devoncarew:
It looks like this PR may be stale now?

shilangyu:
I am waiting for a review still. All comments have been addressed.

bwilkerson:
Thanks for the ping! Sorry for the delay, I had missed the email indicating that you'd addressed my comments.

shilangyu:
Looks like there are conflicts now, I'll rebase the branch

PR #52563, opened 2023-05-30
Add missing void in function declarations to avoid warning

kristofb:
at build time: "This function declaration is not a prototype"

*Building an application which includes dart_api.h produces warnings with XCode:

This function declaration is not a prototype

Only three functions prototypes are missing (void).
.*


sortie:
Copybara is working to avoid the duplicate pull request events. Perhaps their attempts lost this pull request. I'll let them know about it so they can take a look. Otherwise this pull request looks ready with CLA.

@kristofb You can try to push (or force push) to your branch to trigger our copybara automation that imports this pull request into gerrit as a code review. Development of Dart is done on the https://dart-review.googlesource.com gerrit so the pull request must be imported before we can merge your changes. :)

PR #52666, opened 2023-06-10
Removed Non Null Check

aryan-more:
Removed unnecessary ArgumentError.checkNotNull as dart became null safe starting from Dart 2.12

### DIFF
--- a/sdk/lib/io/data_transformer.dart
+++ b/sdk/lib/io/data_transformer.dart
@@ -131,17 +131,17 @@ final class ZLibCodec extends Codec<List<int>, List<int>> {
 
   /// Get a [ZLibEncoder] for encoding to `ZLib` compressed data.
   ZLibEncoder get encoder => new ZLibEncoder(
-        gzip: false,
-        level: level,
-        windowBits: windowBits,
-        memLevel: memLevel,
-        strategy: strategy,
-        dictionary: dictionary,
-        raw: raw,
-      );
+      gzip: false,
+      level: level,
+      windowBits: windowBits,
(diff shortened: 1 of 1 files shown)


sdk/lib/io/data_transformer.dart
```diff
@@ -526,20 +510,17 @@ class _FilterSink extends ByteConversionSink {
         _sink.add(out);
       }
     } catch (e) {
-      // TODO(kevmoo): not sure why this isn't a try/finally
```
lrhn, line 529: Or why the sink isn't closed if it throws.
Looks weird. Consider keeping the comment, until someone investigates.

Maybe it assumes that only the sink `add` can throw, but that's optimistic, and why is the filter operation inside the `try` then?

lrhn (approved):

devoncarew:
@lrhn - you'll need to review the associated CL.

lrhn:
The file has some overly long lines, it probably needs to be `dart format`ted again.

sdk/lib/io/data_transformer.dart
aryan-more: Ok

aryan-more:
> The file has some overly long lines, it probably needs to be `dart format`ted again.

My vscode format settings messed up formatting.

aryan-more:
@lrhn why did PR got closed?

lrhn:
Success!
The associated [Gerrit CL](https://dart-review.googlesource.com/c/sdk/+/308421), created from this GitHub PR by the copybara script, was accepted and landed.

PR #52860, opened 2023-07-06
Fix Uri.toString() for empty string field

hamberluo:
`_initializeText()` should check if query and fragment string is empty.

### DIFF
--- a/sdk/lib/core/uri.dart
+++ b/sdk/lib/core/uri.dart
@@ -2935,11 +2935,11 @@ class _Uri implements Uri {
       _writeAuthority(sb);
     }
     sb.write(path);
-    if (_query != null)
+    if (_query != null && _query.isNotEmpty)
       sb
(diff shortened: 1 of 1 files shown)


lrhn:
The current behavior is deliberate.
You can disagree with the choice of distinguishing a missing query/fragment from an empty one, but the underlying syntax allows both, and so does the class, which is made so that parsing the output of `toString` gives back the exact same `Uri` again. Including preserving an empty query or fragment part.

hamberluo:
> The current behavior is deliberate. You can disagree with the choice of distinguishing a missing query/fragment from an empty one, but the underlying syntax allows both, and so does the class, which is made so that parsing the output of `toString` gives back the exact same `Uri` again. Including preserving an empty query or fragment part.

However, when you recreate a new one based on the existing URI, calling uri.path gives you a non-null empty string.

PR #52866, opened 2023-07-06
Fix docs of DateTime.from constructors

pourqavam:
The `DateTime.fromMillisecondsSinceEpoch` document says that the result of `1640979000000` is `2022-01-01 10:00:00.000Z` but the correct input value is `1641031200000` and the `DateTime.fromMicrosecondsSinceEpoch` document says that the result of `1640901600000000` is `2021-12-31 19:30:00.000Z` but the correct input value is `1640979000000000`.

---

### DIFF
--- a/sdk/lib/core/date_time.dart
+++ b/sdk/lib/core/date_time.dart
@@ -384,7 +384,7 @@ class DateTime implements Comparable<DateTime> {
   /// time zone (local or UTC).
   /// ```dart
   /// final newYearsDay =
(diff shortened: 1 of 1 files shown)


lrhn:
Thanks for the fix. 

It's an embarrasing mistake to make, and so very easily checkable. 

(Would it make sense to have a mode for our example code checker which detects a `print(...); // something` and checks that what is printed contains `something`. Not on by default, since not all print statement comments are inteded that way, but worth running occasionally, to catch examples like this.)

PR #52906, opened 2023-07-11
[vm/io] Fall back to IfIndex if Ipv6IfIndex is zero

jnschulze:
On Windows, network interfaces have two indices, `IfIndex` and `Ipv6IfIndex` (see https://learn.microsoft.com/en-us/windows/win32/api/iptypes/ns-iptypes-ip_adapter_addresses_lh).

If both IPv4 and IPv6 are enabled for a particular interface, `IfIndex` corresponds to `Ipv6IfIndex`.
For disabled protocols, however, the corresponding index is zero.

As Dart currently only respects the IPv6 index, `NetworkInterface.list` will return invalid indices for interfaces with IPv6 turned off which will make `RawDatagramSocket.joinMulticast` fail (even when attempting to join an IPv4 multicast group).

This PR adds a fallback to ensure that `NetworkInterface.index` is correct even if IPv6 is turned off.

---


mraleph:
cc @brianquinlan

brianquinlan:
Add some comments to https://dart-review.googlesource.com/c/sdk/+/313161

PR #53164, opened 2023-08-09
Add extra documentation to Enum to point to EnumName.

Hixie:
I keep forgetting you can use `.name` because the only sign that it exists in the `Enum` documentation is the in-passing mention that `Enum` has an extension, without mentioning why. This add an explicit callout to the `Enum` class definition.

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR, as far as I can tell.

### DIFF
--- a/sdk/lib/core/enum.dart
+++ b/sdk/lib/core/enum.dart
@@ -11,7 +11,7 @@ part of dart.core;
 /// Non-platform classes cannot extend or mix in this class.
 /// Concrete classes cannot implement the interface.
 ///
-/// The intentifier used to name an `enum` value is available as a [String],
+/// The identifier used to name an `enum` value is available as a [String],
 /// via the [EnumName.name] extension property on the `enum` value.
 @Since("2.14")
 abstract interface class Enum {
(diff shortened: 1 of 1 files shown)


sdk/lib/core/enum.dart
```diff
@@ -10,6 +10,9 @@ part of dart.core;
 /// introduced using an `enum` declaration.
 /// Non-platform classes cannot extend or mix in this class.
 /// Concrete classes cannot implement the interface.
+///
+/// The string name of a value is available via the `name` property
```
lrhn, line 14: I'd go with 
```dart
/// The intentifier used to name an `enum` value is available as a `String`, 
/// via the [`name`](EnumName.name) extension property on the `enum` value.
```
(If that's thow you link to other members without their full name in DartDoc. Where is the DartDoc syntax specification?)

Hixie: I'm not sure if that works, but I'm happy to change it to that if you would prefer. Let me know. :-)

mraleph:
@Hixie would you like to follow up on the comments or should we close the PR?

Hixie:
I was awaiting clarification about whether to apply @lrhn's suggestion or not. Is that the preferred syntax? It wasn't clear to me if that was what was being suggested or not.

Hixie:
I applied the change but without the part that's risky (which has the side-effect of being more explicit, which is probably good anyway).

mraleph:
@lrhn friendly ping

PR #53242, opened 2023-08-16
web_audio is not part sky_engine

rmasarovic:
- I want add web_audio to flutter sky_engine but first i need add build gni scripts to packages
- its reaction for this bug https://github.com/flutter/flutter/issues/49849 and comment [mraleph](https://github.com/mraleph) 
---

<details>

</details>


sdk/lib/indexed_db/indexed_db_sources.gni
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+indexed_db_sdk_sources = [
+    "dart2js/indexed_db_dart2js.dart",
+]
```
mraleph, line 7: Missing the new line here and in the svg_sources.gni below (make sure to run `gn format ...`) on these newly added files.

mraleph (commented):
Thanks! Could you just fix the formatting?

sigmundch:
Sorry I'm just noticing this PR.

It may be worth circling back on the conversation about how we want to expose these APIs. I'll add some comments to https://github.com/flutter/flutter/issues/49849 as well, but the quick summary is this:
* We'd like to use js-interop in the long run. This is important if we want flutter web plug-ins to work both in JavaScript and Web Assembly (dart2wasm doesn't support the `dart:html` and related libraries).
* Wnlike 3 years ago, today there are packages generated from IDLs that expose web-audio and other APIs directly with JS-interop. The only caveat is that those packages (e.g. `package:web`) are not stable yet. 

So the question is: Do we still make this change today and wait for `package:web` to become stable? Or do we still ban using these APIs and move start using that package early?

cc @srujzs @eyebrowsoffire @kevmoo

srujzs:
If we're going to need users to migrate a couple months from now anyways, it makes sense for me to say that we should hold off and have users use `package:web`/their own interop in the interim.

I don't know what the usage is like for `web_audio`, but I assume it's significantly lower than `dart:html`. If there are bugs either because usage is low or because the browsers have changed in the last 6 years, having users deal with debugging that when we're not planning on fixing it is a poor experience.

mraleph:
@sigmundch @srujzs I think it is worth highlighting the following: people can _already_ use these libraries with `dart2js` and probably with DDC because (last time I checked) Web compilers do *not* forbid them when compiling Flutter Web apps. It's only *analyzer* that is complaining that these libraries are unavailable in Flutter because of how SDK is wired up.

sigmundch:
Makes sense - I approved the PR in gerrit, so feel free to move forward.

Just wanted to clarify to others on this thread that if you use `dart:web_audio`, you may find that you will need to migrate to `package:web` in the future. While `dart:web_audio` won't be deprecated that quickly, it wont be supported in all scenarios (for example `dart2wasm` will not support it). If adopting `package:web` is possible for your use case, I'd recommend giving it a try once it's released with a regular (non beta) version number.

Issue #49166, opened 2022-06-02
`.dart` and `.dartServer` is stored in home folder on macOS


probablykasper:
On macOS, Dart stores data in the unconventional folders `~/.dart` and `~/.dartServer`. This contributes to a mess where app files are scattered everywhere, the home folder is cluttered, and backups get filled with unnecessary junk.

This should be fixed and moved to proper paths, like `~/Library/Application Support`, `~/Library/Caches` and `~/Library/Preferences`.

* Dart SDK version: 2.17.1 (stable) (Tue May 17 17:58:21 2022 +0000) on "macos_x64"
* macOS 10.15

lrhn:
Not sure who uses `.dart`, but `.dartServer` sounds like the analysis server.

probablykasper:
`.dart` seems to be for user-level settings. If `.dartServer` is for the analysis server, is there a different repo this should be reported in?

bwilkerson:
Yes, `.dartServer` is from the analysis server, and this repo is the right place for reporting the issue.

There's a larger conversation happening about unifying all of our tools and cleaning up where we're putting temporary files. Unfortunately, I don't remember who's leading that effort.

bwilkerson:
@mit-mit Do you remember where this was recently discussed?

Assigning p3 until I can figure out what the larger plan is.

mit-mit:
I think there was some discussion about this for pub, @jonasfj ?

I think we might be able to create a package with FFI for determining the canonical folders for each host OS?

jonasfj:
See: https://github.com/dart-lang/sdk/issues/41560#issuecomment-919071070

But that's really only for configuration.

There is XDG specification details for where to put cache folders. So if we want `.dartServer` to live somewhere else we should probably add a `applicationCacheHome()` to `cli_util` similar to how I added `applicationConfigHome` in https://github.com/dart-lang/cli_util/pull/66

derolf:
In your `.zshrc`, add:

``
export ANALYZER_STATE_LOCATION_OVERRIDE=$HOME/Library/Caches/dart_server
```

bartekpacia:
related https://github.com/dart-lang/sdk/issues/42813

Issue #42813, opened 2020-07-23
Any way I could change the location of `.dart` and `.dartServer` folders?


UtkarshVerma:
My home directory is getting cluttered up so I wanted to move the `.dart` and `.dartServer` folders elsewhere. Is there any way I could achieve this?

TamasBarta:
I had the same issue, because my home directory is read-only, so `~/.dartServer` could not be created there, so the Dart language server was crashing. 

I could control the `~/.dartServer` location with the `ANALYZER_STATE_LOCATION_OVERRIDE` environment variable. However, I'm not sure about `~/.dart`, or what issues I could have because of that directory never getting created.

Source: https://github.com/dart-lang/sdk/blob/28738dfad0cb557b2d5f67f6125d5c0a9ac5e3b7/pkg/analyzer/lib/file_system/physical_file_system.dart#L18-L22

Issue #42499, opened 2020-06-25
Use $HOME/.dart instead of $HOME/.dartServer


kevmoo:
The dart CLI is using `$HOME/.dart` for its analytics data.

It'd be great if the analysis server used this, too – so there are fewer places we litter a users home directory.

Not a big priority – but something to consider.

CC @jwren

kevmoo:
Also CC @mit-mit RE keeping our SDK footprint on a users machine as clean/sane as possible

jwren:
+ some analyzer folks that have some context here

mit-mit:
I don't think `$HOME/.dart` should be used for content that is "temporary"; I thought that was for settings and similar.

For temporary files, did we consider using a system-level temp folder?

mraleph:
I think we should pack everything into a single folder - and that folder should follow established conventions. 

On Linux this probably means following XDG (https://github.com/dart-lang/sdk/issues/41560)

On Mac this probably means `~/Library/Caches/...`

kevmoo:
Sorry – didn't mean temporary. I meant settings and such.

Since .dartServer IS for caches and not settings, I think Slava's idea is
solid here.

scheglov:
To do this right we would need to implement a package to support XDG directories on all platforms (Linux, Mac, Windows).
Then we could use this package in the analysis server.

jonasfj:
A utility to find an application specific config file was added to cli_util, see: [applicationConfigHome](https://pub.dev/documentation/cli_util/latest/cli_util/applicationConfigHome.html)

PR #53312, opened 2023-08-23
Adhere to XDG base directory spec

### LINKED ISSUE #41560: Adhere to XDG base directory spec
Currently, on Linux, Dart stores some data in `~/.dart` and `~/.dartServer`, which is annoying because it clutters up the user's home directory. Please consider following the [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) by moving these files to appropriate locations specified in the environment variables `$XDG_CONFIG_HOME`, `$XDG_DATA_HOME`, and `$XDG_CACHE_HOME`. (typically, and by default, `~/.config`, `~/.local/share`, and `~/.cache`).

* Dart SDK Version (`dart --version`) = 2.7.2
* Applies to Linux only

SingularisArt:
I moved the default configuration directory from `~/.dart` and `~/.dartServer` to:

1. Linux: `~/.config/dart` and `~/.config/dartServer`.
2. Mac: `~/Library/Application Support/dart/`.
3. Windows: `%APPDATA%/dart/`

You can overwrite the `~/.config/dart` path with the `DART_CONFIG_DIR` variable.
I believe this resolves #41560 for the most part.

### DIFF
--- a/pkg/analysis_server/benchmark/benchmarks.dart
+++ b/pkg/analysis_server/benchmark/benchmarks.dart
@@ -48,7 +48,7 @@ String get analysisServerSrcPath {
 }
 
 void deleteServerCache() {
-  // ~/.dartServer/.analysis-driver/
+  // ~/.config/dartServer/.analysis-driver/
   var resourceProvider = PhysicalResourceProvider.INSTANCE;
   var stateLocation = resourceProvider.getStateLocation('.analysis-driver');
   try {
--- a/pkg/analyzer/lib/file_system/physical_file_system.dart
+++ b/pkg/analyzer/lib/file_system/physical_file_system.dart
@@ -16,19 +16,6 @@ import 'package:watcher/watcher.dart';
 /// store data across sessions.
 string _serverdir;
 
-if (Platform.isLinux) {
-  var xdgConfigHome = Platform.environment['XDG_CONFIG_HOME'];
-  if (xdgConfigHome != null) {
-    _serverDir = '$xdgConfigHome/dartServer';
-  } else {
-    _serverDir = '${Platform.environment['HOME']}/.config/dartServer';
-  }
-} else if (Platform.isMacOS) {
-  _serverDir = '${Platform.environment['HOME']}/Library/Application Support/dartServer';
-} else if (Platform.isWindows) {
-  _serverDir = '${Platform.environment['APPDATA']}/dartServer';
-}
-
 /// Returns the path to default state location.
 ///
 /// Generally this is ~/.config/dartServer. It can be overridden via the
@@ -40,10 +27,18 @@ String? _getStandardStateLocation() {
     return env['ANALYZER_STATE_LOCATION_OVERRIDE'];
   }
 
-  final home = io.Platform.isWindows ? env['LOCALAPPDATA'] : env['HOME'];
-  return home != null && io.FileSystemEntity.isDirectorySync(home)
-      ? join(home, _serverDir)
-      : null;
+  if (Platform.isLinux) {
+    var xdgConfigHome = Platform.environment['XDG_CONFIG_HOME'];
+    if (xdgConfigHome != null) {
+      return '$xdgConfigHome/dartServer';
+    } else {
+      return '${Platform.environment['HOME']}/.config/dartServer';
+    }
+  } else if (Platform.isMacOS) {
+    return '${Platform.environment['HOME']}/Library/Application Support/dartServer';
+  } else if (Platform.isWindows) {
+    return '${Platform.environment['APPDATA']}/dartServer';
+  }
 }
 
 /// A `dart:io` based implementation of [ResourceProvider].
--- a/pkg/analyzer/lib/file_system/physical_file_system.dart
+++ b/pkg/analyzer/lib/file_system/physical_file_system.dart
@@ -14,11 +14,24 @@ import 'package:watcher/watcher.dart';
 
 /// The name of the directory containing plugin specific subfolders used to
 /// store data across sessions.
-const String _serverDir = ".dartServer";
+string _serverdir;
+
+if (Platform.isLinux) {
+  var xdgConfigHome = Platform.environment['XDG_CONFIG_HOME'];
+  if (xdgConfigHome != null) {
+    _serverDir = '$xdgConfigHome/dartServer';
+  } else {
+    _serverDir = '${Platform.environment['HOME']}/.config/dartServer';
+  }
+} else if (Platform.isMacOS) {
+  _serverDir = '${Platform.environment['HOME']}/Library/Application Support/dartServer';
+} else if (Platform.isWindows) {
+  _serverDir = '${Platform.environment['APPDATA']}/dartServer';
+}
 
 /// Returns the path to default state location.
 ///
-/// Generally this is ~/.dartServer. It can be overridden via the
+/// Generally this is ~/.config/dartServer. It can be overridden via the
 /// ANALYZER_STATE_LOCATION_OVERRIDE environment variable, in which case this
 /// method will return the contents of that environment variable.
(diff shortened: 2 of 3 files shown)


ykmnkmi:
Please add custom location support for both folders.

srawlins:
I love it. Also related, https://github.com/dart-lang/sdk/issues/42499 (DAS should use `~/.dart` instead of `.dartServer`), https://github.com/dart-lang/sdk/issues/42813 (customize location), and https://github.com/dart-lang/sdk/issues/49166 (use `~/Library/Application Support`, `~/Library/Caches` and `~/Library/Preferences` on mac.).

SingularisArt:
@ykmnkmi The `~/.config/dartServer` path with the `ANALYZER_STATE_LOCATION_OVERRIDE` variable.

I'll update the macOS paths.

@srawlins I'm not sure if you're suggesting that we put the ~/.config/dart and the ~/.config/dartServer into the same folder.

srawlins:
I'm not suggesting anything. Just linking issues.

SingularisArt:
Is there anything I need to change to have this PR merged?

srawlins:
Lemme review with the team. Not sure of any consequences of moving this directory between releases.

jacob314:
Ping?

mraleph:
@srawlins is this on your plate now?

srawlins:
Ah sorry, missed the ping during the conference. Will discuss this morning in a team meeting.

srawlins:
Sorry for the delay in pointing this out, but as copybara-service [points out above](https://github.com/apps/copybara-service), the code review is done at https://dart-review.googlesource.com/c/sdk/+/322300, and there are some unresolved comments there. Please look at those and upload a new commit if you make changes in response to the comments.

srawlins:
XDG actually specifies a few override environment variables, which might be interesting or important here:

> There is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable `$XDG_CACHE_HOME`.

This change could be an opportunity to use `$XDG_CACHE_HOME`, which might allow the OS to clean the cache occasionally. Today, with DAS's cache in `$HOME/.dartServer`, we can store GBs of cache in there, and the OS does not clean it up.

I think not necessary for this PR, but it would be good to address this while it's paged into our brains.

richchurcher:
Yeah, agree with @srawlins. Most people use XDG to keep clutter out of their `$HOME` and have a predictable location to find various file types. Which means:
- config lives in `$XDG_CONFIG_HOME` (usually `~/.config`)
- cache lives in `$XDG_CACHE_HOME` (usually `~/.cache`)
- data lives in `XDG_DATA_HOME` (usually `~/.local/share`)
- executables live in `$XDG_BIN_HOME` (usually `~/.local/bin`)
- libraries live in `$XDG_LIB_HOME` (usually `~/.local/lib`)

So for example, just now when I installed `patrol`, I'd normally expect to see it show up in `~/.local/bin`, but instead it's in `~/.pub-cache/bin` which is a bit odd as it's not a cache file.

richchurcher:
If it's helpful (and mostly for selfish reasons, so I get to clean up my `$HOME` :sunglasses:) I could submit a PR that more completely follows the spec, after this one is merged?

srawlins:
Actually, sorry to say this, but I think in order to minimize any intermediate locations, we should switch from `$XDG_CONFIG_HOME` to `$XDG_CACHE_HOME` before landing this; we offer nightly (ish?) builds on the Flutter master channel, so any commit could be published to users.

We chatted internally, and we'd like to provide a mechanism that slowly cleans up the data in the old location, using an existing mechanism: there is code that slowly removes old cache entries, and we can use that to remove the data in `~/.dartServer` after users switch to `$XDG_CACHE_HOME`. I like this mechanism, but wouldn't want to apply it to more and more directories as we shift where the data is stored.

CC @DanTup @scheglov would you say _everything_ in `~/.dartServer` is "cache" information, and none of it is "configuration" data? I think so, but I'm not sure. And it looks like the analytics data is moved (in this PR) from `~/.dart/` to `$DART_CONFIG_DIR ?? $XDG_CONFIG_HOME/dart/` (basically), so that seems good.

I think all I'm asking for then is to change pkg/analyzer/lib/file_system/physical_file_system.dart to use `$XDG_CACHE_HOME` instead of `$XDG_CONFIG_HOME`.

richchurcher:
If it's analytics data, I'd argue it belongs in `$XDG_DATA_HOME/dart` TBH! `~/.config/dart` would be reserved for things like configuration files specific to Dart itself which, I guess there aren't any? :slightly_smiling_face: And cache data is things that are actually cache files i.e. semi-temporary, won't be a disaster if they get deleted etc.

srawlins:
> I'd argue it belongs in $XDG_DATA_HOME/dart TBH!

That sounds fair.

> And cache data is things that are actually cache files i.e. semi-temporary, won't be a disaster if they get deleted etc.

True. But we want to be careful not to blow away these directories with abandon. For example, one design might be to delete the `~/.dartServer` directory any time we see it on server start up. But if the user has another Dart SDK in play, an older Flutter or an in-progress migration, and an IDE launching an older Dart Analysis Server, then there could be performance or stability issues. Similarly, if we stored server cache data in one directory, and then a week later we stored it in another directory, as commits land, we'd have to track those intermediate locations for the gradual data removal.

DanTup:
> CC @DanTup @scheglov would you say everything in ~/.dartServer is "cache" information, and none of it is "configuration" data? 

Here's what I have in this folder (`C:\Users\danny\AppData\Local\.dartServer` on Windows) and _my_ understanding of them (which may or may not be completely correct!)

- .analysis-driver - **cache** - I believe this is all cache that could be thrown away (and I think is periodically cleaned up by `_cacheCleanUpFunction` in `analyzer\lib\src\dart\analysis\file_byte_store.dart`)
- .dcm-teams-170 - **cache** - I think this is the same cache for DCM.. I don't know if the path comes from the analyzer or DCM itself (or whether it gets the same cleanup - @incendial?)
- .instrumentation - **???** this contains a single `uuid` file - unsure what this is used for
- .plugin_manager - **cache** - I think these are cached versions of plugins that could also be thrown away (and will rebuild)?
- .prompts - **not cache** these are user preferences (such as saying "don't show me again" to the prompt about being able to run "dart fix"
- .pub-package-details-cache - **cache** - this is a list of pub packages we fetch from Pub used in code-completion in `pubspec.yaml`
- one - **test** - generated by server integration tests
- two - **test** - generated by server integration tests

There's also some Dart-related stuff in (note: Roaming, not Local):

- `C:\Users\danny\AppData\Roaming\.dart`
- `C:\Users\danny\AppData\Roaming\.dart-tool`

These both seem telemetry related, but I'm not sure who is writing them (I suspect the new unified analytics package is doing at least one of them). I'm not sure how these two different Windows paths (Local AppData vs RoamingAppData) map on to Mac/Linux (I suspect they're merging into the home dir?).

incendial:
> .dcm-teams-170 - cache - I think this is the same cache for DCM.. I don't know if the path comes from the analyzer or DCM itself (or whether it gets the same cleanup - @incendial?)

Yes, it's just cache and we use the analyzer's API to get the location. And yes, the same bytestore is used so `_cacheCleanUpFunction` gets called as well.

bwilkerson:
@eliasyishak Could you look at the https://github.com/dart-lang/sdk/pull/53312#issuecomment-1780829314 for the telemetry question?

mraleph:
@srawlins friendly ping.

srawlins:
Thanks for the ping, @mraleph . Closing.

OP has not reponded since I asked [on Oct 19](https://github.com/dart-lang/sdk/pull/53312#issuecomment-1771332108) to respond to code review comments.

 @SingularisArt feel free to re-open.

PR #53534, opened 2023-09-14
Update double.dart

Jonadabe:
- 80 columns comments fix.

---
This is my first contribution. ;-)

<details>

</details>


sdk/lib/core/double.dart
```diff
@@ -123,13 +123,13 @@ abstract final class double extends num {
   /// Rounds away from zero when there is no closest integer:
   ///  `(3.5).roundToDouble() == 4` and `(-3.5).roundToDouble() == -4`.
   ///
-  /// If this is already an integer valued double, including `-0.0`, or it is not
-  /// a finite value, the value is returned unmodified.
+  /// If this is already an integer valued double, including `-0.0`, or it is
+  /// not a finite value, the value is returned unmodified.
```
lrhn, line 127: This is the only line which is too long, the rest do not need to be reflowed.
And I don't think most of them *should* be.

sdk/lib/core/double.dart
```diff
@@ -6,21 +6,21 @@ part of dart.core;
...
 ///
 /// It is a compile-time error for a class to attempt to extend or implement
 /// double.
 ///
 /// **See also:**
 /// * [num] the super class for [double].
-/// * [Numbers](https://dart.dev/guides/language/numbers) in
-/// [A tour of the Dart language](https://dart.dev/guides/language/language-tour).
+/// * [Numbers](https://dart.dev/guides/language/numbers) in [A tour of the Dart
+///   language](https://dart.dev/guides/language/language-tour).
```
lrhn, line 23: (TIL: You can break a markdown link across lines.)

sdk/lib/core/double.dart
```diff
@@ -123,13 +123,13 @@ abstract final class double extends num {
...
+  /// not a finite value, the value is returned unmodified.
   ///
-  /// For the purpose of rounding, `-0.0` is considered to be below `0.0`,
-  /// and `-0.0` is therefore considered closer to negative numbers than `0.0`.
-  /// This means that for a value `d` in the range `-0.5 < d < 0.0`,
-  /// the result is `-0.0`.
+  /// For the purpose of rounding, `-0.0` is considered to be below `0.0`, and
+  /// `-0.0` is therefore considered closer to negative numbers than `0.0`. This
+  /// means that for a value `d` in the range `-0.5 < d < 0.0`, the result is
+  /// `-0.0`.
```
lrhn, line 132: This doc is formatted with semantic line-breaks, where a new line starts where the sentence has a logical or semantic break, rather than in the middle of something.
That has the positive effect that if you need to change something, you're likely to only change the related lines, and won't have to touch unrelated lines at all.

That is, this doc was deliberately broken at the existing points, and reflowing it is not an improvement. Maximum words per line is not a goal. When dartdoc is displayed as HTML, it will reflow, but when read directly in the source, it's easier to read when the lines start at 
"and", "This" and "the", than at later words in the sentence.

(Not all docs are formatted like that, some predate me even knowing that documentation strategy, but some are and shouldn't be reflowed.)

lrhn:
Thanks for noticing the overlong line, and fixing it. This PR is trying to do a little *too* much, but that particular long line was too long, and reflowing it was the correct choice.

mraleph:
@Jonadabe would you like to follow up or should we close this PR?

PR #53754, opened 2023-10-13
 Correct import statements in code documentation

KagemaNjoroge:
# fix: Correct import statements in code documentation

The import statements in the code documentation were incorrect. This commit corrects them to use the standard 'dart:' prefix for importing Dart libraries.

- Changed `import:convert` to `import 'dart:convert'`
- Changed `import:io` to `import 'dart:io'`

This change ensures that the code documentation accurately reflects the correct import syntax in Dart.

### DIFF
--- a/sdk/lib/isolate/isolate.dart
+++ b/sdk/lib/isolate/isolate.dart
@@ -207,7 +207,9 @@ final class Isolate {
   /// (see http://dartbug.com/36983) or, if the state includes objects that
   /// can't be spent between isolates, a runtime failure.
   ///
-  /// ```dart import:convert import:io
+  /// ```dart
+  /// import 'dart:convert';
+  /// import 'dart:io';
   ///
   /// void serializeAndWrite(File f, Object o) async {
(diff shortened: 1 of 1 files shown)


mraleph:
These import statements are correct. They are special syntax understood by tooling which can type check the code in doc comments.

sdk/lib/isolate/isolate.dart
```diff
@@ -207,7 +207,9 @@ final class Isolate {
   /// (see http://dartbug.com/36983) or, if the state includes objects that
   /// can't be spent between isolates, a runtime failure.
   ///
-  /// ```dart import:convert import:io
```
lrhn, line 210: This style is deliberate. 

The import declarations are not intended to be part of the example code that is visible in the generated DartDoc.

The 
````
```dart import:convert import:io
````
are not intended as imports, but as instructions to our own "code sample validator" tool, which analyzes the sample in a context where those two imports are assumed to have been performed.

PR #53770, opened 2023-10-17
String.fromEnvironment: suggest `Platform.environment` for runtime

gmpassos:
- [x ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


lrhn:
Thanks for the PR.
It's a good idea, but I think it needs something more to properly distinguish the compilation environment from the process environment. The entire documentation could use that. So I went and updated the docs for all the `fromEnvironment` constructors, including adding a reference to `Platform.environment`.

Which means that is should subsume this PR.
Still, thanks for pointing out that the update was needed.

PR #53796, opened 2023-10-18
[dart2js] Fix compatiblity with Node.js 21

### LINKED ISSUE #53784: Node v21 - issue with `navigator` object
Runtime goes kaboom 💥 in Node v21 with following error: 
> `TypeError: Cannot read properties of undefined (reading 'indexOf')`

All the details can be found -> [here](https://github.com/sass/dart-sass/issues/2122)

ntkme:
This PR fixes #53784.


sigmundch:
Thanks so much for the contribution, added @rakudrama on the gerrit CL for review.

ntkme:
There are a few other places in this repo where generated codes need to be updated. I'm not sure what's the right way to get each of those updated. Please advice.

rakudrama:
> There are a few other places in this repo where generated codes need to be updated. I'm not sure what's the right way to get each of those updated. Please advice.

Can you explain where these places are, e.g. an example of some generated code, or a location in the repo?

ntkme:
@rakudrama `pkg/dds/test/web/sse_smoke_driver.dart.js` contains the generated code.

ntkme:
@rakudrama The latest commit should address your concerns in review. Other generated codes need updates are all tests, so they can probably be updated later as long as the sdk repo is not being tested with node 21.

As for how this changed tested, I'm tesing it with dart-sass, which is what most of the users are complaining about. I first ran dart-sass' test with node 21 with official dart-sdk 3.1.4 and confirm it fails. Then I build a custom 3.1.4 with this patch applied on top, and ran dart-sass' test again, everything seems to be fine.

ntkme:
Looks like tests in `pkg/compiler/test/dump_info` failed due to outdated generated code, but I'm not sure how these are supposed to be updated.

ntkme:
@rakudrama I update the dump_info with newly built sdk and now the failed test passes for me locally. Would you please retry it on CI?

ntkme:
@rakudrama Sorry that I'm very not familiar with gerrit and its flow. It looks like the tests have all passed but gerrit dismissed a previous vote because I merged latest main into this feature branch after the vote. I'm not sure what action is needed from here. Please let me know if anything is needed.

rakudrama:
I think it looks good.
Let me run the tests again, and commit it from Gerrit if everything looks ok.
Don't push _anything_ unless I say a fix is needed. It is kind of annoying how changes halt the tests and cancel votes.

Thanks for your help and patience!

PR #53908, opened 2023-10-30
Add missing #include <cstdlib>

slackito:
A recent patch to libc++ (https://github.com/llvm/llvm-project/commit/a65070a76a7c823c28f1c95a21e4857ed6e123ef) removed some transitive includes, which causes build failures when using libc++ at HEAD. In particular, this file uses `std::free` without including `<cstdlib>`:

```
error: missing '#include <cstdlib>'; 'free' must be declared before it is used
  592 |   typedef std::unique_ptr<char, decltype(std::free)*> CStringUniquePtr;
```

Relevant issue: b/308435622

### DIFF
--- a/runtime/platform/utils.h
+++ b/runtime/platform/utils.h
@@ -5,6 +5,7 @@
 #ifndef RUNTIME_PLATFORM_UTILS_H_
 #define RUNTIME_PLATFORM_UTILS_H_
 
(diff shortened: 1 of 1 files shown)


srawlins:
cc @rmacnak-google @aam @a-siva

PR #53953, opened 2023-11-03
Add Mach-O support to analyze_snapshot.

### LINKED ISSUE #53938: Add Mach-O support for analyze_snapshot
analyze_snapshot is intentionally disabled for non-elf platforms:
https://github.com/dart-lang/sdk/blob/c2ef69dff5f341aba89a110a43508cc2698cc724/runtime/runtime_args.gni#L91
https://github.com/dart-lang/sdk/blob/c2ef69dff5f341aba89a110a43508cc2698cc724/runtime/bin/analyze_snapshot.cc#L12

I have an in-progress patch to fix this.  Mostly it's moving analyze_snapshot to use AppSnapshot::TryReadAppSnapshot from bin/snapshot_utils.cc and then teaching that how to read from Mach-O.  And of course removing all the compile/GN guards.

Dart actually already has (at least?) 3 partial implementations of Mach-O parsing:
https://github.com/dart-lang/sdk/blob/c2ef69dff5f341aba89a110a43508cc2698cc724/runtime/platform/mach_o.h
https://github.com/dart-lang/sdk/blob/c2ef69dff5f341aba89a110a43508cc2698cc724/pkg/dart2native/lib/macho.dart
https://github.com/dart-lang/sdk/blob/c2ef69dff5f341aba89a110a43508cc2698cc724/pkg/native_stack_traces/lib/src/macho.dart

None of which have support for Mach-O fat files (which iOS and modern Mac systems use).

But I think what's needed here is a more-complete C++ implementation, akin to ElfLoader.  (Probably both ElfLoader and MachOLoader should get moved out of /bin and into /runtime, but that can be a separate issue.)

Mostly just filing this for commentary, and to later have a place to hang any patches I send upstream off of.  If someone else already knows of a better way to solve this, happy to use it.

FYI @perks

eseidel:
Fixes https://github.com/dart-lang/sdk/issues/53938

This also removes the build guards around analyze_snapshot since it now should support all AOT snapshots on all platforms.

I don't know how to run the tests for this change.

FYI @perks.  I don't really plan to spend a bunch of time trying to land this, but I wrote this for Shorebird, and it seems to work (on our branch), figured you might want it upstream as well.

---

### DIFF
--- a/runtime/bin/BUILD.gn
+++ b/runtime/bin/BUILD.gn
@@ -913,38 +913,35 @@ dart_executable("dart_precompiled_runtime_product") {
   extra_deps += [ ":elf_loader_product" ]
 }
 
-# This flag is set in runtime/runtime_args.gni
-# The analyze_snapshot tool is only supported on 64 bit AOT builds running under
-# linux and android platforms
-if (build_analyze_snapshot) {
-  dart_executable("analyze_snapshot") {
-    use_product_mode = dart_runtime_mode == "release"
-    extra_configs = [ "..:dart_precompiled_runtime_config" ]
+dart_executable("analyze_snapshot") {
+  use_product_mode = dart_runtime_mode == "release"
+  extra_configs = [ "..:dart_precompiled_runtime_config" ]
 
-    if (use_product_mode) {
-      extra_deps = [
-        "..:libdart_precompiled_runtime_product",
-        "../platform:libdart_platform_precompiled_runtime_product",
-      ]
-    } else {
-      extra_deps = [
-        "..:libdart_precompiled_runtime",
-        "../platform:libdart_platform_precompiled_runtime",
-      ]
-    }
-
-    extra_sources = [
-      "analyze_snapshot.cc",
-      "builtin.cc",
-      "loader.cc",
-      "loader.h",
+  if (use_product_mode) {
+    extra_deps = [
+      "..:libdart_precompiled_runtime_product",
+      "../platform:libdart_platform_precompiled_runtime_product",
     ]
+  } else {
+    extra_deps = [
+      "..:libdart_precompiled_runtime",
+      "../platform:libdart_platform_precompiled_runtime",
+    ]
+  }
 
-    if (use_product_mode) {
-      extra_deps += [ ":elf_loader_product" ]
-    } else {
-      extra_deps += [ ":elf_loader" ]
-    }
+  extra_sources = [
+    "analyze_snapshot.cc",
+    "mach_o_loader.cc",
+    "mach_o_loader.h",
+    "builtin.cc",
+    "loader.cc",
+    "loader.h",
+  ]
+
+  if (use_product_mode) {
+    extra_deps += [ ":elf_loader_product" ]
+  } else {
+    extra_deps += [ ":elf_loader" ]
   }
 }
 
--- a/runtime/bin/analyze_snapshot.cc
+++ b/runtime/bin/analyze_snapshot.cc
@@ -163,12 +163,15 @@ int RunAnalyzer(int argc, char** argv) {
 
   AppSnapshot* app_snapshot = Snapshot::TryReadAppSnapshot(script_name);
   if (app_snapshot == nullptr) {
-    Syslog::PrintErr("Failure reading snapshot\n");
+    if (File::Exists(/*namespc=*/nullptr, script_name)) {
+      Syslog::PrintErr("Failure reading snapshot\n");
+    } else {
+      Syslog::PrintErr("Snapshot file does not exist\n");
+    }
     return kErrorExitCode;
   }
-  app_snapshot->SetBuffers(
-        &vm_snapshot_data, &vm_snapshot_instructions,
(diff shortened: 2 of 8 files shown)


runtime/bin/analyze_snapshot.cc
```diff
@@ -169,24 +161,14 @@ int RunAnalyzer(int argc, char** argv) {
   const char* script_name = nullptr;
   script_name = inputs.GetArgument(0);
 
-  // Dart_LoadELF will crash on nonexistent file non-gracefully
-  // even though it should return `nullptr`.
-  File* const file = File::Open(/*namespc=*/nullptr, script_name, File::kRead);
```
eseidel, line 174: Actually probably want to add this back.  It doesn't *crash* anymore, but it does give a nonsense error message about snapshot being invalid, instead of saying the file doesn't exist.

runtime/bin/analyze_snapshot.cc
```diff
@@ -2,25 +2,17 @@
...
 
-#include "bin/elf_loader.h"
 #include "bin/error_exit.h"
 #include "bin/file.h"
 
 #include "bin/options.h"
 #include "bin/platform.h"
+#include "bin/snapshot_utils.h"
 
-#if defined(TARGET_ARCH_IS_64_BIT) && defined(DART_PRECOMPILED_RUNTIME) &&     \
```
eseidel, line 12: Maybe this is still wanted for DART_PRECOMPILED_RUNTIME?

runtime/bin/mach_o_loader.cc
```diff
@@ -0,0 +1,466 @@
...
+#include "vm/compiler/runtime_api.h"
+
+namespace dart {
+namespace bin {
+
+namespace mach_o {
+
+// This mappable abstraction matches what ElfLoader uses, but probably
+// could be removed since we don't need to support loading from memory.
+class Mappable {
```
eseidel, line 31: I tried to model this off of what elf_loader does for easiest maintenance.  I did not try to share code between them, but someone certainly could do that in a follow-up pass.  Both of them likely need to get moved into the VM instead of bin/

eseidel:
FYI @mraleph.  Happy to also just close this.  Unfortunately I don't think I'm in a position to chaperone this through the landing process in Dart, so please feel free to reject.  I just wrote this this week and after a bit more testing we're likely to ship it as part of our Shorebird solution for linking two snapshots on iOS.

runtime/bin/BUILD.gn
```diff
@@ -913,38 +913,35 @@ dart_executable("dart_precompiled_runtime_product") {
   extra_deps += [ ":elf_loader_product" ]
 }
 
-# This flag is set in runtime/runtime_args.gni
-# The analyze_snapshot tool is only supported on 64 bit AOT builds running under
-# linux and android platforms
-if (build_analyze_snapshot) {
```
eseidel, line 919: It might be easier to land this change w/o removing these guards (or just adding MacOS to the list).  I just removed them in our codebase as it seemed to work on all the targets we build, or has so far.

runtime/bin/snapshot_utils.cc
```diff
@@ -238,9 +239,60 @@ static AppSnapshot* TryReadAppSnapshotElf(
...
+  const char* error = nullptr;
+  const uint8_t *vm_data_buffer = nullptr, *vm_instructions_buffer = nullptr,
+                *isolate_data_buffer = nullptr,
+                *isolate_instructions_buffer = nullptr;
+  Dart_LoadedMachO* handle = nullptr;
+  handle = Dart_LoadMachO(script_name, &error, &vm_data_buffer,
+                          &vm_instructions_buffer, &isolate_data_buffer,
+                          &isolate_instructions_buffer);
+  if (handle == nullptr) {
+    Syslog::PrintErr("Loading failed: %s\n", error);
```
eseidel, line 288: It's possible this should be removed on landing. I I don't think the elf pathway logs here and this likely causes a spurious log during the "fall back to elf" pathway.

runtime/platform/mach_o.h
```diff
@@ -44,6 +68,73 @@ struct mach_header_64 {
...
+  uint32_t flags;
+};
+
+static constexpr uint32_t VM_PROT_NONE = 0x00;
+static constexpr uint32_t VM_PROT_READ = 0x01;
+static constexpr uint32_t VM_PROT_WRITE = 0x02;
+static constexpr uint32_t VM_PROT_EXECUTE = 0x04;
+
+// Segments may contain zero or more sections.
+struct section_64 {
```
eseidel, line 123: I didn't actually end up needing segment parsing, so this probably can be removed too.

mraleph:
I have spent time considering this - but decided to err on the side of keeping maintenance burden outside of the SDK's code base, because we don't really need this sort of tool at the moment.

eseidel:
Thanks!  We’ve made further improvements to analyze_snapshot since and
happy to offer the larger bundle of improvements should they ever be
useful. Thanks for looking!

Issue #50776, opened 2022-12-19
Some minor bugs in the Antlr grammar.


modulovalue:
I think these are some minor bugs in the Antlr grammar:

---

https://github.com/dart-lang/sdk/blob/17ccbecc6173f0591f33b5b510b45548a7dff57e/tools/spec_parser/Dart.g#L788-L789
* `'>' '>' '>' '='` should be a single literal i.e. `'>>>='` e.g. because
```dart
void main() {
  a > > > = 2;
}
```
Is not valid Dart code.
* `'>' '>' '='` should be a single literal i.e. `'>>='` e.g. because
```dart
void main() {
  a > > = 2;
}
```
is not valid Dart code.

---
https://github.com/dart-lang/sdk/blob/17ccbecc6173f0591f33b5b510b45548a7dff57e/tools/spec_parser/Dart.g#L864-L865
* `'>' '>' '>'` should be a single literal i.e. `'>>>'` e.g. because
```dart
void main() {
  2 > > > 2;
}
``` 

is not valid Dart code.
* `'>' '>'` should be a single literal i.e. `'>>'` e.g. because
```dart
void main() {
  2 > > 2;
}
```
is not valid Dart code.

---

https://github.com/dart-lang/sdk/blob/17ccbecc6173f0591f33b5b510b45548a7dff57e/tools/spec_parser/Dart.g#L830

* `'>' '='` should be a single literal i.e. `'>='` e.g. because
```dart
void main() {
  #> =;
}
```
is not valid Dart code.
---

https://github.com/dart-lang/sdk/blob/17ccbecc6173f0591f33b5b510b45548a7dff57e/tools/spec_parser/Dart.g#L441-L442
* `'[' ']'` should be a single literal i.e. `'[]'` e.g. because
```dart
void main() {
  #[ ];
}
```
is not valid Dart code.

* `'[' ']' '='` should be a single literal i.e. `'[]='` e.g. because
```dart
void main() {
  #[ ] =;
}
```

is not valid Dart code.

---

https://github.com/dart-lang/sdk/blob/17ccbecc6173f0591f33b5b510b45548a7dff57e/tools/spec_parser/Dart.g#L1354
The topLevelDefinition in the partDeclaration should allow leading metadata i.e.

```
partDeclaration
    :    partHeader (metadata topLevelDefinition)* EOF
    ;
``` 

e.g. because

```dart
part of lib;
@override
int? i;
```

Is valid Dart code.

[...]

PR #54225, opened 2023-12-04
Update Dart.g

modulovalue:
This change introduces actions that, until implemented, serve as documentation for all the locations where whitespace is considered significant. 

See https://github.com/dart-lang/sdk/issues/50776 for context around significant operator whitespace.

See https://github.com/dart-lang/language/blob/main/accepted/3.0/records/feature-specification.md#ambiguity-with-metadata-annotations for context around significant metadata whitespace. 

It seems to me that there are two ambiguities related to metadata and records.

Consider:

`@Foo (a, b) foo() {}`

It can be a constructorDesignation followed by arguments and an untyped function 

or an identifier followed by a function with a record return type.

We disambiguate the case where whitespace is present in favor of the record by requiring the arguments to be preceded by no whitespace. So this is an identifier followed by a function with a record return type.

Consider:

`@Foo(a, b) foo() {}`

It can be a constructorDesignation followed by arguments and an untyped function 

or an identifier followed by a function with a record return type.

We disambiguate the case where whitespace is not present by requiring the other metadata productions to have trailing whitespace. So this is a constructorDesignation followed by arguments and an untyped function.

### DIFF
--- a/tools/spec_parser/Dart.g
+++ b/tools/spec_parser/Dart.g
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 // CHANGES:
+// v0.38 Highlight locations where whitespace is significant.
 //
 // v0.37 Correct `libraryExport` to use `configurableUri`, not `uri`.
 //
@@ -515,8 +516,8 @@ operatorSignature
 operator
     :    '~'
     |    binaryOperator
-    |    '[' ']'
-    |    '[' ']' '='
+    |    '[' NO_SKIP ']'
+    |    '[' NO_SKIP ']' NO_SKIP '='
     ;
 
 binaryOperator
@@ -616,9 +617,9 @@ metadata
     ;
 
 metadatum
-    :    constructorDesignation arguments
-    |    identifier
-    |    qualifiedName
+    :    constructorDesignation NO_SKIP arguments
+    |    identifier YES_SKIP
+    |    qualifiedName YES_SKIP
     ;
 
 expression
@@ -856,8 +857,8 @@ compoundAssignmentOperator
     |    '+='
     |    '-='
     |    '<<='
-    |    '>' '>' '>' '='
-    |    '>' '>' '='
+    |    '>' NO_SKIP '>' NO_SKIP '>' NO_SKIP '='
+    |    '>' NO_SKIP '>' NO_SKIP '='
     |    '&='
     |    '^='
     |    '|='
@@ -898,7 +899,7 @@ relationalExpression
     ;
 
 relationalOperator
-    :    '>' '='
+    :    '>' NO_SKIP '='
     |    '>'
     |    '<='
     |    '<'
@@ -932,8 +933,8 @@ shiftExpression
 
 shiftOperator
     :    '<<'
-    |    '>' '>' '>'
-    |    '>' '>'
+    |    '>' NO_SKIP '>' NO_SKIP '>'
+    |    '>' NO_SKIP '>'
     ;
 
 additiveExpression
@@ -2186,14 +2187,24 @@ IDENTIFIER
     :    IDENTIFIER_START IDENTIFIER_PART*
     ;
 
+SKIPPABLE
+    :    (SINGLE_LINE_COMMENT | MULTI_LINE_COMMENT | WS) { skip(); }
+    ;
+
+NO_SKIP
+    :    { /* TODO: accept if first(SKIPPABLE) can't be recognized */ }
+    ;
+
+YES_SKIP
+    :    { /* TODO: accept if first(SKIPPABLE) can be recognized */ }
+    ;
+
 SINGLE_LINE_COMMENT
     :    '//' (~('\r' | '\n'))* NEWLINE?
-         { skip(); }
     ;
 
 MULTI_LINE_COMMENT
     :    '/*' (MULTI_LINE_COMMENT | .)*? '*/'
-         { skip(); }
     ;
 
 fragment
@@ -2207,5 +2218,4 @@ FEFF
 
 WS
     :    (' ' | '\t' | '\r' | '\n')+
-         { skip(); }
     ;
--- a/tools/spec_parser/Dart.g
+++ b/tools/spec_parser/Dart.g
@@ -2192,11 +2192,11 @@ SKIPPABLE
     ;
 
 NO_SKIP
-    :    { /* TODO: accept if first(SKIPPABLE) can't be recognized */ }
+    :    { true; /* TODO: accept if first(SKIPPABLE) can't be recognized */ }
     ;
 
 YES_SKIP
-    :    { /* TODO: accept if first(SKIPPABLE) can be recognized */ }
+    :    { true; /* TODO: accept if first(SKIPPABLE) can be recognized */ }
     ;
 
 SINGLE_LINE_COMMENT


eernstg:
Hello @modulovalue, I can see how this might work, but I'm not quite sure why it would be an improvement. Could you expand a bit on that?

The main perspective I have on this issue is that the syntactic structure requires us to consider each character in an operator like `>>>=` as a separate token, and we can then parse both expressions like `x >>>= 2` and expressions like `List<List<List<int>>>==List<List<List<int>>>` without any special exceptions. In that sense, `>>>=` _is_ four tokens, they just happen to be written with no whitespace in common usage.

We might then perform an extra check on the whitespace (outside the grammar, this is a completely ad-hoc extra check) reporting a compile-time error if we _do_ encounter whitespace, like `>>  /* Look, a comment! */  >=`. But that's no different from a rule that says "no comments just before semicolon" or "we must have exactly 2 empty lines between each top-level declaration and the next one", or whatever we might come up with. The grammar doesn't care, but we might like it for aesthetic reasons.

It could be argued that the grammar would be more consistent if we were to tokenize _all_ punctuation characters separately. However, that's again a choice, and so far we've chosen to use separate characters at the token level in the situations where it is necessary (because we'll otherwise get tokenization that is plain wrong), and then we haven't bothered to do it in all the other cases.

modulovalue:
I appear to have made some wrong assumptions about the analyzer.

I've produced the following table to clear things up:

```dart
void foo() {
  //                                    ANTLR               Analyzer            My Assumption       Clash
  // Index operator
  #[]; //                               Success             Success             Success             -
  #[ ]; //                              Success             Error               Error               -
  // Index assignment operator
  #[]=; //                              Success             Success             Success             -
  #[ ]=; //                             Success             Error               Error               -
  #[] =; //                             Success             Error               Error               -
  #[ ] =; //                            Success             Error               Error               -
  // GGGEq
  a >>>= b; //                          Success             Success             Success             -
  a > >>= b; //                         Success             Error               Error               -
  a >> >= b; //                         Success             Error               Error               -
  a >>> = b; //                         Success             Error               Error               -
  a > > >= b; //                        Success             Error               Error               -
  a > >> = b; //                        Success             Error               Error               -
  a >> > = b; //                        Success             Error               Error               -
  a > > > = b; //                       Success             Error               Error               -
  List<List<List<foo>>>==bar; //        Success             Error               Success             ↯
  List<List<List<foo> >>==bar; //       Success             Error               Success             ↯
  List<List<List<foo>> >==bar; //       Success             Error               Success             ↯
  List<List<List<foo>>> ==bar; //       Success             Success             Success             -
  List<List<List<foo> > >==bar; //      Success             Error               Success             ↯
  List<List<List<foo> >> ==bar; //      Success             Success             Success             -
  List<List<List<foo>> > ==bar; //      Success             Success             Success             -
  List<List<List<foo> > > ==bar; //     Success             Success             Success             -
  // GGEq
  a >>= b; //                           Success             Success             Success             -
  a > >= b; //                          Success             Error               Error               -
  a >> = b; //                          Success             Error               Error               -
  a > > = b; //                         Success             Error               Error               -
  List<List<foo>>==bar; //              Success             Error               Success             ↯
  List<List<foo> >==bar; //             Success             Error               Success             ↯
  List<List<foo>> ==bar; //             Success             Success             Success             -
  List<List<foo> > ==bar; //            Success             Success             Success             -
  // GEq
  a >= b; //                            Success             Success             Success             -
  a > = b; //                           Success             Error               Error               -
  List<foo>==bar; //                    Success             Error               Success             ↯
  List<foo> ==bar; //                   Success             Success             Success             -
  // GGG
  a >>> b; //                           Success             Success             Success             -
  a > >> b; //                          Success             Error               Error               -
  a >> > b; //                          Success             Error               Error               -
  a > > > b; //                         Success             Error               Error               -
  List<List<List<foo>>>; //             Success             Success             Success             -
  List<List<List<foo> >>; //            Success             Success             Success             -
  List<List<List<foo>> >; //            Success             Success             Success             -
  List<List<List<foo> > >; //           Success             Success             Success             -
  // GG
  a >> b; //                            Success             Success             Success             -
  a > > b; //                           Success             Error               Error               -
  List<List<foo>>; //                   Success             Success             Success             -
  List<List<foo> >; //                  Success             Success             Success             -
}

```


@eernst so the "Analyzer" column matches the spec, and the "My Assumption" column does not, correct?

eernstg:
You can't really say that a particular expression would be parsed according to the language specification in a specific way, because it contains tokens like `>>>=`, and the existence of such tokens would preclude a successful parse of any of the List-List-List examples where there's to little white space between the characters of `>>>=` to disable tokenization into something that includes `>>`, `>>>`, `>>>=`, `>>=`, or `>=`.

So we actually have to assume that the language specification uses the same approach as Dart.g, or we have to assume that it is corrected in some other way. If we don't fix it then `List<List<List<foo>>>==bar` is derivable but not parsable, and this implies that we're not talking about a Dart parser. In other words, we have to take into account that the language specification grammar has a problem before we can even start resolving these finer details of whitespace processing. I tend to rely on the solution which is used in Dart.g (because it seems clean and well-defined to me, and surely also ever so slightly because I came up with it ;-).

modulovalue:
@eernstg what do you think about the metadata-related changes for documenting the record/metadata disambiguation strategy in the ANTLR grammar?

modulovalue:
I have implemented the noSkip approach for operators:

```dart
void foo() {
  //                                    ANTLR               Analyzer            My Assumption       Clash
  // Index operator
  #[]; //                               Success             Success             Success             -
  #[ ]; //                              Error               Error               Error               -
  // Index assignment operator
  #[]=; //                              Success             Success             Success             -
  #[ ]=; //                             Error               Error               Error               -
  #[] =; //                             Error               Error               Error               -
  #[ ] =; //                            Error               Error               Error               -
  // GGGEq
  a >>>= b; //                          Success             Success             Success             -
  a > >>= b; //                         Error               Error               Error               -
  a >> >= b; //                         Error               Error               Error               -
  a >>> = b; //                         Error               Error               Error               -
  a > > >= b; //                        Error               Error               Error               -
  a > >> = b; //                        Error               Error               Error               -
  a >> > = b; //                        Error               Error               Error               -
  a > > > = b; //                       Error               Error               Error               -
  List<List<List<foo>>>==bar; //        Success             Error               Success             ↯
  List<List<List<foo> >>==bar; //       Success             Error               Success             ↯
  List<List<List<foo>> >==bar; //       Success             Error               Success             ↯
  List<List<List<foo>>> ==bar; //       Success             Success             Success             -
  List<List<List<foo> > >==bar; //      Success             Error               Success             ↯
  List<List<List<foo> >> ==bar; //      Success             Success             Success             -
  List<List<List<foo>> > ==bar; //      Success             Success             Success             -
  List<List<List<foo> > > ==bar; //     Success             Success             Success             -
  // GGEq
  a >>= b; //                           Success             Success             Success             -
  a > >= b; //                          Error               Error               Error               -
  a >> = b; //                          Error               Error               Error               -
  a > > = b; //                         Error               Error               Error               -
  List<List<foo>>==bar; //              Success             Error               Success             ↯
  List<List<foo> >==bar; //             Success             Error               Success             ↯
  List<List<foo>> ==bar; //             Success             Success             Success             -
  List<List<foo> > ==bar; //            Success             Success             Success             -
  // GEq
  a >= b; //                            Success             Success             Success             -
  a > = b; //                           Error               Error               Error               -
  List<foo>==bar; //                    Success             Error               Success             ↯
  List<foo> ==bar; //                   Success             Success             Success             -
  // GGG
  a >>> b; //                           Success             Success             Success             -
  a > >> b; //                          Error               Error               Error               -
  a >> > b; //                          Error               Error               Error               -
  a > > > b; //                         Error               Error               Error               -
  List<List<List<foo>>>; //             Success             Success             Success             -
  List<List<List<foo> >>; //            Success             Success             Success             -
  List<List<List<foo>> >; //            Success             Success             Success             -
  List<List<List<foo> > >; //           Success             Success             Success             -
  // GG
  a >> b; //                            Success             Success             Success             -
  a > > b; //                           Error               Error               Error               -
  List<List<foo>>; //                   Success             Success             Success             -
  List<List<foo> >; //                  Success             Success             Success             -
}

```

It seems to be an improvement as it appears to only reject programs that aren't considered valid Dart programs by the analyzer.


The updated grammar with actions written in Dart can be found here: https://github.com/modulovalue/dart_antlr_dart_experiments.dart/blob/8e94c4422f2b17e5468f1ac2f66c12ad33304bb1/dart_antlr_dart/lib/parsers/antlr/main/gen/Dart.g4#L166-L168 

I haven't managed to get record/metadata disambiguation to work yet with this approach.

modulovalue:
I manage to either fix the record/metadata ambiguity or enforce the operator whitespace restriction, but not both at the same time.

Doing both at the same time doesn't seem to work because a failure to match an action is reported as a `FailedPredicateException`.

The operators need to be rejected, but a `false` action does not appear to reject them.

However, a `false` action does appear to reject the one production that we need to reject for the record/metadata ambiguity.

I suspect that this is a limitation of ANTLR and its LL-based approach, but I have an idea for a workaround. I'll try that and if I don't manage to get both to work at the same time, I'll keep the metadata/record disambiguation strategy since that one can't be implemented in a post-processing step.

vsmenon:
Is this still being iterated on or should it be closed?

eernstg:
@vsmenon wrote:
> Is this still being iterated on or should it be closed?

It is a matter of priorities. I very much appreciate the detailed and carefully worded input from @modulovalue, and I do recognize that there are several issues with the Dart grammar.

In this particular case, however, I'm not convinced that it would be an improvement on Dart to adopt the "no skip" rule suggested in this issue. I think it would be a better way ahead to use the same approach in the language specification as the one that we're using in the spec parser [Dart.g](https://github.com/dart-lang/sdk/blob/main/tools/spec_parser/Dart.g).

This means that we would parse the operator `>>>` as 3 separate tokens with the text `>`, and similarly for compound assignment operators like `>>>=`, and also several other tokens.

The point is that if we proceed and parse the text ">>>=" as the single token `>>>=`, and consider that to be a compound assignment operator, then we will reject some perfectly legitimate Dart expressions (e.g.,  `List<List<List<X>>>==Y`). In other words, we cannot unconditionally consider `>>>=` to be a token, and we can't (reasonably) express the condition at the lexical level.

The parser actually does things like tokenizing ">>" as a single token `>>`, and then breaking it up into `>` and `>` in case it turns out to be two angle brackets that occur at the end of a type argument list. We probably don't want to mimic this kind of "tokenize as one thing, then redo it when we detect that it doesn't work" tricks in a specification (including Dart.g).

It makes the syntax considerably more complex, and it relies on properties of the source code which are not expressible using a context free grammar (unless we change the grammar radically and specify whitespace in all details). The Dart grammar cannot be expressed as a context free grammar today, but every step we take to make it _more_ incompatible with context free grammars will make it harder to support access to Dart source code using other tools than our own tool chain (e.g., for IDE support, github highlighting, etc).

Also note that it doesn't create any problems if the Dart parser accepts "> >" (note the space) as a legitimate spelling of the right-shift operator. We could complain about the space, but that doesn't actually have to be part of the parser (or the language), it could be more like a style hint.

modulovalue:
> Is this still being iterated on or should it be closed?

My plan is to, at least, land the record/metadata disambiguation part of this PR. 
I have a working POC that uses actions written in Dart. I forgot to mention it, but I'm waiting on https://github.com/dart-lang/sdk/issues/54540#issuecomment-1880688005 to clear up before I continue working on this.

So far, I haven't quite managed to get ANTLR to do what's needed to implement the rest of this PR, so I'm probably going to defer that to a different PR.

modulovalue:
Because there are many more constructs that are extremely inconvenient to express with the tools provided by ANTLR (or any other open-source tool for that matter), I've decided to close this PR.

PR #54449, opened 2023-12-24
Update linked_list.dart

EmreKaratopuk:
Replace double-linked list with circular doubly linked list in description.

### DIFF
--- a/sdk/lib/collection/linked_list.dart
+++ b/sdk/lib/collection/linked_list.dart
@@ -4,7 +4,7 @@
 
 part of dart.collection;
 
-/// A specialized double-linked list of elements that extends [LinkedListEntry].
+/// A specialized circular doubly linked list of elements that extends [LinkedListEntry].
 ///
 /// This is not a generic data structure. It only accepts elements that *extend*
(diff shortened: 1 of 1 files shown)


sdk/lib/collection/linked_list.dart
```diff
@@ -4,7 +4,7 @@
 
 part of dart.collection;
 
-/// A specialized double-linked list of elements that extends [LinkedListEntry].
+/// A specialized circular doubly linked list of elements that extends [LinkedListEntry].
```
lrhn, line 7: I guess it is circular, in that the first element's previous element is the last.

The `LinkedList` itself doesn't really use that, it just happens to be a property of the links,
so I don't think I want to emphatise it.

If anything, it's more useful saying that it's a double-ended list, which allows operations at both start and end. The link from first to last element is really just an implementation detail of that.

vsmenon:
Should we close this PR?

lrhn:
Let's. It's a small change, and I disagree (a little) with the focus it adds.
It's easy enough to start from scratch if we find a better approach.

PR #54618, opened 2024-01-14
HTMLSelectElement: fix `selectedOptions` when `selectedIndex` < 0

gmpassos:
- [ OK] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

https://github.com/dart-lang/sdk/blob/79df0b278994e5210f2fd4ea4e5a7f64db0f5398/tools/dom/templates/html/impl/impl_HTMLSelectElement.darttemplate#L23

This line should check if `this.selectedIndex` returns `< 0`.

`selectedIndex` returns `-1` when the `select` has no options or there's no value selected.

See:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex

---

The code below shows the error:

```dart
import 'dart:html';

void main() {

  window.document.documentElement?.appendHtml('<select id="sel1"></select>');
  
  var sel = querySelector('#sel1') as SelectElement;
  
  print(sel);
  
  print('selectedIndex: ${sel.selectedIndex}'); // -1
  print(sel.selectedOptions); // error
}

```


mraleph:
cc @sigmundch

sigmundch:
Thanks for the PR @gmpassos - we left a couple messages for you on the code-review tool: https://dart-review.googlesource.com/c/sdk/+/346220. In case you haven't seen it, the change looks good, the only thing we'd also need is to re-generate `dart:html` by calling `(cd tools/dom/scripts; ./go.sh)`. Thanks!

gmpassos:
> Thanks for the PR @gmpassos - we left a couple messages for you on the code-review tool: https://dart-review.googlesource.com/c/sdk/+/346220. In case you haven't seen it, the change looks good, the only thing we'd also need is to re-generate `dart:html` by calling `(cd tools/dom/scripts; ./go.sh)`. Thanks!


I have some questions:

- What should I install to run this script?
- Do I need to build the SDK first?

I see it uses `vpython3`, is it the same as `python3`? How to install?

Best regards.

*NOTE: I believe this will regenerate many files, and you won't check line by line in these files. IMHO, for security reasons, this type of source code modification should be executed by a trusted/admin developer of the project. I can generate the files for you, but this is not a security protocol that the Dart SDK should rely on.*

sigmundch:
> I see it uses vpython3, is it the same as python3? How to install it?

Yes, precisely, just python-3. That said, if you run into trouble, let us know and we'd be happy to do it on your behalf.

> NOTE: I believe this will regenerate many files, and you won't check line by line in these files. IMHO, for security reasons, this type of source code modification should be executed by a trusted/admin developer of the project. I can generate the files for you, but this is not a security protocol that the Dart SDK should rely on.

Generally we try to avoid landing generated code, but in this case we do because the emitted code is the library that we provide as an API. As such, we do go through the process of reviewing it. Thankfully, the generated files are up to date on every change we make to a template input file, which means the diffs and reviews are similar to reviewing the template files. For example, in this PR I'd only expect to see a 2-line change in `html_dart2js.dart`.

PR #54631, opened 2024-01-15
No Longer Considering lineMode to check whether CR is line break

### LINKED ISSUE #54588: [dart:io] - stdin.readLineSync on Windows Powershell keeps waiting until something being typed
```dart
void main() {
    final a = stdin.readLineSync();
    print(a);
}
```
If The above program  executed on windows powershell, and if u press enter it would just move the cursor to next line and keep repeatedly do this until u type something, this not letting to pass empty input. Originally Founded from mason_logger https://github.com/felangel/mason/issues/1189

### LINKED ISSUE #54630: [dart:io] stdio handle is being invalid in git bash in Windows
While Analyzing this https://github.com/dart-lang/sdk/issues/54588, 
I came to address this same issue in deno https://github.com/denoland/deno/issues/21950

The Stdio Handle is being invalid in git bash (these cases are positive in powershell), so i guess there could be different resource id for stdio for git bash, since i guess git bash window is generated differently than powershell.

This below program would throw an error in git bash (Windows),
```dart
import "dart:io";

void main(){
        stdin.echoMode = false;
        stdin.lineMode = false;
        print(stdin.lineMode);
}

```

Error: 
```
Unhandled exception:
StdinException: Error setting terminal echo mode, OS Error: The handle is invalid.
, errno = 6
#0      Stdin.echoMode= (dart:io-patch/stdio_patch.dart:85:7)
#1      main (package:playground/lineMode.dart:4:8)
#2      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:296:19)
#3      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:189:12)
```

codesculpture:
Primarily fixes https://github.com/dart-lang/sdk/issues/54588
`lineMode` is seems unnecessary to check for crnewline, since regardless of `lineMode` the <Enter> is interpreted as `CR` in windows (Native Console Based Application such as xterm)

Also planning to fix https://github.com/dart-lang/sdk/issues/54630
Edit: Plan to fix this separately https://github.com/dart-lang/sdk/issues/54630

### DIFF
--- a/sdk/lib/io/stdio.dart
+++ b/sdk/lib/io/stdio.dart
@@ -59,9 +59,8 @@ class Stdin extends _StdStream implements Stream<List<int>> {
     const LF = 10;
     final List<int> line = <int>[];
     // On Windows, if lineMode is disabled, only CR is received.
-    bool crIsNewline = Platform.isWindows &&
-        (stdioType(stdin) == StdioType.terminal) &&
-        !lineMode;
+    bool crIsNewline =
+        Platform.isWindows && (stdioType(stdin) == StdioType.terminal);
     if (retainNewlines) {
       int byte;
       do {


a-siva:
@codesculpture I have added @brianquinlan as reviewer for the Gerrit CL https://dart-review.googlesource.com/c/sdk/+/346381

codesculpture:
Thanks @a-siva , this is not a finalized solution. Had to discuss the consequences of removing `lineMode`, need to know why it is used in `crnewline`.

brianquinlan:
Let me know when you are ready for review and thanks so much for looking at this!

codesculpture:
Its ready for review. @brianquinlan

mraleph:
I am going to close this because the underlying issue is a bug in [Windows Terminal](https://github.com/microsoft/terminal/issues/16223) which sometimes causes terminal to eat LFs. 

Thanks for trying to fix it @codesculpture!

PR #54781, opened 2024-01-31
Add DSP.

### LINKED ISSUE #54540: [ANTLR] Migrating from Java actions to Dart actions.
Consider [Dart.g](https://github.com/dart-lang/sdk/blob/master/tools/spec_parser/Dart.g).

It contains platform-dependent actions to make it easier to express certain syntactical restrictions.

The platform-dependent actions are written in Java.

There's a [Dart target](https://github.com/antlr/antlr4/pull/2728) that landed a couple of years ago, and I've been [using](https://github.com/modulovalue/dart_antlr_dart_experiments.dart) it with great success for finding inconsistencies between the spec and implementation.

I want to propose migrating the actions in Dart.g from Java to Dart.

This would simplify things by removing a dependency on Java. It would also make it much easier to plug in parsers generated from the ANTLR grammar into some existing pipeline, allowing us to ensure that the analyzer and ANTLR grammar are in sync in terms of what they implement and specify, respectively.

The main risk that I see is that the Java target is better maintained, and there's a possibility that the Dart target gets abandoned. However, there's reason to believe that the probability of that is not as high as it might appear, as [for the past three years, the Dart target has received plenty of contributions](https://pub.dev/packages/antlr4/versions) that have kept it up to date.

@eernst how strongly do you feel about having Java bindings, and what do you think about moving from Java to Dart? I can submit a PR if you have no objections.

modulovalue:
Fixes https://github.com/dart-lang/sdk/issues/54540.

Other changes:

- There's a new FAQ section that describes how to generate parsers.
- I've dedented the custom code because ANTLR adds its own indentation to generated parsers.

Also, just so that we don't forget this, it would be great to maintain the git history of `Dart.g`, i.e., if `Dart.g` is removed, it might be better to replace it with the contents of `Dart.g4` in this PR instead of just removing `Dart.g` and keeping this `Dart.g4`.

### DIFF
--- a/tools/spec_parser/dart_spec_parser/Dart.g4
+++ b/tools/spec_parser/dart_spec_parser/Dart.g4
@@ -1,0 +1,2224 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// CHANGES:
+// v0.39 Translate actions from Java to Dart.
+//
+// v0.38 Broaden `initializerExpression` to match implemented behavior.
+//
+// v0.37 Correct `libraryExport` to use `configurableUri`, not `uri`.
+//
+// v0.36 Update syntax from `inline class` to `extension type`, including
+// a special case of primary constructors.
+//
+// v0.35 Change named optional parameter syntax to require '=', that is,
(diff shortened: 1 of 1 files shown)


eernstg:
Hi @devoncarew, thanks for taking a look at this! I actually landed the CL without problems. So now we just need a newer ANTLR on all systems. ;-)

PR #54799, opened 2024-02-01
fix(bin/stdio)[win]: Retrieving Original Input Handle if its piped

codesculpture:
https://github.com/dart-lang/sdk/issues/54630 Trying To Fix this,
The Solution is to retrieve the original input handle as mentioned [here](https://github.com/dart-lang/sdk/issues/54630#issuecomment-1902236707)

But i need help implementing the solution because i dont want to introduce new complexities in the interface.

I need `stdioMode` during runtime of `SetEchoMode` and `setLineMode` where i can get this from, either i need to generate during the method invocation or maybe i can get that as argument but that would break the existing `stdin.h` interface since its only required for the windows. 

Welcome for suggestions  @mraleph @brianquinlan @a-siva

### DIFF
--- a/runtime/bin/stdio_win.cc
+++ b/runtime/bin/stdio_win.cc
@@ -5,8 +5,8 @@
 #include "platform/globals.h"
 #if defined(DART_HOST_OS_WINDOWS)
 
+#include "bin/file.h"
 #include "bin/stdio.h"
-
 // These are not always defined in the header files. See:
 // https://msdn.microsoft.com/en-us/library/windows/desktop/ms686033(v=vs.85).aspx
 #ifndef ENABLE_VIRTUAL_TERMINAL_INPUT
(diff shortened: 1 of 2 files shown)


a-siva:
//cc @aam can you review this PR

runtime/bin/stdio_win.cc
```diff
@@ -42,8 +54,8 @@ bool Stdin::GetEchoMode(intptr_t fd, bool* enabled) {
   return true;
 }
 
-bool Stdin::SetEchoMode(intptr_t fd, bool enabled) {
-  HANDLE h = GetStdHandle(STD_INPUT_HANDLE);
+bool Stdin::SetEchoMode(intptr_t fd, bool enabled, int stdioMode) {
+  HANDLE h = GetInpStdHandle(stdioMode);
```
aam, line 58: Immediate question that comes up is whether updating console mode (via CONIN$) will affect behavior of redirected stdin that dart app deals with. Basically if dart app reads from stdin which might originate from console, might originate from somewhere else, updating actual console mode might give developer false sense of successfully completed operation.

codesculpture: > The standard handles of a process may have been redirected by a call to SetStdHandle, in which case GetStdHandle will return the redirected handle. If the standard handles have been redirected, you can specify the CONIN$ value in a call to the CreateFile function to get a handle to a console's input buffer. Similarly, you can specify the CONOUT$ value to get a handle to the console's active screen buffer.

As by this from [here](https://learn.microsoft.com/en-us/windows/console/setstdhandle) i believed by CreatFule with Conin$ we would get the actual handle which is the effective handle controlling the console, but ur question eagers me to disbelief this i will test this and will let u know, thanks for reviewing.

PR #55296, opened 2024-03-25
Add enumerate method to List

aglpy:
This contribution add a new feature to List, it allow the creation of a new List using the method ".enumerate" in analogy to other languages, this method allows to use the value and the position of this value in the List to do operations and create the new List.

The use look like this:

List\<String\> sections = ['apple', 'orange', 'lemon'];

List\<String\> titles = sections.enumerate((index, section) => '${index + 1} - $section').toList();

### DIFF
--- a/sdk/lib/collection/list.dart
+++ b/sdk/lib/collection/list.dart
@@ -342,6 +342,10 @@ abstract mixin class ListBase<E> implements List<E> {
     return ListMapView<E>(this);
   }
 
+  Iterable<R> enumerate<R>(R Function(int index, E item) mapper) {
+    return asMap().entries.map((entry) => mapper(entry.key, entry.value));
+  }
+
   List<E> operator +(List<E> other) => [...this, ...other];
 
   List<E> sublist(int start, [int? end]) {


julemand101:
`List` is an interface for any kind of lists and any changes to this interface are considered a major breaking change where the change must be valued as important enough to break any existing List implementations.

This change does not do anything which we can't already do with existing functionality in Dart. E.g. your example could easily be written as:
```dart
  List<String> titles = [
    for (final (index, section) in sections.indexed) '${index + 1} - $section'
  ];
```

Or if you import the `package:collection` you can do the following which the same as you suggest just with a different name of the method:
```dart
  List<String> titles = sections
      .mapIndexed((index, section) => '${index + 1} - $section')
      .toList();
```

So even if your suggestion are more fitting `package:collection`, your change are already part of that package.

kevmoo:
@lrhn ?

lrhn:
The `indexed` extension method on `Iterable` already implements similar functionality, by exposing an `Iterable<(int, E)>`. It's available in the platform libraries.

As @julemand101 mentions, the exact functionality is also in `package:collection` under another name 

We have no plans to add that to the platform libraries.

(And as also mentioned, it will definitely not be an interface member on `List`. Adding members to a widely implemented interface is hugely breaking.)

PR #55412, opened 2024-04-09
Fix issue with invoking main function from browser 

akalankavinda:
When trying to invoke main method the following error occurs.
image

this happens because of we store the result from `WebAssembly.instantiate()`, which is a **WebAssemblyInstantiatedSource** type value. We should store the instance value inside the returned object

image

### DIFF
--- a/pkg/dart2wasm/lib/js/runtime_blob.dart
+++ b/pkg/dart2wasm/lib/js/runtime_blob.dart
@@ -149,12 +149,12 @@ const jsRuntimeBlobPart5 = r'''
         "substring": (s, a, b) => s.substring(a, b),
     };
 
-    dartInstance = await WebAssembly.instantiate(await modulePromise, {
+    const wasmInstance = await WebAssembly.instantiate(await modulePromise, {
         ...baseImports,
         ...(await importObjectPromise),
         "wasm:js-string": jsStringPolyfill,
     });
-
+    dartInstance = wasmInstance.instance;
     return dartInstance;
 }
 


osa1:
@akalankavinda you are instantiating the Wasm module wrong. You should be using the instantiation function provided by the .mjs file. Here's a full example:

```html
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta charset="utf-8">
  <meta name="dart.unittest" content="full-stack-traces">
  <title> test.dart.wasm </title>
  <link rel="preload" href="/test.wasm" as="fetch" crossorigin>
</head>
<body>
  <script type="module">
  const dartModulePromise = WebAssembly.compileStreaming(fetch('/test.wasm'));
  const imports = {};
  let dart2wasm_runtime = await import('/test.mjs');
  let moduleInstance =
      await dart2wasm_runtime.instantiate(dartModulePromise, imports);
  dart2wasm_runtime.invoke(moduleInstance);
  </script>
</body>
</html>
```

You can't instantiate the module yourself because you don't know what imports the module expects and how to provide them. Use the code above.

If you really want to instantiate it yourself, you have to use the second overloading of `WebAssembly.instantiate` here: https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/instantiate_static#secondary_overload_%E2%80%94_taking_a_module_object_instance

Which returns "A `Promise` that resolves to an `WebAssembly.Instance` object".

PR #55489, opened 2024-04-17
Add isPositive in Duration

knottx:
- Add `isPositive` in `Duration`
- Add Test for `isPositive`

---

<details>

</details>

### DIFF
--- a/sdk/lib/core/duration.dart
+++ b/sdk/lib/core/duration.dart
@@ -372,6 +372,12 @@ class Duration implements Comparable<Duration> {
   /// earlier time.
   bool get isNegative => _duration < 0;
 
+  /// Whether this [Duration] is positive.
+  ///
+  /// A positive [Duration] represents the difference from an earlier time to a
+  /// later time.
+  bool get isPositive => _duration > 0;
+
   /// Creates a new [Duration] representing the absolute length of this
   /// [Duration].
   ///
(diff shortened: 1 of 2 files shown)


viniciusddrft (approved):

lrhn:
Not a bad idea.

Only issue is that `Duration` is not a `final` class, so someone might implement it, and will then be broken when they don't implement `isPositive`. That's a bad issue, though, and the reason for not doing most possible and reasonable changes in the platform libraries.

If the class had been made `final`, it would be more likely that we could add new members. (Not immediately, since pre-3.0 code can still implement final SDK classes, but eventually!) 
The fact that the class wasn't made final, even if it is a plain data object and other such types were made final, suggests that someone is already implementing the type. (Unless it was an oversight. Those do happen.)

There is a `_FakeDuration` class in tests in two packages I can see. Those are probably both mockito generated, so wouldn't be broken by adding a method, they just need to be re-generated.

lrhn:
I think my summary would be: Nice idea, too bad it's breaking.
If it had more direct value, it might we worth going through the breaking change process, but it doesn't seem like it's woth the effort, when the workaround is just to use `!d.isNegative`.

knottx:
> I think my summary would be: Nice idea, too bad it's breaking. If it had more direct value, it might we worth going through the breaking change process, but it doesn't seem like it's woth the effort, when the workaround is just to use `!d.isNegative`.

Yes, it's more of a 'nice to have' than a 'must-have' depending on the specific use case.
Just to clarify, d.isPositive specifically checks if d > 0, while !d.isNegative includes d = 0 as well. So, while isPositive isn't essential, it does offer a clearer distinction when needed. While it may not be crucial for all cases, having it as an option could enhance readability and understanding in certain contexts. Thank you for highlighting the distinction!

Issue #55693, opened 2024-05-11
Track the `ExtensionTypeElement` associated with constant `DartObject`s.


mattrberry:
I'd like to propose a feature that would allow users of the analyzer to work back to the extension type for a DartObject.

This will give lints and code generators a little bit of visibility into how an object is constructed and its type.

The particular limitation I'm seeing now is when an extension type is defined over a private type. For a reduced and contrived example, take

```dart
class _Foo {
  const _Foo();
}
extension type const Foo._(_Foo _) implements _Foo {
  const Foo() : this._(const _Foo());
}
```

In this case, the lint or code generator might want to do something with the static type `Foo`. Today, the analyzer believes that the type of `const Foo()` is a `_Foo` which is true at runtime, but doesn't allow much further. A lint or generator might want to understand what a `Foo` is, even if it won't exist at runtime. The lint or generator sometimes can't even tell that the object _is_ an extension type, let alone what the extension is.

For an example with Mockito, see

```dart
@GenerateNiceMocks([MockSpec<Foo>()])
```

which produces the error

```
Invalid @GenerateMocks annotation: Mockito cannot mock a private type: '_Foo'.
```

(FWIW this might be a bad example, but it's one of the cases that breaks with my real use-case for the contrived example above. Maybe Mockito itself could be updated to handle this case, I'm not sure.) I have been looking at supporting this for Angular, though, where an app might want to provide an instance of `Foo`, but the framework can't figure out how to produce code for that because all it sees is `_Foo`.

I'll conclude here rather than let this get too rambly :) Please let me know if there are cases I'm not accounting for regarding why this couldn't or shouldn't be implemented! If there's just a bandwidth issue, I'm happy to look into it myself because it's potentially limiting some other high impact work. Thanks!

bwilkerson:
@scheglov

PR #55897, opened 2024-06-01
Store the static type on TypeState

mattrbeck:
This allows users to access the static type of a DartObject through `DartObject#toStaticTypeValue`.

I'm not sure whether this is the best approach or whether it'll solve all future use-cases. I suspect it might be nicer to instead have the TypeState maintain the static type as the default "type value" and have the analyzer erase the types for comparison when necessary. At least as a maintainer of Angular, I feel like I only really care about the static type while the runtime type isn't particularly useful to me.

Please let me know if you think there's a better way to address https://github.com/dart-lang/sdk/issues/55693 or if you think this won't cover it.

For an example in use, see cl/639257690.

Apologies in advance if this PR is formatted weirdly. This is the first time I've used the built-in editor in GitHub.

### DIFF
--- a/pkg/analyzer/lib/dart/constant/value.dart
+++ b/pkg/analyzer/lib/dart/constant/value.dart
@@ -145,4 +145,13 @@ abstract class DartObject {
   /// * this object is not of type 'Type', or
   /// * the value of the object being represented is `null`.
   DartType? toTypeValue();
+
+  /// Return the representation of the static type corresponding to the value of
+  /// the object being represented, or `null` if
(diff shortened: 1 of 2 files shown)


mattrberry:
I'm not sure this is sufficient. It likely covers the case where we have a const of `Foo` where `Foo` is an extension type, but it may not work where we have a `Parameterized<Foo>`..

mattrberry:
Closing in favor of https://github.com/dart-lang/sdk/issues/56118, which provides more detail and doesn't try to prescribe a solution

Issue #55562, opened 2024-04-24
There should be a way to use already connected Socket in HttpClient connectionFactory.


LacticWhale:
* Dart SDK Version 3.3.4
* Issue related to dart core library "dart:io"

My library (socks proxy client) worked before class modifiers existed by implementing `ConnectionTask` making it constructor *public*.
Now `ConnectionTask` is *final* class and I am unable to pass already connected `socket` to `connectionFactory`.
Some code that used to work:

```dart
class SocketConnectionTask implements ConnectionTask<Socket> {
  SocketConnectionTask(this.socket);
  @override
  final Future<Socket> socket;

  @override
  void cancel() => socket.ignore();
}
```

```dart
httpClient.connectionFactory =
      (uri, proxyHost, proxyPort) async {
        // Returns instance of SocksSocket which implements Socket
        final client = SocksTCPClient.connect(...);
        
        // Secure connection after establishing Socks connection
        if(uri.scheme == 'https')
          return SocketConnectionTask((await client).secure(uri.host, ...));

        return SocketConnectionTask(client);
      };
```

I have no idea how to fix it.

haodong404:
I have the same problem, and is it possible to provide an approach for constructing a `ConnectionTask` instance?

LacticWhale:
I think laziest solution will be making `ConnectionTask` constructor public. Anyway it only used in `Socket` and considering we can extend/implement our own `Socket` I don't see why `ConnectionTask` shouldn't be available to us.

brianquinlan:
Let me see if I can recover our reasoning for making `ConnectionTask` `final`. It was probably an oversight.

brianquinlan:
Here is the relevant comment: https://dart-review.googlesource.com/c/sdk/+/291343/4/sdk/lib/io/socket.dart (line 533):
> Final, data class.
> Nobody needs to create another version.

This really is a data class so we should probably just make the constructor public. Any objections @lrhn ?

lrhn:
Reasonable to make constructor public.

A little worried that `<S>` is not constrained in any way. This is basically a genetic `CancelableOperation`, which isn't linked to sockets in any way except the naming of the future variable.

If anyone can use it as a pair of `Future<AnyType>` and `void Function()`, the name no longer makes much sense.

Not sure what to do about that. It's there any reasonable supertype of sockets that the type variable can be bounded by? If not, should we add one?

brianquinlan:
[...]

PR #56001, opened 2024-06-13
Make ConnectionTask constructor public

LacticWhale:
Something to deal with #55562.

I also changed call to constructor in few places to use public constructor.

I think there also should be no worries about `ConnectionTask` actually being `CancelableOperation`. [comment](https://github.com/dart-lang/sdk/issues/55562#issuecomment-2099414124)
No one should use ConnectionTask outside of Socket and if they do it isn't a problem.
And if later there will be `CancelableOperation` class with same functianality, nothing stops us from adding ``extends CancelableOperation`` to `ConnectionTask`.

### DIFF
--- a/sdk/lib/_internal/vm/bin/socket_patch.dart
+++ b/sdk/lib/_internal/vm/bin/socket_patch.dart
@@ -1927,7 +1927,7 @@ class _RawSocket extends Stream<RawSocketEvent>
         }
         return _RawSocket(nativeSocket);
       });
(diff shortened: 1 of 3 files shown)


brianquinlan:
Hi @LacticWhale ,

Thanks very much for the PR but there is already one in progress here:
https://dart-review.googlesource.com/c/sdk/+/367820

I'll try to finish it up in the next few days.

LacticWhale:
@brianquinlan Oh that's great. Thank you for your work.

I really need this feature.

PR #56105, opened 2024-06-29
docs: `Iterable.reduce` throws when empty

Luckey-Elijah:
Added documentation to describe the behavior when `Iterable.reduce` is called when it is an empty collection. It's also noted that one may consider using `fold` in that case.

---

<details>

</details>

### DIFF
--- a/sdk/lib/core/iterable.dart
+++ b/sdk/lib/core/iterable.dart
@@ -369,8 +369,7 @@ abstract mixin class Iterable<E> {
   /// final result = numbers.reduce((value, element) => value + element);
   /// print(result); // 17.5
   /// ```
-  /// If the iterable has no elements it will throw a state error.
-  /// Consider using [fold] if the collection may be empty.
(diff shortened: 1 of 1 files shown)


sdk/lib/core/iterable.dart
```diff
@@ -369,6 +369,8 @@ abstract mixin class Iterable<E> {
   /// final result = numbers.reduce((value, element) => value + element);
   /// print(result); // 17.5
   /// ```
+  /// If the iterable has no elements it will throw a state error.
+  /// Consider using [fold] if the collection may be empty.
```
lrhn, line 373: We usually don't document errors thrown this way.
Instead we document requirements that must be satisfied, and then leave it as an implementation details what thappens if they're not.
Thet's what the "The iterable must have at least one element." sentence above says.

Adding a "Consider using \[fold\] if the iterable can be empty." line would reasonable.
(Using "iterable" instead of "collection" because not all iterables are backed by collections.)

PR #56309, opened 2024-07-24
refactor: Clean up Invoke calls in dart_entry

eseidel:
This reduces the number of #define switches for this file.

I also think the TESTING + DART_PRECOMPILED_RUNTIME block was never actually compiled, so this fixes the type conversions in that case if it ever is compiled.

We did this upstream in Shorebird to make it easier to change the invoke syntax (to make the simulator dynamically switchable) but the clean-up seemed worth sending upstream too.


eseidel:
We've not actually landed this in our repo yet (the bots are still running), but I can update this when we do.

PR #56310, opened 2024-07-24
fix: gen_snapshot crashes if passed a non-existent file.

eseidel:
PrintErrAndExit is used from ReadFile and other functions
to exit on error, even if there hasn't yet been an Isolate
created yet.  Calling Dart_ExitScope will no current isolate
will abort, which can trigger ReportCrash on Mac and otherwise
foul up a build process instead of just returning a nice error
message on non-existent file.


eseidel:
See https://github.com/shorebirdtech/shorebird/issues/2383 for an example of the crash.

Issue #56058, opened 2024-06-20
[ffi] Varargs should give an error on uint8, int8, uint16, and int16


dcharkes:
Passing 8 bit or 16 bit ints in varargs is undefined behavior.

Clang will alert us:

```
Second argument to 'va_arg' is of promotable type 'uint16_t' (aka 'unsigned short'); this va_arg has undefined behavior because arguments will be promoted to 'int'clang(-Wvarargs)
```

But we allow this in `dart:ffi`, leading to undefined behavior.

We should consider adding an error message.

Discovered on:

* https://github.com/dart-lang/native/issues/1191

vaishnavi-2901:
hey
can work on it, please assign this task to me

dcharkes:
> hey can work on it, please assign this task to me

Hi @vaishnavi-2901! That's great!

Please start with https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md

A PR for this issue will involve adding error messages both to the CFE (the compiler used when doing `dart ...`) and analyzer (used in the IDE for giving error messages).

Some pointers to point you in the right direction:

* Analyzer
  * https://dart-review.googlesource.com/c/sdk/+/278343 analyzer implementation for varargs, not including any error messages.
  * https://dart-review.googlesource.com/c/sdk/+/336040 How error messages are defined.
  * You can see in the ffi_verifier.dart file in the first PR how to emit error messages from the analyzer.
* CFE
  * https://dart-review.googlesource.com/c/sdk/+/276921/49/pkg/vm/lib/transformations/ffi/common.dart -> This is where the error checking for the FFI is done.
* Tests
  * The tests should be in here. tests/ffi/static_checks/vmspecific_static_checks_varargs_test.dart Note that this test needs to be converted from `//# 1: compile-time error` to tests/ffi/static_checks/vmspecific_static_checks_test.dart style with `// [cfe]`.

It's probably easiest to

1. First convert the test
2. Then make an analyzer PR
3. And then in a third PR try to tackle the CFE.

trivedikavya:
hey sir can i work on it ..?

dcharkes:
@vaishnavi-2901 are you still working on this?

trivedikavya:
Dear respected sir let she work and let me also work on it. as I am beginning the open source contribution first you will check the request of both and then you will see what to do 
I have one Question sir just tell me in which file I have to contribute .

codesculpture:
[...]

PR #56359, opened 2024-08-01
resolved [ #56058 ]

vijayindalkar:
**Issue**:   #56058 - List.removeWhere method not correctly handling == operator overload

The List.removeWhere method in the Dart SDK's list.dart file was modified to correctly handle the == operator overload. Previously, the method was not properly removing elements that satisfied the test function and matched the element to be removed.

**Changes:**

Modified the removeWhere method implementation to add a check for value == element before removing the element from the list.
Ensured that the == operator overload is correctly used to compare elements.


mraleph:
**Warning**: don't send meaningless PRs or you will get banned from the repo.

PR #56545, opened 2024-08-21
`PropertyAccessorElementImpl`: optimize `name` getter to avoid calls to `considerCanonicalizeString`

gmpassos:
The `name` getter is a hotspot and shouldn't resolve `String` canonicalization for each call.

---

- [✅] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


mraleph:
@gmpassos what is the status of this change?

gmpassos:
@scheglov reviewed this and performed some benchmarks.

I plan to run additional benchmarks to demonstrate that this optimization is significant.

The key discussion is whether `PropertyAccessorElementImpl` should have an additional field to bypass `considerCanonicalizeString`.

IMHO, given the current level of optimization in the analyzer, achieving an extra 10% improvement will likely require accumulating several smaller optimizations, each contributing around 1%.

mraleph:
@bwilkerson @scheglov please make a call if you want to accept or reject this PR and if you would like to accept submit corresponding CL to CQ (FWIW @bwilkerson has already voted +1).

PR #56851, opened 2024-10-04
Update socket_win.cc

beelabcloud:
Fix windows SIO_UDP_CONNRESET bug, the listener would be closed when remote is not reachable if the listener acts as sender for sending data to remote.


mit-mit:
Closing as unactionable as the CLA wasn't completed

PR #56857, opened 2024-10-04
Update the fastestMarathon to 2023 human standards.

FlorinBalint:
Simple change to the documentation of duration class: updating the current world record time for the marathon,


mit-mit:
✅

Issue #56521, opened 2024-08-20
Expose `architecture` from `Abi` object


miquelbeltran:
Hello Dart team!

Currently, the easiest way to obtain the compiled architecture of an app is to call to `Abi.current().toString()` and then parse the `String`, e.g.

```
final abi = Abi.current().toString();
final architecture = abi.split('_').last;
```

Could it be possible to expose the `Architecture` property directly?

Related ticket: https://github.com/dart-lang/sdk/issues/40231

lrhn:
As Josh Bloch said in Effective Java, don't provide information in `toString` that you don't provide otherwise. People *will* parse `toString`, and then its format becomes part of your API.

miquelbeltran:
Even the Dart team does it ;)

https://github.com/dart-lang/native/blob/4a18bf2964b69075d9ec3ec1295d4985494741bb/pkgs/native_assets_cli/test/api/target_test.dart#L18

lrhn:
There's probably a reason he said that - everybody does it, until they learn why not to, often the hard way.

dcharkes:
Hi @miquelbeltran!

I guess the main question is whether such functionality should live in the Dart SDK, or in a package.
And if it lives in a package, whether it should be a different package than `native_assets_cli`: https://pub.dev/documentation/native_assets_cli/latest/native_assets_cli/Architecture-class.html

Also, if we do `Architecture`, we should also talk about `OS`.
We also don't have an `OS` in `dart:`. And the `Platform` can only talk about the current OS.
`package:platform` also doesn't have an `OS`, but it can use `FakePlatform` to talk about a different OS.
Because `package:platform` is already the way to deal with OS (not the cleanest, but okay), I don't think we'll ever add `OS` to `dart:` (I believe I opened an issue about this years back, but i can't find it anymore.)

Pros for package:

* Smaller SDK
* Easier to iterate
* OS will also not be in `dart:`

Pros for `dart:`

* No rolling when `Abi`s are added

I think I'm leaning towards a package instead of `dart:`.

Even though `package:native_assets_cli` is meant for writing build hooks, nothing prevents you from using that package for the `Architecture` class in other contexts.

miquelbeltran:
Hi Daco!

My pros for having that in the SDK itself is discoverability and one less 3rd party dependency in code.

`native_assets_cli` is currently listed as experimental, so I was a bit wary about adding it to my project.

[...]

PR #56959, opened 2024-10-24
[dart:io] Adds Platform.architecture

RossComputerGuy:
- Adds `Platform.architecture` into `dart:io`, this could be useful to easily detect the CPU architecture
- Adds `--target-arch` to pair with the `--target-os` flag, maybe someday these two can support full cross compilation

---

<details>

</details>


pkg/dart2native/lib/generate.dart
```diff
@@ -120,6 +122,11 @@ class _Generator {
   /// Specifies the file debugging information should be written to.
   final String? _debugFile;
 
+  /// Specifies the CPU architecture the executable is being generated for. This
+  /// must be provided when [_kind] is [Kind.exe], and it must match the current
+  /// CPU architecture.
+  final String? _targetArch;
```
alestiago, line 128: According to Dart Doc, the first line should be brief: https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

Doing as suggested will improve the generated API documentation (`dart doc`) experience.

```suggestion
  /// Specifies the CPU architecture the executable is being generated for. 
  ///
  /// This must be provided when [_kind] is [Kind.exe], and it must match the current
  /// CPU architecture.
  final String? _targetArch;
```

pkg/dart2native/lib/generate.dart
```diff
@@ -182,6 +191,13 @@ class _Generator {
         _sourcePath = _normalize(sourceFile)!,
         _packages = _normalize(packages) {
     if (_kind == Kind.exe) {
+      if (_targetArch == null) {
+        throw ArgumentError('targetArch must be specified for executables.');
```
alestiago, line 195: I believe the Flutter team once went over the errors the tooling could throw and added clear explanation points on how to move forward. I would expect such practice to also be present here.

If I encounter "targetArch must be specified for executables" I would ask myself:
- "what is targetArch", is an abbreviation for "target architecture" (see [Effective Dart for abbreviations](https://dart.dev/effective-dart/documentation#avoid-abbreviations-and-acronyms-unless-they-are-obvious))
- How can I specify so? Is it a parameter flag `--targetArch`>
- What are the available architectures I can compile into? Is there documentation stating all the supported architectures? Maybe we can point the reader to: https://github.com/dart-lang/sdk/blob/main/docs/Supported-Architectures.md?
- What architecture is my system running? Could the message include a suggested resolution with the current architecture my system is using?

mraleph:
cc @lrhn what is out stance on this? see also related issue: https://github.com/dart-lang/sdk/issues/56521

pkg/dart2native/lib/generate.dart
RossComputerGuy: Alright, I wasn't aware of that since I based it on `_targetOS` for the doc comment.

pkg/dart2native/lib/generate.dart
RossComputerGuy: > "what is targetArch", is it an abbreviation for "target architecture"?

Yes, it is an abbreviation for "target architecture". I was thinking `targetCPU` but in my experience, that usually is for specific CPU optimizations (eg. zen1, skylake, generic, baseline).

> How can I specify so? Is it a parameter option `--targetArch`? Is there an example on how to specify it?

`--target-arch` was added to pair with `--target-os`. However, `--target-os` doesn't support cross compiling and so I haven't been able to get `--target-arch` to cross compile. I at least added the logic for `--target-arch` to pair with the existing `--target-os` flag.

> What are the available architectures I can compile into? Is there documentation stating all the supported architectures? Maybe we can point the reader to: https://github.com/dart-lang/sdk/blob/main/docs/Supported-Architectures.md? Note that these are already in the "help command with TargetArch.names, but it is not as clear as the table provided in the documentation.

The `TargetArch.names` I added in `pkg/vm/lib/target_arch.dart` was based on that table + observations I saw when using `grep` to figure out how this repo works.

> What architecture is my system running? Could the message include a suggested resolution with the current architecture my system is using?

Could you provide more information to what this sort of means?

> For example, a new improved message would be (ignore bad formatting and bad command

Should we also change the one for the target OS?

pkg/dart2native/lib/generate.dart
RossComputerGuy: I've updated the comment.

lrhn:
If we do add this to the platform libraries, I'd put it in `dart:ffi`.
Nothing else in the platform libraries cares about which kind of CPU the program is running on.

The usual caveat about exposing information applies: You can always provide more detailed information, until you have made a complete API for querying `cpuid`.

What is an "architecture"?
Is it something identifying the underlying CPU's capabilities, or is it about the current process? (Will a 32-bit x86 Dart SDK running on an x64 CPU report itself as one or the other? Do you need to know the other? What about something compiled as x32?)

Will any answer we give here actually be enough for users? (And for what?)

Nothing new here. We have the same issue with `Platform.operatingSystem`.
Sometimes it's just not fine-grained enough for what people want to do

RossComputerGuy:
> What is an "architecture"?
> Is it something identifying the underlying CPU's capabilities, or is it about the current process?

Ime, it's always referring to the actual ISA of the CPU.

> Will a 32-bit x86 Dart SDK running on an x64 CPU report itself as one or the other?

It reports what the host is. I think the closest example to look at is Node.js's `os.arch()` which reports the architecture the node binary was compiled for. So if Dart was compiled for 32-bit x86, it would report as that. But I guess it depends on how the build was configured.

> Will any answer we give here actually be enough for users? 

I think it depends, if it's an ffi related use then the abi should be used to determine the architecture. This property could be useful when a program wishes to download a binary and it depends on the architecture or maybe like a compiler where ffi isn't used inside the compiler.

> Nothing new here. We have the same issue with `Platform.operatingSystem`.
Sometimes it's just not fine-grained enough for what people want to do

Definitely, it's hard to determine what exactly every use case is.

PR #56988, opened 2024-10-29
chore: add actual vs. expected in error logging message

a-wallen:
Having a more verbose error message comparing the actual versus the expected message helps when you're debugging the lint. Especially since the analyzed code is a raw string, so it doesn't have intellisense to point out when you've made a typo.


a-wallen:
cc @srawlins

PR #57086, opened 2024-11-14
add modeOctalString to FileStat for compatibility with Posix tools

stephane-archer:

add modeOctalString to FileStat for compatibility with Posix tools

### DIFF
--- a/sdk/lib/io/file_system_entity.dart
+++ b/sdk/lib/io/file_system_entity.dart
@@ -213,6 +213,17 @@ FileStat: type $type
   }
 }
 
+/// The mode value as a Unix octal string.
+String modeOctalString() {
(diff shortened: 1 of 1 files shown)


mraleph:
Thanks for the PR but this is just an equivalent to `stat.mode.toRadixString(8)`. I don't think there is any need to add it as a separate method. Also it is a breaking change for anything implementing the `FileStat` class.

stephane-archer:
I was not aware of "stat.mode.toRadixString(8)" is it exactly equivalent?

mraleph:
I guess you might also need to mask things if you don't care about top bits: `(stat.mode & 0x1FF).toRadixString(8)`

Issue #55989, opened 2024-06-12
[macros] JsonCodable enum's "Only classes are supported as field types for serializable classes"


jimmyff:
Enum's aren't supported by the `JsonCodable` macro, this is a bit of a shame as string's don't enforce options and classes are still verbose if you have a lot of enum data types. 

Couldn't the enums be generated as built enum classes like [built_value](https://github.com/google/built_value.dart) does?

jimmyff:
Is there a way I could polyfill the solution for now? If the package supported annotations for custom fromJson/toJson' (like `json_serializable`) that could work:

```dart
  @JsonKey(
      name: 'foobar',
      fromJson: _deserializeStringToEnumClass,
      toJson: _serializeEnumClassToString)
```

jakemac53:
> Is there a way I could polyfill the solution for now? If the package supported annotations for custom serializer and deserializer (a bit like `json_serializable` does it with it's `JsonKey()`)?

The package doesn't support configuration at all because annotations don't work for macros yet.

Note that this package is not ready for production use (the feature is experimental still and will have breaking changes).

cedvdb:
Should enum work out of the box without being annotated with `@JsonCodable` or should they require it ?

Adding JsonCodable to the enum would generate a `static fromJson` , which is currently not supported by the macro. 

It's also possible to just do enums automatically:

```dart
    if (declaration is EnumDeclaration) {
      return RawCode.fromParts([
        if (nullCheck != null) nullCheck,
        valueReference,
        '.name'
      ]);
    }
```

PR #56665, opened 2024-09-07
Support handling classDecl fromJson(String s) in JsonDecodable

morlay:
To support usage like

```dart
@JsonCodable()
class Account {
    AccountType accountType
}

class AccountType {
  static const USER = AccountType("USER");
  static const AGENT = AccountType("AGENT");

  const AccountType(this.value);

  final String value;

  factory AccountType.fromJson(String value) {
    return AccountType(value);
  }

  String toJson() {
    return value;
  }
}
```

PR #59581, opened 2024-11-21
Json codable no type assumption in custom fromJson

### LINKED ISSUE #59580: [json_codable] the object passed to fromJson is not necessarily a Map
Currently json_codable assumes that a Map is passed to the fromJson constructor when that constructor is defined by the user. 

This is not supported:

```dart

@JsonCodable()
class A {
  final CustomStringSerialization customStringSerializationField;
}

class CustomStringSerialization {
  final String a;

  CustomStringSerialization(this.a);

  String toJson() => a;

  factory CustomStringSerialization.fromJson(String a) =>
      CustomStringSerialization(a);
}

```

Use cases:

  - deserializing values to object, for example a 'EUR' value could be deserialized to a currency object with its format method
  - it will be necessary for enum support

cedvdb:
- fix https://github.com/dart-lang/sdk/issues/59580
- related: https://github.com/dart-lang/sdk/issues/55989

 > json_codable wrongly assumes fromJson receives a Map

This allows an user defined class to have a `fromJson` method with a parameter that is not a `Map`. Before this change the `@JsonCodable` macro would check if a field has a fromJson constructor and assume that the parameter type was a `Map`. This PR uses the type of the parameter instead.

Note: I have this weird behavior where I can only run the tests with the macros enabled once per branch, so I have to switch branches and run the tests on another.

### DIFF
--- a/pkg/json/test/json_codable_test.dart
+++ b/pkg/json/test/json_codable_test.dart
@@ -25,6 +25,10 @@ void main() {
         'mapOfSerializableField': {
           'c': {'x': 3}
         },
+        'customStringSerializationField': 'hi',
+        'customMapSerializationField': {
+          'z': 'zzz',
+        }
       };
 
       var a = A.fromJson(json);
@@ -36,6 +40,9 @@ void main() {
       expect(a.listOfSerializableField.single.x, 1);
       expect(a.setOfSerializableField.single.x, 2);
       expect(a.mapOfSerializableField['c']!.x, 3);
+      expect(a.customStringSerializationField, CustomStringSerialization('hi'));
+      expect(
+          a.customMapSerializationField, CustomMapSerialization({'z': 'zzz'}));
 
       expect(a.toJson(), equals(json));
     });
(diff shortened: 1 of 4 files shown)


cedvdb:
@jakemac53 not sure if updates to this package are welcome yet or not. I've another down the drain for enums but I cannot make it work currently because the behavior is different during test than in the augmented code which is puzzling

cedvdb:
It seems there is already a PR for this here : https://github.com/dart-lang/sdk/pull/56665

natebosch:
Can a class also define how to handle `null`? Would `factory Foo.fromJson(Object? o)` or `factory Foo.fromJson(String? s)` work? Can we add a test if so?

cedvdb:
> Can a class also define how to handle `null`? Would `factory Foo.fromJson(Object? o)` or `factory Foo.fromJson(String? s)` work? Can we add a test if so?

It does not currently handle that case no

pkg/json/test/json_codable_test.dart
```diff
@@ -38,6 +43,11 @@ void main() {
       expect(a.listOfSerializableField.single.x, 1);
       expect(a.setOfSerializableField.single.x, 2);
       expect(a.mapOfSerializableField['c']!.x, 3);
+      expect(a.customStringSerializationField, CustomStringSerialization('hi'));
+      expect(a.customNullableStringSerialization,
```
cedvdb, line 47: @natebosch added requested test

cedvdb:
@natebosch I added a test for String? o

jakemac53:
This is released now 👍

PR #60037, opened 2025-02-03
Replace link to non-existent language versioning document with current dev docs on subject

timmaffett:
This PR removes a link to a language versioning document that no longer exists and replaces it with a link to the current language versioning documentation.

### DIFF
--- a/tools/experimental_features.yaml
+++ b/tools/experimental_features.yaml
@@ -48,8 +48,7 @@
 #
 #    A version less than this version may be specified in a .packages file
 #    or in a library language version override (e.g. // @dart = 2.1)
-#    to disable this feature. For more on library language version override, see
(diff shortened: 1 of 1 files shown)


tools/experimental_features.yaml
```diff
@@ -49,7 +49,7 @@
 #    A version less than this version may be specified in a .packages file
 #    or in a library language version override (e.g. // @dart = 2.1)
 #    to disable this feature. For more on library language version override, see
-#    https://github.com/dart-lang/language/blob/master/accepted/future-releases/language-versioning/language-versioning.md
+#    https://dart.dev/resources/language/evolution#per-library-language-version-selection
```
parlough, line 52: Consider updating this link to the https://dart.dev/to/language-version-override redirect. That way if the doc location or fragment ever changes, this link won't need updating again.

```suggestion
#    https://dart.dev/to/language-version-override
```

PR #60048, opened 2025-02-04
Add credentials for bearer token authentication

richirisu:
I recently needed _bearer_ authentication for a proxy, however, the current HTTP implementation only offered credentials for _basic_ and _digest_ authentication.

It applies to the [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) and [Proxy-Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization) headers.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes

**Implementation Details**

The token provided to the HttpClientBearerCredentials is checked for valid characters according to [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750).

It is possible to provide any string as token by applying base64Encode in conjunction with utf8.encode first. I was considering making a factory method that automates this, but then I felt it would just complicate things, as bearer tokens are most often randomly generated anyway.

**Tests**

Sorry, but I didn't provide any tests here.

### DIFF
--- a/sdk/lib/_http/http.dart
+++ b/sdk/lib/_http/http.dart
@@ -2065,6 +2065,13 @@ abstract final class HttpClientBasicCredentials
       _HttpClientBasicCredentials(username, password);
 }
 
+/// Represents credentials for bearer token authentication.
+abstract final class HttpClientBearerCredentials
+    implements HttpClientCredentials {
+  factory HttpClientBearerCredentials(String token) =>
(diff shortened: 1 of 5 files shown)


mraleph:
@richirisu We can't look at the code before CLA is signed - could you sign it?

PR #60229, opened 2025-02-28
Fix Flow Analysis Bug in Dart SDK (#60114)

### LINKED ISSUE #60114: Wrong flow analysis after non-nulable type and `==null`
The following code works well and produces no errors in both CFE and the analyzer.
```dart
main() {
  late int i;
  String s = "";
  if (s == null) {  // Warning: unnecessary_null_comparison
    i = 42;
  }
  i; // No error
}
```
But according to the flow analysis [specification](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#expressions):

> operator== If N is an expression of the form `E1 == E2`, where the static type of `E1` is `T1` and the static type of `E2` is `T2`, then:
> ...
> - Otherwise, if `equivalentToNull(T1)` and `T2` is non-nullable, or `equivalentToNull(T2)` and `T`1 is non-nullable, then:
>   - Let `true(N) = unreachable(after(E2))`.
>   - Let `false(N) = after(E2)`.

According to the above, the body of `if (s == null) ...` is unreachable and therefore initialization of `i` occurs in a dead code. This leaves `i` definitely unassigned and therefore `i;` should be a compile-time error.

Probably this happens because `s == null` is not always `false` in an unsound mode, but there are no mentions of unsound mode in the flow analysis spec.

cc @johnniwinther 

`Dart SDK version: 3.8.0-73.0.dev (dev) (Thu Feb 6 20:02:31 2025 -0800) on "windows_x64"`

jhaaryan597:
This PR fixes the incorrect flow analysis after non-nullable type comparison with `== null`.  
- Ensures proper unreachable code detection  
- Prevents late variable usage without assignment  
Fixes #60114

### DIFF
--- a/pkg/_fe_analyzer_shared/lib/src/flow_analysis/flow_analysis.dart
+++ b/pkg/_fe_analyzer_shared/lib/src/flow_analysis/flow_analysis.dart
@@ -4758,15 +4758,23 @@ class _FlowAnalysisImpl<Node extends Object, Statement extends Node,
     assert(_unmatched == null);
     assert(_scrutineeReference == null);
   }
-
+//aryanjha597
   @override
-  void for_bodyBegin(Statement? node, Expression? condition) {
+void for_bodyBegin(Statement? node, Expression? condition) {
+  if (condition != null && condition.staticType.isNonNullable && condition is BinaryExpression && condition.operator == "==") {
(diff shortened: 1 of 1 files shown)


mraleph:
Please sign the CLA if you want the code to be reviewed.

mraleph:
Closing as stale as the author did not sign CLA.

PR #60412, opened 2025-03-26
Update runners to ubuntu-24.04 from deprecated ubuntu-20.04 label

google-admin:
This is a LSC run by http://go/ghss to upgrade all depreacated ubuntu-20.04 runners to ubuntu-24.04.

On April 1, 2025, GitHub will stop supporting ubuntu-20.04 runners and workflows configured with this label will cease to run.  This PR is an attempt to save you work by doing the upgrade for you.

WARNING: We do not know if the updated label is compatiable with your workflow or not.

If you do not want to merge this PR, feel free to close it and deal with the problem yourselves.

More context and feedback: http://b/406537467

### DIFF
--- a/.github/workflows/third-party-deps-scan.yml
+++ b/.github/workflows/third-party-deps-scan.yml
@@ -14,7 +14,7 @@ permissions: read-all
 jobs:
   extract-deps:
     name: Extract Dependencies
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


athomas:
@billnapier What the...? Why was this pushed straight to our GitHub mirror!? This broke our copybara sync and now we have to do manual cleanup.

PR #61071, opened 2025-07-07
reimplement `MapEntry` as a record

ghost:
As stated by its documentation, `MapEntry` has been planned to be changed to a value type. This change achieves that by making it a `Record` `(K, V)`. Backwards compatibility is mostly achieved by preserving the original API through an `extension type` and ensuring it `implements Object`. See https://github.com/dart-lang/sdk/pull/61071#issuecomment-3046299969 for what would not be backwards compatible

`asPair` is added to the API to allow direct usage as a record, it was required anyways since it is the backing field

### DIFF
--- a/sdk/lib/core/map.dart
+++ b/sdk/lib/core/map.dart
@@ -478,13 +478,7 @@ abstract interface class Map<K, V> {
 /// ]);
 /// print(map); // {1: A, 2: B, 3: C, 4: D}
 /// ```
-///
-/// Do not extend or implement the `MapEntry` class.
-/// If the Dart language introduces value types,
-/// the `MapEntry` class will be changed to such a type,
-/// and will likely no longer be able to be implemented or extended
-/// by classes.
-final class MapEntry<K, V> {
+extension type const MapEntry<K, V>.fromPair((K, V) asPair) implements Object {
   /// The key of the entry.
   ///
   /// ```dart
@@ -492,7 +486,7 @@ final class MapEntry<K, V> {
   /// var entry = map.entries.first; // MapEntry<String, String>
   /// print(entry.key); // 'theKey'
   /// ```
-  final K key;
+  K get key => asPair.$1;
 
   /// The value associated to [key] in a map.
   ///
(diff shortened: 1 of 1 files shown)


ghost:
Now the record constructor is accessible and named `fromPair`

ghost:
cc @bwilkerson @keertip

mraleph:
See also discussions on https://github.com/dart-lang/sdk/issues/54965

This is technically a breaking change so it requires breaking change process, and even before that it requires @lrhn to pre-approve the design.

ghost:
@mraleph thanks for linking that and tagging @lrhn. This change uses an `extension type` and includes the original API as a subset of its API* and therefore should not be breaking

\* the `_` constructor is removed but that is internal and was not used anywhere, so I simply omitted it since it was not needed

mraleph:
It is breaking because it changes runtime semantics: the difference between old and new version is observable via `is` / `as` checks as well as through equality and identity of `MapEntry` (original class did not override `==`). 

That being said the breakage is probably very limited - `MapEntry` should only really occur as temporary objects.

ghost:
> It is breaking because it changes runtime semantics: the difference between old and new version is observable via `is` / `as` checks as well as through equality and identity of `MapEntry` (original class did not override `==`).
> 
> That being said the breakage is probably very limited - `MapEntry` should only really occur as temporary objects.

Good points, I did not consider those things. The original class-based `MapEntry` is replaced in this change so any `MapEntry` `is`/`as` behaviour should still be preserved for `MapEntry` but if `(K, V)` or `Record` happens to be checked or casted it would now work differently, as you have said eg
```dart
(1, 'hello') is MapEntry<int, String> // used to be false, now true
(1, 'hello') as MapEntry<int, String> // used to throw Error, now it works

f (MapEntry<int, String> e) {
  e is Record  // used to be false, now true
  e is (int, String)  // used to be false, now true
}
```

Also agree this changes `==` to the record defined behaviour from the original identity based one

lrhn:
It's not the time, yet. 

The Dart 3.0 language changes to the platform libraries, including adding `final` to `MapEntry`, were made conditional on the language version of the client library.

That means that pre-3.0 language version libraries can extend `MapEntry`, and changing the class to an extension type will break that. 
So it's a breaking change as long as the SDK supports language version 2.19.

Other than that, I don't mind `MapEntry` having equality. In fact, I thought I did that already.
 
I wouldn't expose `asPair`.
I would consider using a `({K key, V value})` as representation record, so that dynamic access would keep working. 

But not yet.

ghost:
Sounds good, I will close this and leave it up since it was informative about the specifics of the challenges

I agree with `({K key, V value})` and I suppose `asPair` is better off not being exposed since it would just introduce extra complications to code that can also limit future changes

Issue #60191, opened 2025-02-22
Linux: Changing timezone doesn't update DateTime


AndreaRicchi:
The issue occurs because the `DateTime.now()` method retrieves the current local time based on the system clock when it is called. However, in Linux, when the system timezone is changed while the application is running, the runtime may not immediately reflect the updated timezone. Instead, it continues using the previously cached timezone settings.

The issue was opened in the Flutter project ( https://github.com/flutter/flutter/issues/163460 )  and moved here.

This is the test program:
```dart
import 'package:process_run/process_run.dart';

void main() async {
  String systemDateTime = '';
  var shell = Shell();

  while (true) {
    var results = shell.runSync('date');
    systemDateTime = results.outText.trim();
    print('Current time - `${DateTime.now()}`');
    print('Current timezone - `${DateTime.now().timeZoneName}`');
    print('Current time from shell - `$systemDateTime`');
    await Future.delayed(Duration(seconds: 3));
  }
}
```

This is the output:
```
❯ dart bin/test_clock_dart.dart
Current time - `2025-02-22 20:34:24.384515`
Current timezone - `CET`
Current time from shell - `Sat Feb 22 08:34:24 PM CET 2025`
Current time - `2025-02-22 20:34:27.398776`
Current timezone - `CET`
Current time from shell - `Sat Feb 22 07:34:27 PM GMT 2025`
Current time - `2025-02-22 20:34:30.409463`
Current timezone - `CET`
Current time from shell - `Sat Feb 22 11:34:30 AM PST 2025`
Current time - `2025-02-22 20:34:33.423042`
Current timezone - `CET`
Current time from shell - `Sat Feb 22 03:34:33 PM -04 2025`
Current time - `2025-02-22 20:34:36.435628`
Current timezone - `CET`
Current time from shell - `Sat Feb 22 09:34:36 PM CAT 2025`
Current time - `2025-02-22 20:34:39.449030`
Current timezone - `CET`
Current time from shell - `Sun Feb 23 12:34:39 AM +05 2025`
Current time - `2025-02-22 20:34:42.461992`
Current timezone - `CET`
Current time from shell - `Sun Feb 23 04:19:42 AM +0845 2025`
Current time - `2025-02-22 20:34:45.471931`
Current timezone - `CET`
Current time from shell - `Sun Feb 23 08:34:45 AM NZDT 2025`
```

lrhn:
[...]

Issue #53276, opened 2023-08-18
Timezone reported as UTC even though it is GMT+02:00


andynewman10:
### Steps to reproduce

1. Create a Flutter application on a device running Android 7.1.1. You can use an Android 7.1.1 emulator or, eg., a Sony Xperia Z5 Compact device running Android 7.1.1 (both experience the same issue)
2. Go to the Settings, Date/Time and make sure automatic time is on ; manually set the time zone to something other than GMT (for instance: I use Paris, France, which is GMT+2 in Summer)
3. Add the following code:
```
DateTime dateTime = DateTime.now();
```
4. Run the code and inspect the value of `dateTime`, either by printing it or using the debugger (I use VS code).

### Expected results

In Paris, France, `dateTime.toString()` should be equal to
```
DateTime (2023-08-17 21:23:23.241901)
```
and the time zone should be:
```
timeZoneName: "CEST"
timeZoneOffset: Duration (2:00:...)
```
In other words, the time should be `21:23` in the `CEST` timezone.

### Actual results

`dateTime.toString()` is
```
DateTime (2023-08-17 19:23:23.241901)
```
and the timezone reported is:
```
timeZoneName: "GMT"
timeZoneOffset: Duration (0:00:...)
```
In other words, the time is correct (in its reported timezone: it really is `19:23` in `UTC/GMT` timezone), but the timezone is not correct.

The same program runs fine (no timezone error) on

- an iPhone 8 with iOS 16.6
- a Samsung S5 Mini (SM-G800F) with Android 6.0.1

### Comparison with a native Android implementation

I tested with native (Java) code on a native Android application on my Sony Z5 Compact. I use Android Studio Electric Eel.

1. Create a native Android application in Android Studio, use the "Basic Activity" template
2. Add the following code:

```
                SimpleDateFormat timezoneOffsetFormatter = new SimpleDateFormat("ZZZZZ");
                String timezoneNotation = timezoneOffsetFormatter.format(new Date()); // "+02:00"
                String timezoneLocation = TimeZone.getDefault().getID(); // "Europe/Paris"
                String tzString = timezoneNotation + "," + timezoneLocation;
```
3. Run the application on the same device (Android 7.1.1 emulator or Sony Xperia Z5 Compact with Android 7.1.1).

Look at the `tzString` value. This value is `+02:00,Europe/Paris`.

### Code sample

<details><summary>Code sample</summary>

```dart
DateTime dateTime = DateTime.now();
```
</details>

[...]

PR #61302, opened 2025-08-13
[vm/runtime] Call `tzset()` before `localtime_r()`.

JesseRiemens:
Relevant info in #60191 (And duplicate #61303)

POSIX doesn't guarantee that the timezone is set correctly before calling `localtime_r()`, so we need to call `tzset()` first to ensure that the timezone information is up-to-date.

As the glibc manual states:
> According to POSIX.1-2001, `localtime()` is required to behave as though `tzset(3)` was called, while `localtime_r()` does not
> have this requirement.  For portable code, `tzset(3)` should be called before `localtime_r()`.

---

<details>

</details>

### DIFF
--- a/runtime/vm/os_android.cc
+++ b/runtime/vm/os_android.cc
@@ -102,6 +102,9 @@ intptr_t OS::ProcessId() {
 static bool LocalTime(int64_t seconds_since_epoch, tm* tm_result) {
   time_t seconds = static_cast<time_t>(seconds_since_epoch);
   if (seconds != seconds_since_epoch) return false;
+  // No need to call tzset() before localtime_r() because bionic
+  // will handle timezone changes for us (starting from Android O).
+  // See https://android.googlesource.com/platform/bionic/+/ea87716696bf635706b6f3fa56b8a145add83aff
   struct tm* error_code = localtime_r(&seconds, tm_result);
   return error_code != nullptr;
 }
--- a/runtime/vm/os_linux.cc
+++ b/runtime/vm/os_linux.cc
@@ -418,6 +418,7 @@ intptr_t OS::ProcessId() {
 static bool LocalTime(int64_t seconds_since_epoch, tm* tm_result) {
   time_t seconds = static_cast<time_t>(seconds_since_epoch);
   if (seconds != seconds_since_epoch) return false;
+  tzset();  // Not guaranteed by POSIX to be called by `localtime_r`.
   struct tm* error_code = localtime_r(&seconds, tm_result);
   return error_code != nullptr;
 }


mraleph:
Do you know what is the cost associated with calling `tzset` here? I would imagine it does a bunch of stuff to figure out the system timezone, how much does this add on top of what `localtime_r` costs?

JesseRiemens:
> Do you know what is the cost associated with calling `tzset` here? I would imagine it does a bunch of stuff to figure out the system timezone, how much does this add on top of what `localtime_r` costs?

I made [a quick benchmark](https://github.com/JesseRiemens/dart_timezone_bug_mwe/blob/main/bin/time_benchmark.dart) that does ten runs of 10 million DateTime.now() calls, and the results are below:

The modded version, with the `tzset()` call:
```
~/Applications/dart-sdk-test/dart-modified/sdk/out/ReleaseX64/dart-sdk/bin/dart run ./bin/time_benchmark.dart  
Total time taken for 10000000 iterations: 345942 microseconds
Average time taken: 0.0345942 microseconds
Total time taken for 10000000 iterations: 345456 microseconds
Average time taken: 0.0345456 microseconds
Total time taken for 10000000 iterations: 387595 microseconds
Average time taken: 0.0387595 microseconds
Total time taken for 10000000 iterations: 344673 microseconds
Average time taken: 0.0344673 microseconds
Total time taken for 10000000 iterations: 343620 microseconds
Average time taken: 0.034362 microseconds
Total time taken for 10000000 iterations: 344166 microseconds
Average time taken: 0.0344166 microseconds
Total time taken for 10000000 iterations: 339798 microseconds
Average time taken: 0.0339798 microseconds
Total time taken for 10000000 iterations: 337510 microseconds
Average time taken: 0.033751 microseconds
Total time taken for 10000000 iterations: 337566 microseconds
Average time taken: 0.0337566 microseconds
Total time taken for 10000000 iterations: 340555 microseconds
Average time taken: 0.0340555 microseconds
Total time taken for all iterations: 3466881 microseconds
Average time taken per iteration: 0.03466881 microseconds
```

The non-modified version:
```
~/Applications/dart-sdk-test/dart-without-mod/sdk/out/ReleaseX64/dart-sdk/bin/dart run ./bin/time_benchmark.dart
Total time taken for 10000000 iterations: 344923 microseconds
Average time taken: 0.0344923 microseconds
Total time taken for 10000000 iterations: 350685 microseconds
Average time taken: 0.0350685 microseconds
Total time taken for 10000000 iterations: 355197 microseconds
Average time taken: 0.0355197 microseconds
Total time taken for 10000000 iterations: 338760 microseconds
Average time taken: 0.033876 microseconds
Total time taken for 10000000 iterations: 338500 microseconds
Average time taken: 0.03385 microseconds
Total time taken for 10000000 iterations: 339348 microseconds
Average time taken: 0.0339348 microseconds
Total time taken for 10000000 iterations: 338644 microseconds
Average time taken: 0.0338644 microseconds
Total time taken for 10000000 iterations: 345835 microseconds
Average time taken: 0.0345835 microseconds
Total time taken for 10000000 iterations: 344627 microseconds
Average time taken: 0.0344627 microseconds
Total time taken for 10000000 iterations: 339836 microseconds
Average time taken: 0.0339836 microseconds
Total time taken for all iterations: 3436355 microseconds
Average time taken per iteration: 0.03436355 microseconds
```

That's 0.03466881 with mod vs 0.03436355 without mod; 
That's 0.9% (1.0089) increase. 

If you have any improvement to my benchmark method, please let me know. 
It seems like the difference compared to the rest of the overhead is very small. 

Edit: It actually looks like after repeated testing that the modified build is faster often, and I can't understand why.

mraleph:
I think this looks good to me. Could you apply the same patch to `os_android` (I assume it has the same problem?) and then I can land it.

JesseRiemens:
If I'm correct, this fix was already applied on Android and MacOS but reverted for all three platforms for performance issue reasons on Android:
https://github.com/dart-lang/sdk/issues/53276

I could create that patch no issue, but I don't have a clue how I would be able to performance test it for Android. I do have a Mac mini m1 here, so if it's required, I could do the same test as I did for Linux.

mraleph:
@JesseRiemens thanks for the link. It seems to have been reverted because there were some TSAN failures. Though there is no information on what exactly was failing and I don't see any warnings around multithreading in `tzset` man pages. I will add a TSAN bot to the CL to test. 

Anyway, ignore my comment about Android because apparently they have fixed in Android O - `localtime_r` behaves as if it calls `tzset`, so we don't need to call it explicitly.

mraleph:
Could you maybe add the following to `os_android.cc` before `localtime_r`? So that we don't ask this question next time:

```
// No need to call tzset() before localtime_r() because bionic 
// will handle timezone changes for us (starting from Android O).
// See https://android.googlesource.com/platform/bionic/+/ea87716696bf635706b6f3fa56b8a145add83aff
```

JesseRiemens:
Done! I've added the comment in `os_android.cc`.

Just tested on my M1 mini, it doesn't seem to have this issue.

`os_android.cc` already has the following done:

```c++
void OS::Init() {
  // Calling tzset() is only necessary in Android API version 25 or earlier.
  if (android_get_device_api_level() < 26) {
    // In API version 25, calling tzset() results in a ~0.5% increase in
    // Flutter startup latency. In API version 31, calling tzset() results in
    // a >25% increase in startup latency.
    tzset();
  }
}
```

So we accept the tradeoff that for older Android versions, we only initialize the timezone on startup and don't update it runtime so we don't have a slower `DateTime.now()` call on newer platforms?

mraleph:
Could you apply the following patch on top of your change:

```
diff --git a/build/sanitizers/tsan_suppressions.cc b/build/sanitizers/tsan_suppressions.cc
index 9a3fb8764dd..62634cf9262 100644
--- a/build/sanitizers/tsan_suppressions.cc
+++ b/build/sanitizers/tsan_suppressions.cc
@@ -260,6 +260,10 @@ char kTSanDefaultSuppressions[] =
 // False positive in libc's tzset_internal, http://crbug.com/379738.
 "race:tzset_internal\n"

+// Suppression above does not work if we don't properly symbolize
+// tzset_internal frame. Suppress the closest caller.
+"race:dart::DN_HelperDateTime_timeZoneOffsetInSeconds\n"
+
 // http://crbug.com/380554
 "deadlock:g_type_add_interface_static\n"
```

This should fix TSAN breakage (which is a false positive).

JesseRiemens:
Done!

JesseRiemens:
It seems this doesn't satisfy LUCI, anything else we could try?

mraleph:
I am working on TSAN stuff here: https://dart-review.googlesource.com/c/sdk/+/445941, turns out all this code was just dead. I will ping here once I figure out the best shape of it. Sorry for the trouble :)

mraleph:
I have sent https://dart-review.googlesource.com/c/sdk/+/445941 for review. Once that lands I will ping you here and you should be able to rebase your change (dropping changes to build/sanitizers/tsan_suppressions.cc which I have included into my CL). After that we should be able to land the change. 

I also noticed that presubmit checking is not happy because the code is not formatted. You should run `git cl format` to reformat the code before uploading (this requires `depot_tools` to be in the path).

mraleph:
Thank you for the contribution. The patch has landed!

PR #61466, opened 2025-09-09
`flutter_style_todos`: allow leading space, add and edit tests

ghost:
allow todo comments to begin with or without arbitrary leading spaces from the comment marker. this makes it more independent of general code formatting guidelines used by an end-developer's project eg i use no spaces after `//` which is currently not supported. this also makes it more consistent with other lints such as those for ignore comments eg `//ignore:lint_rule` where no space is recognized by related lints

add related tests

also update TODO for `test_badPatterns` to clarify why one pattern is not linted and what to do next


ghost:
cc @bwilkerson @srawlins

PR #61917, opened 2025-11-06
Add additional definitions to JSSymbol

nex3:

---

<details>

</details>


srujzs:
You may need to resolve conflicts before gerrit picks up the latest commit as a patch that we can CQ+1 on.

nex3:
Working on it, but unfortunately that means I need to do a bunch of recompiling so I can regenerate accurate IR tests. (Do these things really get a bunch of opaque numbers updated every time someone touches the JS interop library?)

nex3:
Merge conflicts resolved.

srujzs:
> (Do these things really get a bunch of opaque numbers updated every time someone touches the JS interop library?)

dart2wasm emits JS methods that then makes the desired interop call for each *unique* interop method. To minimize code size, we just use a monotonically increasing number. I'm guessing they all increased by 18 because we added 18 interop methods (although I count only 17?).

What is surprising though is that we include all the SDK interop members even if they don't get used.

----

We add generated interop procedures into the `Library` that declares the interop member here: https://github.com/dart-lang/sdk/blob/7bb5e4c19246a8b15f8083f15cafb282b11d24f4/pkg/dart2wasm/lib/js/method_collector.dart#L34.

When we're emitting the JS module, we iterate over all the `Library`s in the program that have import or export pragmas: https://github.com/dart-lang/sdk/blob/7bb5e4c19246a8b15f8083f15cafb282b11d24f4/pkg/dart2wasm/lib/functions.dart#L44.

Importantly, in this later step, interop procedures that were not called in user libraries are no longer contained in the `Library` node. I have to assume TFA removed these procedures, but it isn't clear why it doesn't remove unused members in the SDK. @mkustermann, is this WAI? Does TFA not tree-shake SDK members?

mkustermann:
> I have to assume TFA removed these procedures, but it isn't clear why it doesn't remove unused members in the SDK. @mkustermann, is this WAI? Does TFA not tree-shake SDK members?

The reason for those changes to the IR tests is that
* We [run](https://github.com/dart-lang/sdk/blob/9cab17f37cc9fd2976f80d807413ddfdd54a2b61/pkg/dart2wasm/lib/compile.dart#L442) the JS interop transformer before TFA
* It [uses](https://github.com/dart-lang/sdk/blob/9cab17f37cc9fd2976f80d807413ddfdd54a2b61/pkg/dart2wasm/lib/js/method_collector.dart#L24) simple increasing number naming scheme
* TFA tree shakes unused members
* We are left with `_<number>` JS imports that have holes in them

An easy way to fix this would be to rename those static methods after running TFA.

srujzs:
Right, we generate `Procedure`s (with pragmas) and their corresponding `_<number>` JS imports for every interop method, used or unused. We add the `Procedure` to the interop member's declaring library. This generation and addition happens during the JS interop transformations and the imports are stored as strings to be included in the module later.

But later, after TFA, we take the set of JS imports/strings, and only include them in the output JS module if the corresponding `Procedure` (with the associated pragma) is marked as "used":

https://github.com/dart-lang/sdk/blob/0e9676ea1ff0e02a76c2cbb7ff05bee0f5588a0d/pkg/dart2wasm/lib/js/runtime_generator.dart#L59

The set of "used" `Procedure`s come from here:

https://github.com/dart-lang/sdk/blob/0e9676ea1ff0e02a76c2cbb7ff05bee0f5588a0d/pkg/dart2wasm/lib/functions.dart#L348

and they're marked as used here:

https://github.com/dart-lang/sdk/blob/0e9676ea1ff0e02a76c2cbb7ff05bee0f5588a0d/pkg/dart2wasm/lib/functions.dart#L62

However, they don't get marked if they don't exist in the post-TFA AST:

https://github.com/dart-lang/sdk/blob/0e9676ea1ff0e02a76c2cbb7ff05bee0f5588a0d/pkg/dart2wasm/lib/functions.dart#L44

I see the `Procedure`s we generated for unused user interop methods no longer in this post-TFA AST, but I still see the `Procedure`s for unused SDK interop methods in the post-TFA AST. I'm curious as to why why those `Procedure`s still exist in the AST.

It doesn't look like this is specific to the test suite as a local program has the same behavior. There aren't that many members in `dart:js_interop` for this to add considerable code size to the JS module, but it does mean every JS interop member we add to that library today is expanding the JS module, regardless of whether the member is used.

mkustermann:
Firstly, this discussion was triggered due to https://github.com/dart-lang/sdk/commit/9947e20c364ae5cbca1b2afcf6eb379f1fdef4c8  which e.g. modified the IR test for `pkg/dart2wasm/test/ir_tests/interop.bool.dart` and the question came up why tests needed updating, the reason being https://github.com/dart-lang/sdk/pull/61917#issuecomment-3709844352. There don't seem to be any changes to the wasm/mjs files except for the number changes. Also the kernel AST after TFA seems to be the same modulo different numbers.

> I see the Procedures we generated for unused user interop methods no longer in this post-TFA AST, but I still see the Procedures for unused SDK interop methods in the post-TFA AST. I'm curious as to why why those Procedures still exist in the AST.

I assume this is unrelated to https://github.com/dart-lang/sdk/commit/9947e20c364ae5cbca1b2afcf6eb379f1fdef4c8?

Which unused SDK interop methods you see in post-TFA AST? Could you describe what you compile and what you see post-TFA that you don't expect?

srujzs:
Okay, so I tried to reproduce what I saw (where unused SDK interop members are in the `.mjs` file) when I first posted that comment....and I can't reproduce it. 🙃 I tried rebasing to when I posted that comment and still nothing. So I'm not sure what factors led me to see that, but anyways, nothing to do. Sorry for the noise!

Re: Natalie's CL, the increment differences are inevitable unless we do something smarter with the JS string naming scheme because the JS strings are determined before tree-shaking. I guess we can hash the name of the member, but then there are possible collisions to worry about. So nothing to do there either unless updating expectation files gets particularly annoying. My other comments stemmed from investigating the test changes.

PR #62073, opened 2025-11-23
Add tests covering more cases for Uri.resolve

stof:
As part of the scssphp project (which is a PHP port of dart-sass), I had to implement the `Uri.resolve` API of Dart in PHP (as this API does not implement only the RFC3986 resolution but more cases). My initial work used a testsuite ported from the Dart SDK tests but collecting code coverage on my PHP code revealed that those tests were far from covering all non-RFC3986 cases. It missed several cases:

- schemeless base URIs (for which the resolution is derived from the RFC3986 behavior)
- base URIs with a scheme but not authority
- base URIs with an empty path
- base URIs being an absolute path
- base URIs with only a query string
- base URIs being a relative path, resolving a reference that is _not_ a relative path

This is adding tests covering all those cases.

---

<details>

</details>

### DIFF
--- a/tests/corelib/uri_test.dart
+++ b/tests/corelib/uri_test.dart
@@ -131,12 +131,47 @@ void testUriPerRFCs() {
   // Additional tests (not from RFC 3986).
   testResolve("http://a/b/g;p/h;s", "../g;p/h;s");
 
+  base = Uri.parse("http://a?q");
+  testResolve("http://a?p", "?p");
+  testResolve("http://a/b/c", "b/c");
+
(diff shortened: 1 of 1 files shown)


lrhn:
ACK. The more supported cases are resolution against a base that is not a full URI, just a URI reference.

Some choices were made to support those, like allowing a resolution against a relative path to end up with leading `../`.

stof:
Note that I'm not doing any choice there. Those are tests covering the _existing_ behavior of the implementation.

PR #62191, opened 2025-12-07
Fix parameter_assignments documentation for null safety and updated dart syntax

### LINKED ISSUE #7048: The documentation of the lint `parameter_assignments` seems to need updates
### Page URL

https://dart.dev/tools/linter-rules/parameter_assignments

### Page source

_No response_

### Describe the problem

The documentation at https://dart.dev/tools/linter-rules/parameter_assignments seems to have several elements that weren't updated when null safety was introduced.

For example, `badFunction` contains a null-aware assignment to the parameter `optional`, but the type of this parameter is non-nullable (which will give rise to a warning on that assignment). Also, this assignment will never have an effect because `optional` is never null, which also detracts from the value of the example. Also, the default value is specified using `:` which has been replaced by `=` already several versions ago.

Similar issues exist with other examples on this page.

### Expected fix

The code should be updated to take null safety into account such that each good example has no diagnostics and each bad example has just the diagnostic which is being documented on this page.

### Additional context

_No response_

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

IldySilva:
Fix [7048](https://github.com/dart-lang/site-www/issues/7048)

- Fixed incorrect use of ??= on non-nullable parameters
- Converted affected parameters to nullable where appropriate (int → int?)
- Ensured each BAD example triggers exactly one diagnostic (the parameter_assignments lint)
- Ensured each GOOD example produces zero diagnostics
- Removed logically dead code caused by non-nullable ??= usage

### DIFF
--- a/pkg/linter/messages.yaml
+++ b/pkg/linter/messages.yaml
@@ -8261,6 +8261,7 @@ LinterLintCode:
     hasPublishedDocs: false
     documentation: |-
       **DON'T** assign new values to parameters of methods or functions.
+
       Assigning new values to parameters is generally a bad practice, even when using
       operators such as `??=`. Arbitrarily reassigning parameters is usually a
       mistake and can make code harder to reason about.
--- a/pkg/linter/messages.yaml
+++ b/pkg/linter/messages.yaml
@@ -8259,7 +8259,7 @@ LinterLintCode:
       stable: "2.0"
     categories: [style]
     hasPublishedDocs: false
-    documentation: |-
+    deprecatedDetails: |-
       **DON'T** assign new values to parameters of methods or functions.
 
       Assigning new values to parameters is generally a bad practice, even when using
--- a/pkg/linter/messages.yaml
+++ b/pkg/linter/messages.yaml
@@ -8259,12 +8259,11 @@ LinterLintCode:
       stable: "2.0"
     categories: [style]
     hasPublishedDocs: false
-    deprecatedDetails: |-
+    documentation: |-
       **DON'T** assign new values to parameters of methods or functions.
-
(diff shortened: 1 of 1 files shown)


IldySilva:
cc @parlough

pkg/linter/messages.yaml
```diff
@@ -8259,12 +8259,11 @@ LinterLintCode:
       stable: "2.0"
     categories: [style]
     hasPublishedDocs: false
-    deprecatedDetails: |-
+    documentation: |-
```
parlough: This one should still be `deprecatedDetails`. `documentation` is reserved for a format consistent with the analyzer's diagnostic docs.

pkg/linter/messages.yaml
```diff
@@ -8259,12 +8259,11 @@ LinterLintCode:
...
       **DON'T** assign new values to parameters of methods or functions.
-
-      Assigning new values to parameters is generally a bad practice unless an
-      operator such as `??=` is used.  Otherwise, arbitrarily reassigning parameters
-      is usually a mistake.
+      Assigning new values to parameters is generally a bad practice, even when using
```
parlough:
```suggestion

      Assigning new values to parameters is generally a bad practice, even when using
```

parlough (commented):
Thanks for updating this here!

Do note the rest of the review will be in https://dart-review.googlesource.com/c/sdk/+/466600

mraleph:
@IldySilva are you still working on this? If you are please address comments - otherwise please close PR. Thanks.

IldySilva:
> @IldySilva are you still working on this? If you are please address comments - otherwise please close PR. Thanks.

@mraleph  i'll fix  it

PR #62286, opened 2025-12-22
[analyzer] Improve 'not a type' error message

### LINKED ISSUE #48603: Improve error message for typo in setter name
Code like this:

```dart
class A {}

class B {
  A _a;
  A get a => _a;
  set A(A a) => _a = a;
}
```
gives error:
```
Error: 'A' isn't a type.
  A _a;
```
which is a little confusing, as the type is clearly defined above. It took me a long while to find out what is the actual issue (I first thought incorrect import / conflicting names imported).

It also causes analyzer to highlight all the places except where the actual problem is:
8pTMspp9ZPUtgWb

nathannewyen:
Improves the error message when a non-type is used where a type is expected.

**Before:** `'A' isn't a type.`
**After:** `'A' is a setter, but a type was expected.`

The element kind is now included (setter, getter, function, method, variable, etc.) to help users understand what the identifier refers to.

Fixes #48603

### DIFF
--- a/pkg/analyzer/lib/src/dart/resolver/named_type_resolver.dart
+++ b/pkg/analyzer/lib/src/dart/resolver/named_type_resolver.dart
@@ -396,6 +396,9 @@ class NamedTypeResolver with ScopeHelpers {
             importPrefixElement.getMethod(name) ??
             importPrefixElement.getSetter(name);
       }
+      var elementKind = element != null
+          ? _ErrorHelper._getElementKindDescription(element)
+          : 'not a type';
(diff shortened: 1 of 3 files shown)


mraleph:
@nathannewyen Are you still working on this? 

If you do please review comments in Gerrit and note that the code does not build and has merge conflicts - so it is not possible to test it. 

If you don't working on this anymore please close PR. 

Thanks.

mraleph:
Seems stale. Ping us if you decide to return to this.

PR #62359, opened 2026-01-07
Update documentation links in directory.dart

gisborne:
Updated documentation links for file handling resources.

The page previously linked to no longer exists. This is a reasonable alternative.


mraleph:
I think the link has to be retargeted to: https://dart.dev/libraries/dart-io#files-and-directories

mraleph:
Seems stale

PR #62368, opened 2026-01-07
Fix generic type mismatch caused by missing type argument.

SeifShaheen:
This pull request introduces enhancements to the type inference and subtyping logic for extension types and Fixes #62242 , adds new targeted test cases for extension type inference and subtyping. The most important changes are grouped below by theme.

**Type Inference and Subtyping Improvements**

* Updated `TypeConstraintGenerator` in `type_analyzer_operations.dart` to handle subtype constraint generation for extension type declarations, ensuring correct erasure and comparison logic.
* Modified `TypeSchemaEnvironment` in `type_schema_environment.dart` to add a fallback subtype check for extension type variables, allowing legacy/undetermined extension type parameters to be assignable to their non-nullable equivalents when the bound is valid.

**Extension Type Test Coverage**

* Added a new test case `issue_inference.dart` to verify inference and behavior for classes and extensions using extension types as type parameters.
* Added a new test case `test_rep_subtyping.dart` to check subtyping relationships between extension types and their representation types.


mraleph:
@SeifShaheen Thank you for your contribution, unfortunately it is fixing things in the wrong way - an engineer on the front-end team will submit the proper fix instead. Type system and inference rules are rather complex so next time consider choosing a simpler target as your first contribution.

Issue #62384, opened 2026-01-09
[spec_parser] Metadata/record type ambiguity when whitespace separates annotation from parentheses


modulovalue:
The ANTLR grammar in `tools/spec_parser` does not disambiguate between metadata with arguments and metadata followed by a record type based on whitespace.

Currently, both of these parse identically:

```dart
@Foo() f() {}   // metadata with empty arguments, function with no return type
@Foo () f() {}  // should be: metadata without arguments, function with record return type ()
```

The second case should parse `@Foo` as metadata without arguments, and `()` as an empty record return type for function `f`. Instead, the parser greedily consumes `()` as metadata arguments regardless of whitespace.

## Expected behavior

Whitespace between the constructor designation and the opening parenthesis should prevent the parentheses from being parsed as metadata arguments:

- `@Foo()`: metadata with arguments
- `@Foo ()`: metadata without arguments, followed by record type

## Proposed solution

Add an `isNoSkip()` semantic predicate to the `metadatum` rule that checks whether there is whitespace between the previous token and the current token by comparing their character positions in the source text.

FMorschel:
Current output:

```dart
class C {
  const C();
}

@C // no_annotation_constructor_arguments 
() foo() => ();

@C ()
// ^ expected_executable
void bar() {}
```

CC @lrhn

lrhn:
SGTM, @eernstg

PR #62385, opened 2026-01-09
[spec_parser] Add isNoSkip() to disambiguate metadata from record types

modulovalue:
This change adds a semantic predicate to the metadatum rule that checks whether there is whitespace between the constructor designation and the opening parenthesis of arguments.

Without this fix:
- "@Foo () f() {}" parses "()" as metadata arguments (incorrect)

With this fix:
- "@Foo() f() {}" parses "()" as metadata arguments (no whitespace)
- "@Foo () f() {}" parses "@Foo" as metadata, "()" as record return type

The isNoSkip() function compares character positions to detect if any tokens (whitespace, comments) were skipped between the previous and current visible tokens.

Bug: https://github.com/dart-lang/sdk/issues/62384

### DIFF
--- a/tools/spec_parser/dart_spec_parser/Dart.g4
+++ b/tools/spec_parser/dart_spec_parser/Dart.g4
@@ -4,6 +4,13 @@
 
 // CHANGES:
 //
+// v0.60 Disambiguate metadata arguments from record types by speculatively
+// parsing constructorDesignation and only taking the arguments alternative
+// when '(' follows immediately with no skipped tokens.
+// Without this, "@Foo () f() {}" would parse "()" as metadata arguments.
+// With this, whitespace between the identifier and "(" causes the parser to
+// treat "@Foo" as metadata without arguments, and "()" as a record return type.
+//
 // v0.59 Gather some occurrences of `AUGMENT` in a single location. Rename
 // `topLevelDefinition` to `topLevelDeclaration` (as in the specification).
 //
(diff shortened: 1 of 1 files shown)


mraleph:
Is this something we want to land @eernstg ?

eernstg:
Yes, I just haven't gotten around to it yet.

mraleph:
@modulovalue @eernstg friendly ping, I am retriaging PRs and noticed that this one is fairly stale.

eernstg:
Thanks for the heads up, @mraleph! I'm afraid there is no known solution to this issue, though.

The treatment of whitespace probably cannot be expressed in the grammar unless we re-shape the entire grammar to know about whitespace everywhere, which would be a massive change. This is not an attractive option because it surely causes the grammar to be less readable, and more different from the grammar in the language specification documents, and the main purpose of the spec parser is that it bridges the gap between the specification documents and the actual code.

ANTLRv4 doesn't support syntactic predicates, so we can't use the straightforward "customized look-ahead" technique that we used with ANTLRv3.

The ANTLRv4 parser generator does support semantic predicates, but they cannot be used to guide the parsing process: the only outcomes of evaluating a semantic predicate (apart from side effects, if any) is (1) the evaluation yields true, and parsing proceeds as if the predicate had not been there, or (2) the evaluation yields false, in which case the parsing fails and a syntax error is reported. So we probably can't use a semantic predicate to distinguish the two syntactic situations of interest here.

(Exceptions: A semantic predicate can be used to guide the recognition done by _lexical rules_, but that doesn't suffice here. Also, I think semantic predicates can guide parsing when they occur at the front of the alternative, but not when they occur anywhere after the first non-terminal, and I don't think that'll suffice for this purpose, either. I did try that.)

eernstg:
@modulovalue, do you see a way to do it?

modulovalue:
@eernstg I did suggest an alternative and the patch should be live in the CL. I might've forgotten to check a box as it appears you have not received a notification.

> ANTLRv4 doesn't support syntactic predicates, so we can't use the straightforward "customized look-ahead" technique that we used with ANTLRv3.

You mentioned it briefly, but I found the following during my investigation, see https://github.com/antlr/antlr4/blob/master/doc/predicates.md#predicates-in-lexer-rules:

> In parser rules, predicates must appear on the left edge of alternatives to aid in alternative prediction.

Predicates appear to be supported, but since ANTLR is LL based, predicates need to be at the start of a production. With this behavior be can do a "predictive parse" to implement the ambiguity resolution strategy in the ANTLR grammar.

Let me know what you think!

modulovalue:
This disambiguation technique is enough for ANTLR, but it is not enough to explain how to fully disambiguate the metadata ambiguity, as it relies on ANTLR-like semantics.

This PR handles only one part, but there are two. In addition to forbidding spaces before arguments in metadata, the non-argument cases need to be restricted too.

A non-argument metadata case that follows a record can't have a space between the metadata and left parenthesis.

In other words, in the non-skip case, one of the ambiguous trees needs to be rejected.

```dart
void a() { @A() foo() {} }
```

In the skip case, the other tree needs to be rejected (i.e. the one that was not rejected in the non-skip case).

```dart
void a() { @A () foo() {} }
```

Or, to make it clearer, consider a 2x2 matrix:

```          
                            WITH SKIP              NO SKIP
PARSE AS RECORD TYPE           1                      2
PARSE AS ARG LIST              3                      4

```

1 is expected 4 is expected. The ambiguity comes from 2 and 3. This PR fixes 3. 2 remains.

So, since the rule implemented here is incomplete, I think it's best to close this.

If we extend changes like these to their logical conclusion, since ANTLR offers no help in maintaining correctness with such disambiguations, we would be rewriting the disambiguation rules in a turing complete language in ANTLR, which defeats the purpose of a specification. We already have a parser written in a turing complete language, the official Dart parser!

PR #62400, opened 2026-01-14
[infra] Honor `download_reclient=False`

ntkme:
As a non-Googler, I set `download_reclient=False` to skip downloading `reclient`.

I noticed that this option only skips downloading `/buildtools/reclient`, but not `/buildtools/reclient-win` or `/buildtools/reclient-linux`.

This PR ensures that it skips downloading all kinds of `reclient` if `download_reclient=False`.

---

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -496,7 +496,7 @@ deps = {
         'version': Var('reclient_version'),
       }
     ],
-    'condition': 'host_os == "win"',
+    'condition': 'download_reclient and host_os == "win"',
     'dep_type': 'cipd',
   },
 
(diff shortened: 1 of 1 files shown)


ntkme:
>  Download reclient only on the platforms where it has packages available. Unfortunately windows-arm64 gclient uses x64 python which lies in host_cpu, so we have to use a variable to not download reclient there.

Based on the original comment around the usage of `download_reclient`, it seems like the original intention of the variable  is to have a way to prevent double downloading `/buildtools/reclient` + `/buildtools/reclient-win` on windows-arm64 because platform detection is broken on the platform.

I don't think downloading two copies of reclient is really a huge issue. However, for non-Googlers reclient is not useable, I'd rather be able to not download it at all.

PR #62415, opened 2026-01-15
Fix VM service smaps path NUL termination to prevent OOB reads

rootvector2:
Ensure NUL termination when copying mapping paths from /proc/self/smaps.
Prevent potential out‑of‑bounds reads and info disclosure in VM service RSS mappings.


mraleph:
Thanks. CL is in the CQ.

PR #62571, opened 2026-02-02
JSON and UTF-8 fast paths

FelixMittermeier:
## Summary
This change set adds native fast paths for JSON parsing (one-byte and two-byte strings) and optimizes UTF-8 decoding for large inputs. It reduces overhead in hot loops by moving parsing and map construction into C++ and by using faster scanning primitives for strings and ASCII runs.

## Motivation
- The JSON benchmarks and production inputs include non-Latin1 characters, which meant the existing one-byte fast path was bypassed for two-byte strings.
- The Dart JSON parser spends a lot of time scanning strings and building maps/arrays in Dart, even when no reviver is used.
- UTF-8 decode for large inputs was dominated by per-codepoint decoding and validation work that can be streamlined with tight loops and ASCII run detection.

### Dart-side changes
- `sdk/lib/_internal/vm/lib/convert_patch.dart`
  - Adds a fast JSON parse path for large inputs (length >= 128) with no reviver, calling native parsers for one-byte or two-byte strings. Falls back to the existing Dart parser on failure.
  - Adds a native string terminator scanner for one-byte JSON strings to reduce overhead in `parseString` / `parseStringToBuffer`.
  - Adds a small-int string cache (0..999) for short ASCII numeric strings to avoid repeated allocations.
  - Routes large UTF-8 decode operations through optimized native decoders when allowed (`decodeUtf8ToOneByteString` / `decodeUtf8ToTwoByteString`).
- `sdk/lib/_internal/vm_shared/lib/compact_hash.dart`
  - Uses a native map constructor for alternating key/value pairs to reduce Dart map building cost.
- `sdk/lib/internal/internal.dart` and `sdk/lib/_internal/vm/lib/internal_patch.dart`
  - Expose the native map constructor and allow the VM to see the `K,V` type arguments.

### VM/runtime changes
- `runtime/lib/string.cc`
  - Adds native JSON parsers for one-byte and two-byte strings (`Json_parseOneByteFast`, `Json_parseTwoByteFast`).
  - Introduces fast string terminator scanning for JSON strings (one-byte and two-byte), with SIMD-friendly scanning and no-safepoint sections.
  - Adds native map and array builders to avoid Dart-side allocation overhead, including small-key caches and shape caches for common object shapes.
  - Preserves map type arguments in the native fast map builder.
  - Fixes validation in the two-byte number parser to reject invalid forms like "--1".
- `runtime/platform/unicode.cc`
  - Adds ASCII run detection and specialized loops for Latin1 and UTF-16 decoding to reduce per-byte overhead.
  - Tightens validation in the fast decode paths (overlongs, surrogates, range checks).
- `runtime/vm/object.h` and `runtime/vm/object.cc`
  - Adds `TwoByteString::SubStringUnchecked` for validated fast substring creation.
- `runtime/vm/bootstrap_natives.h`
  - Wires new native entries used by the Dart fast paths.

## Why this is faster
- Native parsers avoid Dart object overhead in tight loops and eliminate repeated bounds checks.
- SIMD/word-at-a-time string terminator scans drastically reduce the time spent scanning JSON strings.
- Native map/array builders avoid repeated allocations and rehashing during parsing, and cache common shapes.
- UTF-8 decoding now handles long ASCII runs and well-formed multi-byte sequences with tighter loops.

## Correctness safeguards
- Fast JSON path is only used when `reviver == null`; failures fall back to the Dart parser.
- Two-byte number parsing rejects invalid inputs such as "--1".
- Native map builder retains correct `K,V` type arguments.
- UTF-8 decoding fast paths still enforce overlong, surrogate, and range checks.

## Performance (example)
On macOS ARM64 Release using `benchmarks/IsolateJson/dart/IsolateJson.dart -n 10`:
- `SyncDecode1MBx1`: ~64 ms -> ~24–25 ms (about 60% faster)
- `Decode1MBx1`: ~90 ms -> ~29–32 ms (about 65% faster)
- `Decode1MBx4`: ~177 ms -> ~53–68 ms (about 60–70% faster)

## Testing
- `python3 tools/build.py --mode=release --arch=arm64 dartvm`
- `xcodebuild/ReleaseARM64/dartvm benchmarks/IsolateJson/dart/IsolateJson.dart -n 10`

### DIFF
--- a/runtime/lib/string.cc
+++ b/runtime/lib/string.cc
@@ -14,7 +14,23 @@
 #include "vm/object_store.h"
 #include "vm/symbols.h"
 
+#if defined(TARGET_ARCH_ARM64)
+#include <arm_neon.h>
+#endif
+#include <cstdint>
+
 namespace dart {
+
+class NativeStringDataAccess {
+ public:
+  static uint8_t* OneByteDataStart(const String& str) {
+    return OneByteString::DataStart(str);
+  }
+
+  static uint16_t* TwoByteDataStart(const String& str) {
+    return TwoByteString::DataStart(str);
+  }
+};
 
 DEFINE_NATIVE_ENTRY(String_fromEnvironment, 0, 3) {
   GET_NON_NULL_NATIVE_ARGUMENT(String, name, arguments->NativeArgAt(1));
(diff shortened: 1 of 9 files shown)


mraleph:
@FelixMittermeier Thanks for the contribution. There are two different improvements here one for UTF8 decoding and one about JSON decoding. These should have been split into two different PRs (but lets discuss this first to understand what we actually would want to do here).

First a couple of questions:

1. How much of this is handwritten? 
2. What's the motivation?

Fundamentally our approach in this area is two-pronged:

- We want to make sure you can write high-performance code in Dart. Instead of shifting this work-load into C++ the preferred strategy is to make sure that Dart does job reasonably well itself. 
- We want to make sure that you can fallback to native code as necessary, without bundling this native code into Dart runtime (see native assets)

So in general we favor improvements in Dart compiler and core libraries over moving parts of the code to C++. If anything we prefer to move C++ code into Dart.

For UTF8 decoding are decoding algorithms different? If they are different - how? If they are not different, do you understand where performance difference comes from?

For JSON decoding, I am fairly certain that we do not want adding 2KLOC of C++ code which duplicates an already existing Dart code. 

Going back to the JSON decoding. Try the following patch: 

```dart
diff --git a/benchmarks/IsolateJson/dart/IsolateJson.dart b/benchmarks/IsolateJson/dart/IsolateJson.dart
index ccf80e5a0b3..134fe0cebca 100644
--- a/benchmarks/IsolateJson/dart/IsolateJson.dart
+++ b/benchmarks/IsolateJson/dart/IsolateJson.dart
@@ -81,7 +81,9 @@ Future<Map> decodeJson(bool useSendAndExit, Uint8List encodedJson) async {
 }
 
 Future<void> jsonDecodingIsolate(JsonDecodeRequest request) async {
-  final result = json.decode(utf8.decode(request.encodedJson));
+  final result = kUseFused
+      ? decoder.convert(request.encodedJson)
+      : json.decode(utf8.decode(request.encodedJson));
   if (request.useSendAndExit) {
     Isolate.exit(request.sendPort, result);
   } else {
@@ -89,6 +91,10 @@ Future<void> jsonDecodingIsolate(JsonDecodeRequest request) async {
   }
 }
 
+const kUseFused = const bool.fromEnvironment('use.fused', defaultValue: false);
+
+final decoder = const Utf8Decoder().fuse(const JsonDecoder());
+
 class SyncJsonDecodingBenchmark extends BenchmarkBase {
   SyncJsonDecodingBenchmark(
     String name, {
@@ -100,7 +106,9 @@ class SyncJsonDecodingBenchmark extends BenchmarkBase {
   void run() {
     int l = 0;
     for (int i = 0; i < iterations; i++) {
-      final Map map = json.decode(utf8.decode(sample));
+      final Map map = kUseFused
+          ? decoder.convert(sample)
+          : json.decode(utf8.decode(sample));
       l += map.length;
     }
     assert(l > 0);
```

And the benchmark with `-Duse.fused=true`. This decodes the same input but removes unnecessary UTF8->String decoding. It will be considerably faster then decoding from strings.

mraleph:
Also please see https://dart-review.googlesource.com/c/sdk/+/477540 for failing CQ checks.

PR #62630, opened 2026-02-10
[cfe][analyzer] Add record spreading (record-spreads experiment)

MelbourneDeveloper:
This PR implements record spreading as proposed in [dart-lang/language#2128](https://github.com/dart-lang/language/issues/2128). It adds `...recordExpression` syntax inside record literals, gated behind the `record-spreads` experiment flag.

### The Problem

Records are great for lightweight data composition, but there's no way to build on an existing record without manually destructuring it. If you have a `(int, int)` and want to add a named field, you're stuck doing this:

```dart
var point = (1, 2);
// Tedious: manually extract and re-assemble
var colorPoint = (point.$1, point.$2, color: 'red');
```

This gets worse as record shapes grow. @lrhn's proposal solves it cleanly:

```dart
var colorPoint = (...point, color: 'red');
```

### What This Implements

The spread is a **purely static operation**. During type inference, `...record` is desugared into individual field accesses (`record.$1`, `record.$2`, `record.name`, etc.). By the time any backend sees the code, it's a normal `RecordLiteral`. No VM, dart2js, DDC, or dart2wasm changes were needed.

The core desugaring lives in `_expandRecordSpreads()` in `inference_visitor.dart`:

```
Source: (...point, color: 'red')
  → Parser: RecordSpreadElement(point)
  → Type inference: point has type (int, int), so expand to tmp.$1, tmp.$2
  → Output Kernel IR: RecordLiteral(tmp.$1, tmp.$2, color: 'red')
```

This directly follows @lrhn's original framing: *"it can be desugared into `(point[0], point[1], color: Color.red)` if we have a syntax for directly accessing members of a record."*

### What the Issue Requires (and How We Handle It)

**Concrete record type required** — @lrhn specified that the spread expression must have a *known* record type, not `dynamic`, `Record`, or `T extends Record`. We enforce this: spreading any of those is a compile-time error.

**No dynamic record operations** — @lrhn explicitly rejected `Record concat(Record a, Record b) => (...a, ...b)` because implementation strategies require all record types to be known at compile time. We correctly reject this.

**No `Destructure_N_` interfaces** — @munificent initially floated these for positional destructuring, but @lrhn pushed back and @munificent wasn't attached. We correctly omit them.

**No `if`/`for` in records** — @lrhn rejected conditionally changing record structure. Not implemented.

**Null-aware `...?` not supported** — Record field shapes must be statically known, so you can't conditionally include fields from a nullable record. We emit an error. Users can handle nullable records explicitly with `?? (defaultValues)`.

**`$N` positional name clash** — A named field like `$1` from a spread would clash with positional getter numbering. We detect and report this.

**Argument list spreading** — @munificent *emphatically* supported `foo(...record)`. This is deferred to a follow-up CL. The parser and AST infrastructure added here will support it.

### Changes by Layer

| Layer | Key Files |
|-------|-----------|
| Experiment flag | `tools/experimental_features.yaml` + generated flag files |
| Parser | `parser_impl.dart` — detects `...` in record context, calls `handleRecordSpreadField` |
| CFE body builder | `body_builder.dart` — creates `RecordSpreadElement`, gated behind experiment flag |
| CFE type inference | `inference_visitor.dart` — `_expandRecordSpreads()` validates types, hoists to temp, expands fields |
| CFE const eval | `constant_evaluator.dart` — enables `RecordIndexGet`/`RecordNameGet` for const contexts |
| CFE diagnostics | `messages.yaml` — 4 error messages |
| Analyzer AST | `ast.dart` — `RecordSpreadField` node |
| Analyzer resolver | `record_literal_resolver.dart` — spread field resolution and type expansion |
| Analyzer diagnostics | `messages.yaml` — 3 error messages |

### Test Coverage

Six test files in `tests/language/record_spreads/`, 16 test cases total:

| Test | What It Covers |
|------|---------------|
| `record_spread_basic_test` | Positional, named, mixed, multiple spreads, nested records |
| `record_spread_const_test` | Const spreading, const identity preservation |
| `record_spread_error_test` | 10 error cases: non-record types, `dynamic`, `Record`, generics, `...?`, duplicate names, `$N` clash |
| `record_spread_evaluation_order_test` | Spread evaluated exactly once, left-to-right ordering |
| `record_spread_inference_test` | Static types, downward inference, subtyping |
| `record_spread_mixed_syntax_test` | Collections, pattern matching, closures, async/await, switch, try/catch, type tests |

### Not in Scope

- **Argument list spreading** (`foo(...record)`) — follow-up CL
- **Formatter support** — `dart_style` is a separate package

Because this feature is behind a feature flag, it can remain disabled by default until these other features accompany it.

Original issue: https://github.com/dart-lang/language/issues/2128

---

<details>

</details>


leafpetersen:
Thank you for your contribution, it's great to see people taking an interest in this area.  Because of the nature of a programming language, it's very hard to walk back changes once we have shipped something, and hence we require a fairly substantial amount of design work and discussion on a feature before we accept it.  There is also substantial coordination involved in getting a feature ready for shipping, making space for it in our schedule, and moving it through the pipeline.  We have an established process for making all of this happen, documented [here](https://github.com/dart-lang/language/blob/main/doc/life_of_a_language_feature.md).  The documentation is somewhat more targeted to the common case where Dart team members are doing the design and implementation work, but I think generalizes easily enough to external contributions like this one.  If you are interested in pushing this feature forward, the first step would be to work towards getting a feature specification in place as documented [here](https://github.com/dart-lang/language/blob/main/doc/life_of_a_language_feature.md#design-feedback-and-iteration), either by starting with discussion in issues, or by opening a PR with a proposed specification in that repository.  Before doing the latter, please take some time to look over previous accepted feature specifications in the `accepted` directory to understand what needs to be covered by a feature spec.  In particular, a non-exhaustive list of things to cover here would include new syntax, type checking, type inference, errors and warnings, and dynamic semantics.   If and when we have an accepted feature spec and are able to allocate resources towards shipping it, we can discuss how and when implementation work might proceed.  The linked document above gives a good overview of all of the pieces that are involved in this.  In the meantime, I would suggest closing this out, since we cannot accept this in its current form and would at a minimum need to see it broken out into more granular and reviewable pieces.

cc @dart-lang/language-team @mraleph @mit-mit

mraleph:
@munificent has updated [Life of a language feature](https://github.com/dart-lang/language/blob/main/doc/life_of_a_language_feature.md) to make the process and our expectations more explicit. 

I am going ahead and close this one, because the language feature itself is not at the stage where we can even consider _looking_ at the implementation. 

That's not to discourage you from trying to contribute a language feature - but to set expectations straight: taking a language feature from 0 to 100 is basically a full time job (more often than not a full time job for _multiple_ engineers) over weeks. Contributing a language feature as a side hustle in the evening is probably next to impossible (especially if you expect a clanker to do heavy lifting for you).

MelbourneDeveloper:
Thanks @leafpetersen and @mraleph 

> If you are interested in pushing this feature forward, the first step would be to work towards getting a feature specification in place as documented [here](https://github.com/dart-lang/language/blob/main/doc/life_of_a_language_feature.md#design-feedback-and-iteration), either by starting with discussion in issues, or by opening a PR with a proposed specification in that repository

Will do. I will start with some documentation and get the conversation going.

PR #62680, opened 2026-02-15
[vm] Allow to skip crash handler installation in bin::Platform::Initialize.

### LINKED ISSUE #62679: Process.start with non-existent command delivers SIGPIPE, killing Flutter apps on macOS
## Overview

On macOS, when `Process.start` is called with a command that doesn't exist, the exec failure cleanup path generates a SIGPIPE signal that kills the process. 

I believe this is because `ExitCodeHandler::ExitCodeHandlerEntry` writes to a pipe whose read end has already been closed by the error handling in `ProcessStarter::Start()`

The bug is easily reproducible in Flutter macOS apps [(issue ticket with simple test code)](https://github.com/flutter/flutter/issues/182436) but the root cause is in the Dart VM's runtime/bin/process_macos.cc. Neither flutter nor Dart SIG_IGN sigpipe. 

I have tried but not been able to actually trigger the bug in Dart directly, but I've traced it with a Mk1 eyeball.

## Reproduction
[This flutter ticket](https://github.com/flutter/flutter/issues/182436) has repro code for it, but do note that its a flutter app, I've not been able to reproduce it in native Dart (see below why I still think its a dart issue)

<details close><summary>Code sample</summary>

```dart
import 'dart:io';
import 'package:flutter/material.dart';

void main() {
  debugPrint('=== About to call Process.start("no-such-command") ===');
  Process.start('no-such-command', []).then((process) {
    debugPrint('=== Process started (unexpected): pid=${process.pid} ===');
  }).catchError((error) {
    debugPrint('=== Process.start error: $error ===');
  });

  runApp(const SigpipeTestApp());
}

class SigpipeTestApp extends StatelessWidget {
  const SigpipeTestApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'SIGPIPE Test',
      home: Scaffold(
        appBar: AppBar(title: const Text('SIGPIPE Test')),
        body: const Center(
          child: Text('App is running. If SIGPIPE hits, this will exit.'),
        ),
      ),
    );
  }
}

```

</details>

## Root cause
In `runtime/bin/process_macos.cc`, when exec fails:

1. `ProcessStarter::Start()` error path closes the read end of the exit code pipe.
2. `ExitCodeHandler::ExitCodeHandlerEntry` reaps the child via wait(), then calls `FDUtils::WriteToBlocking` to write the exit code to the write end of the same pipe.
3. Writing to a pipe with no reader delivers `SIGPIPE` to the process.

The code already handles the resulting EPIPE errno correctly (there's even a comment about it), but it doesn't prevent the SIGPIPE signal from being delivered alongside it.

## Proposed solution
There's a couple of ideas I've had but the cleanest one is probably to mask sigpipe in the thread.

```diff
diff --git a/runtime/bin/process_macos.cc b/runtime/bin/process_macos.cc
index 8aebf479c70..40a2213b291 100644
--- a/runtime/bin/process_macos.cc
+++ b/runtime/bin/process_macos.cc
@@ -209,8 +209,17 @@ class ExitCodeHandler {
         intptr_t exit_code_fd = ProcessInfoList::LookupProcessExitFd(pid);
         if (exit_code_fd != 0) {
           int message[2] = {exit_code, negative};
-          ssize_t result =
-              FDUtils::WriteToBlocking(exit_code_fd, &message, sizeof(message));
+          // Block SIGPIPE on this thread. If the read end of the exit
+          // code pipe has already been closed (exec failed and the
+          // parent discarded the pipe), this write will harmlessly
+          // fail with EPIPE instead of killing the process.
+          sigset_t block_sigpipe, old_mask;
+          sigemptyset(&block_sigpipe);
+          sigaddset(&block_sigpipe, SIGPIPE);
+          pthread_sigmask(SIG_BLOCK, &block_sigpipe, &old_mask);
+          ssize_t result = FDUtils::WriteToBlocking(exit_code_fd, &message,
+                                                    sizeof(message));
+          pthread_sigmask(SIG_SETMASK, &old_mask, nullptr);
           // If the process has been closed, the read end of the exit
           // pipe has been closed. It is therefore not a problem that
           // write fails with a broken pipe error. Other errors should
```

I'm not tied to any particular solution, this is just a nice tidy one. I'm sure there are other ways to fix it.

## Workaround
You can workaround this issue by call signal(SIGPIPE, SIG_IGN) early in the app in swift code (e.g., in the macOS `AppDelegate.applicationWillFinishLaunching`) prevents the crash.

## Justification
Despite this not being able to be reproduced in Dart directly, and needing Flutter (so far), I believe this is a bug with Dart VM because the VM code is writing to a pipe it knows may have no reader. The comments even say so:
> // If the process has been closed, the read end of the exit
>  // pipe has been closed. It is therefore not a problem that
> // write fails with a broken pipe error. Other errors should
> // not happen.

The issue is that the code handles the EPIPE but ignores that it will generate a SIGPIPE

Arbaaz123676:
## Summary
On macOS, writing to a closed exit pipe can deliver SIGPIPE which terminates the process even when EPIPE is handled correctly.

This change temporarily blocks SIGPIPE while writing the exit code in "ExitCodeHandlerEntry". This ensures that failed "Process.start" calls do not trigger unintended process termination.

The signal mask is scoped only around the write operation to keep the behaviour localized and avoid affecting global signal handling.

## Fixes: #62679.

### DIFF
--- a/runtime/bin/platform_linux.cc
+++ b/runtime/bin/platform_linux.cc
@@ -65,7 +65,7 @@ static void segv_handler(int signal, siginfo_t* siginfo, void* context) {
   abort();
 }
 
-bool Platform::Initialize() {
+bool Platform::Initialize(bool install_crash_handler /* = true */) {
   // Turn off the signal handler for SIGPIPE as it causes the process
   // to terminate on writing to a closed pipe. Without the signal
   // handler error EPIPE is set instead.
@@ -87,31 +87,35 @@ bool Platform::Initialize() {
     return false;
   }
 
-  act.sa_flags = SA_SIGINFO;
-  act.sa_sigaction = &segv_handler;
-  if (sigemptyset(&act.sa_mask) != 0) {
-    perror("sigemptyset() failed.");
-    return false;
-  }
-  if (sigaddset(&act.sa_mask, SIGPROF) != 0) {
-    perror("sigaddset() failed");
-    return false;
-  }
-  if (sigaction(SIGSEGV, &act, nullptr) != 0) {
-    perror("sigaction() failed.");
-    return false;
-  }
-  if (sigaction(SIGBUS, &act, nullptr) != 0) {
-    perror("sigaction() failed.");
-    return false;
-  }
-  if (sigaction(SIGTRAP, &act, nullptr) != 0) {
-    perror("sigaction() failed.");
-    return false;
-  }
-  if (sigaction(SIGILL, &act, nullptr) != 0) {
-    perror("sigaction() failed.");
-    return false;
+  if (install_crash_handler) {
+    act = {};
+    act.sa_flags = SA_SIGINFO;
+    act.sa_sigaction = &segv_handler;
+
+    if (sigemptyset(&act.sa_mask) != 0) {
+      perror("sigemptyset() failed.");
+      return false;
+    }
+    if (sigaddset(&act.sa_mask, SIGPROF) != 0) {
+      perror("sigaddset() failed");
+      return false;
+    }
+    if (sigaction(SIGSEGV, &act, nullptr) != 0) {
+      perror("sigaction() failed.");
+      return false;
+    }
+    if (sigaction(SIGBUS, &act, nullptr) != 0) {
+      perror("sigaction() failed.");
+      return false;
+    }
+    if (sigaction(SIGTRAP, &act, nullptr) != 0) {
+      perror("sigaction() failed.");
+      return false;
+    }
+    if (sigaction(SIGILL, &act, nullptr) != 0) {
+      perror("sigaction() failed.");
+      return false;
+    }
   }
   return true;
 }
--- a/runtime/bin/platform_macos.cc
+++ b/runtime/bin/platform_macos.cc
@@ -72,7 +72,7 @@ static void segv_handler(int signal, siginfo_t* siginfo, void* context) {
   abort();
 }
 
-bool Platform::Initialize() {
+bool Platform::Initialize(bool install_crash_handler /*= true */) {
   // Turn off the signal handler for SIGPIPE as it causes the process
   // to terminate on writing to a closed pipe. Without the signal
   // handler error EPIPE is set instead.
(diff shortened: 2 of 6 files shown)


mraleph:
There is already code in `Platform::Initialize` which does this. I would expect it is called by Flutter engine. This means something later messes it up.

Arbaaz123676:
Hi @mraleph, Thanks for the pointer! That makes sense — I added the masking here because I observed SIGPIPE during the exit pipe write path on macOS, but if Platform::Initialize already configures it, then something later must be modifying the mask.

I’ll investigate where SIGPIPE handling may be reset after initialization. Do you have a recommended area to check (Flutter embedder vs VM runtime), or should I go with tracing signal mask changes around process start?

Happy to adjust the approach based on the intended design.

mraleph:
See my comment here: https://github.com/dart-lang/sdk/issues/62679#issuecomment-3907059696

runtime/bin/platform_macos.cc
```diff
@@ -94,31 +94,34 @@ bool Platform::Initialize() {
...
-    return false;
-  }
-  if (sigaction(SIGTRAP, &act, nullptr) != 0) {
-    perror("sigaction() failed.");
-    return false;
-  }
-  if (sigaction(SIGILL, &act, nullptr) != 0) {
-    perror("sigaction() failed.");
-    return false;
+  if (install_crash_handler){
```
mraleph: Not properly formatted: missing space here.

In general, I'd recommend running `git cl format` to make sure that all code is properly formatted.

mraleph (changes requested):

mraleph:
The title of this PR should probably be updated to match what we are doing here.

runtime/bin/platform_macos.cc
```diff
@@ -72,7 +72,7 @@ static void segv_handler(int signal, siginfo_t* siginfo, void* context) {
   abort();
 }
 
-bool Platform::Initialize() {
+bool Platform::Initialize(bool install_crash_handler) {
```
mraleph: 1. We would usually write `bool install_crash_hander /* = true */` to make default value obvious (otherwise reader must go to declaration to see it).

2. You need to modify all platform specific `Platform::Initialize` implementations in all `runtime/bin/platform_*.cc` files in the same way. Otherwise the build will break.

mraleph (changes requested):
A better name for the PR is: `[vm] Allow to skip crash handler installation in bin::Platform::Initialize`

runtime/bin/platform_macos.cc
mraleph: Not changing the logic would be inconsistent. They should all do the same thing, so they should all get the additional parameter and respect that parameter.

runtime/bin/platform_linux.cc
```diff
@@ -65,7 +65,7 @@ static void segv_handler(int signal, siginfo_t* siginfo, void* context) {
   abort();
 }
 
-bool Platform::Initialize() {
+bool Platform::Initialize(bool install_crash_handler /* = true */) {
```
mraleph: This should be the same as Mac OS X version, but it is not updated for some reason and just ignores the parameter.

runtime/bin/platform_win.cc
mraleph: I think we want to `if (install_crash_handler)` just this one. Not other things in this function.

Arbaaz123676:
@mraleph, Updated macOS, Linux and Windows implementations to consistently respect "install_crash_handler". Only crash handler setup is conditional now.

Please let me know if anything else needs adjustment.

mraleph:
I have merged the CL. Thanks for the contribution. Now we need to figure what to do with Flutter embedder. It should just call `dart::bin::Platform::Init(/*install_crash_handler=*/false)` somewhere.

Arbaaz123676:
What if we add a call to `dart::bin::Platform::Init(false)` during Flutter engine VM initialization (maybe in `dart_vm.cc`) before Dart runtime startup, so Flutter fully owns the crash handling?

mraleph:
Yeah, something like that. You should ask Flutter engine folks what's the best place to put it in.

Arbaaz123676:
Thanks! I’m not very familiar yet with the Flutter engine workflow — should I open an issue in the Flutter engine repo about where to place the `dart::bin::Platform::Init(false)` call. or is there a specific place you recommend reaching out to?

mraleph:
@Arbaaz123676 you can start by asking on Flutter Discord.

Arbaaz123676:
Hello @mraleph — one of the Flutter contributors replied with:
“Flutter does not currently call `dart::bin::Platform::Initialize()`.

Calling `Platform::Initialize(false)` within the DartVM constructor in https://github.com/flutter/flutter/blob/master/engine/src/flutter/runtime/dart_vm.cc seems reasonable. I can try to build a PR to do this next week.

I’m not sure whether any OS-specific `Platform::Initialize` implementations might introduce undesirable side effects on some target platforms (such as Android). If so, we may need to exclude it on certain platforms.”

Based on this, I’m planning to open an issue in the Flutter repository (https://github.com/flutter/flutter/issues) also with reference of the related Dart issue (#62679) as well as the PR (#62680) for context.

I would really appreciate your guidance on whether this is the right place to track it and if there’s anything specific I should consider before opening the issue. Thank you!

PR #62745, opened 2026-02-23
[dart2js] Remove unused entries from init.types table

### LINKED ISSUE #60919: [dart2js] `init.types` table has unused entries
A certain large app has ~9.5k entries (strings) in the `init.types` table.
References to this table, via `$signature:` properties and arguments to tear-off installers, account for only about 2.5k entries. This suggests entries are being created for entities that do not have tear-offs and are not closures.

Removing the unused entries has a potential saving of > 2/3 the size of this table, or about 0.7% of the main unit.

The integer indexes into the table are efficient:
- Smaller than the corresponding entry (e.g. `123` vs `"ab(xy)"`)
- The table entries are parsed on demand, shifting parse cost from initial load to actual need
- Since the table entry strings are replaced with the parsed Rti, closures with the same type share the parsing cost

Even though there are some entries with many uses, about 70% of the entries have one use. The indirection through the table is wasteful for these entries, though it is not clear where to store the parsed Rti (it is undesirable to update an entry on a prototype).

There are about 20k closures and 30k (static and instance) tear-offs in the large app.

itsmeaabhii:
Fixes #60919

The init.types table was generating entries for all callable classes, even when they don't have tear-offs or aren't closures. This change makes the generation conditional on whether the signature function actually has generated code.

Impact:

Eliminates approximately 2/3 of unused entries (~7k out of 9.5k).

Saves about 0.7% of the main unit size.

I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkg/compiler/lib/src/js_emitter/runtime_type_generator.dart
+++ b/pkg/compiler/lib/src/js_emitter/runtime_type_generator.dart
@@ -134,10 +134,17 @@ class RuntimeTypeGenerator {
       FunctionEntity method = classFunctionType.callFunction;
       FunctionType type = classFunctionType.callType;
 
-      // TODO(johnniwinther): Avoid unneeded function type indices or
-      // signatures. We either need them for mirrors or because [type] is
-      // potentially a subtype of a checked function. Currently we eagerly
-      // generate a function type index or signature for all callable classes.
+      // Only generate function type signatures for closures or when the
+      // signature function exists and has generated code. This avoids
+      // creating unused entries in the types table for callable classes that
+      // don't have tear-offs.
+      js_ast.Expression? signatureFunctionCode =
+          generatedCode[classFunctionType.signatureFunction];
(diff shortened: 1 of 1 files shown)


mraleph:
> Impact:
> 
> Eliminates approximately 2/3 of unused entries (~7k out of 9.5k).
> 
> Saves about 0.7% of the main unit size.

How did you measure this? You are just quoting number from the issue.

Same question I asked on a different PR: Did you read and understand the code and are fairly sure it fixes the problem you are trying to fix?

I am just cautioning you: you have made three PRs in unrelated parts of Dart SDK. If they all end up being AI slop, we will just ban you from the repo.

mraleph:
Note that SDK does not even build with this change. See build failures on https://dart-review.googlesource.com/c/sdk/+/482900

How did you test it?

mraleph:

PR #62747, opened 2026-02-23
Fix pattern completion without label

itsmeaabhii:
Fixes #61879

Description
This PR fixes type completion not working in pattern matching when using the shorthand getter pattern syntax (without explicit field labels).

Problem
When typing case A(:in^ weirdName), the completion system was not suggesting getter names starting with "in". This only affected the shorthand syntax - the labeled version case A(weirdName: in^) worked correctly.

Example from the issue:

class A {
  int? weirdName;
  int? interestingField;
  int? internalValue;
}

void f(A a) {
  switch (a) {
    case A(:in^ weirdName): // Completion didn't work here
    case A(weirdName:in^ ): // This worked fine
  }
}

Solution
Modified visitPatternField in [in_scope_completion_pass.dart](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to check if the cursor is still within the getter name token after the colon. When completing within that range, it now suggests getter names instead of immediately jumping to pattern suggestions.

Changes
Updated completion logic to handle partial getter names in pattern fields
Added test case to verify the fix works correctly

### DIFF
--- a/pkg/analysis_server/lib/src/services/completion/dart/in_scope_completion_pass.dart
+++ b/pkg/analysis_server/lib/src/services/completion/dart/in_scope_completion_pass.dart
@@ -2404,8 +2404,30 @@ class InScopeCompletionPass extends SimpleAstVisitor<void> {
         isTypeNeeded: true,
       );
     } else {
-      collector.completionLocation = 'PatternField_pattern';
-      _forPattern(node, mustBeConst: false);
+      // Check if we're still completing within the name token after the colon.
+      // If so, suggest getter names. Otherwise, suggest patterns.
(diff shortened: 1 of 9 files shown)


mraleph:
This contains unrelated commits as well. 

I really don't want to waste engineers time reviewing things which are just AI slop, so please make sure that you 

1. Have PR clean from unrelated changes.
2. Have tested PRs by building SDK and running appropriate tests
3. Have read and understand all code you are submitting in the PR.

mraleph:

PR #62755, opened 2026-02-24
Mark #sizeOf getter as synthetic in FFI struct transformation

Gurleen-kansray:
The #sizeOf getter generated by _addSizeOfGetter() in pkg/vm/lib/modular/transformations/ffi/definitions.dart is currently not marked as synthetic.

The corresponding #offsetOf getters in the same transformation are already marked with isSynthetic = true. For consistency, the #sizeOf getter should also be marked synthetic.

This issue was identified during GSoC 2026 exploration of the Native Memory Inspector project and suggested by Daco Harkes.

This change adds ..isSynthetic = true to the #sizeOf getter in _addSizeOfGetter().


dcharkes:
@Gurleen-kansray see the Gerrit CL for test failures.

Issue #62776, opened 2026-02-26
Add JsonNumericMode parameter to jsonDecode for opt-in all-double number parsing


fabricio-costa:
# Proposal: Add `JsonNumericMode` parameter to `jsonDecode` for opt-in all-double number parsing

## Problem

When `jsonDecode` encounters `5` in a JSON string, it returns `int`. When it encounters `5.0`, it returns `double`. The JSON specification (ECMA-404, RFC 8259) makes no distinction — both are simply "number".

This causes `TypeError` at runtime in Flutter/Dart applications when a backend returns a number without a decimal point for a field the Dart model expects as `double`:

```dart
class Product {
  final String name;
  final double price;

  Product.fromJson(Map<String, dynamic> json)
      : name = json['name'] as String,
        price = json['price'] as double; // TypeError: type 'int' is not a subtype of type 'double'
}

void main() {
  // Backend returns price as 29 (no decimal) — perfectly valid JSON
  final data = jsonDecode('{"name": "Widget", "price": 29}');
  final product = Product.fromJson(data); // CRASHES
}
```

This is one of the most common runtime errors in Flutter applications consuming REST APIs, because **the developer has no control over whether the server sends `29` or `29.0`** — both are semantically identical in JSON.

## Current workarounds and their cost

The ecosystem has converged on manual casting:

```dart
price = (json['price'] as num).toDouble();
```

This must be applied to **every double field in every model class** across an entire codebase. For a typical Flutter app with 50+ model classes, this means hundreds of manual casts that:

- Are easy to forget (the error only surfaces at runtime, not compile time)
- Add noise to model code
- Must be taught to every new Dart developer
- Are not discoverable — developers hit the crash first, then search for solutions

## Existing issues

This problem has been reported multiple times:

| Issue | Status | Summary |
|-------|--------|---------|
| [#46883](https://github.com/dart-lang/sdk/issues/46883) | Closed (wontfix) | JSON decoding is not JSON-spec compliant for numbers |
| [#42622](https://github.com/dart-lang/sdk/issues/42622) | Closed | `type 'int' is not a subtype of type 'double'` |
| [#56913](https://github.com/dart-lang/sdk/issues/56913) | Closed (as-intended) | Same TypeError, confirmed as intentional behavior |
| [#55499](https://github.com/dart-lang/sdk/issues/55499) | **Open** | Feature request for numeric flags in jsonDecode |
[...]

Issue #56913, opened 2024-10-18
type 'int' is not a subtype of type 'double?'


dhilshana:
This usually happens when an API response or data processing step returns an int value, but my model or function is expecting a double?.

When a double variable is assigned an int value, implicit conversion from int to double is possible. However, the mentioned error occurs in this case.

jasimrazi:
I'm experiencing the same problem. Can you suggest a solution?

mraleph:
Dart[^1] does not implicitly coerce `int` to `double` (or vice versa), because (among other reasons) such conversions are not loss-less. You have to explicitly convert using `num.toDouble()` or `num.toInt()`.  

[^1]: With the exception of JavaScript backend (`dart2js`) which uses JavaScript numbers to represent both `int` and `double`.

ulisseshen:
its confuse, because if you declare a variable as `double my_double = 100;` it will be acepted and no static erros will happen. It suggest to us that the same will work in the `double.parse('100')` when we received it from API or other data source.

bwilkerson:
I can see how that can be confusing. It happens because of the following text in the language spec:

> Let _l_ be an integer literal that is not the operand of by a unary minus operator, and let _T_ be the static context type of _l_. If `double` is assignable to _T_ and `int` is not assignable to _T_, then the static type of _l_ is `double`; otherwise the static type of _l_ is `int`.
>
> _This means that an integer literal denotes a `double` when it would satisfy the type requirement, and an `int` would not. Otherwise it is an `int`, even in situations where that is an error._

That can make it seem as if Dart does coercions, but it doesn't, it just changes the interpretation of a numeric literal under one special condition.

Issue #55499, opened 2024-04-17
Feature Request: Improve support for encoding and decoding 64-bit integers to and from JSON


austinmilt:
### Problem
Currently there is no way to safely encode and decode 64-bit integers to and from JSON using `dart:convert`. For example:

```dart
import 'dart:convert';

void main() {
  print(jsonDecode("18446744073709551615"));
}
```

prints `18446744073709552000` on web due to [well-documented truncation](https://dart.dev/guides/language/numbers)

and trying to encode a `BigInt` - recommended above for arbitrarily-sized integers -

```dart
import 'dart:convert';

void main() {
  print(jsonEncode(BigInt.parse("18446744073709551615")));
}
```

results in an error since `BigInt` is not supported for JSON encoding by default.

While JSON and JavaScript are heavily related and therefore it's normal to expect JS levels of precision, JSON is used widely for non-web applications. 

### Existing workarounds are insufficient
Right now, the most straightforward workaround is to encode and decode `BigInt` to and from `String` by passing custom converters to [json.encode](https://api.dart.dev/stable/3.3.0/dart-convert/JsonCodec-class.html). However, this creates an ambiguous string value at the other end of the wire when performing remote calls, which could be cumbersome if the other end is in a different language such that encoding/decoding libraries aren't shared. It also makes the [JSON less human-readable](https://www.json.org/json-en.html).

Other workarounds have similar issues.

### Qualities of a better approach

It is prudent for `dart:convert` to support 64-bit (and larger) integers out-of-the-box. A better approach should

1. Avoid silent failures such as truncation.
2. Enable developers to decide their own tradeoffs wherever they are using JSON.
3. Facilitate low-dependency development where core package features can solve major issues with incremental changes.

### Option A - Allow overriding stringifying objects and decoding tokens

With this nonbreaking change, developers can customize how an object is added to the JSON string produced by `jsonEncode`. This is a layer deeper than the current `toEncodable` argument which is limited to return JSON-encodable Dart objects. This would enable behavior like `jsonEncode({'a': BigInt.from(1)}) => '{"a": 1}'` rather than the current workaround like `jsonEncode({'a': BigInt.from(1)}) => '{"a": "1"}'` (where `BigInt` becomes JSON string).

Also with this change, developers can customize how a JSON token is converted from a string back into an object within `jsonDecode`. This would enable a developer to reverse the flow above to produce untruncated `BigInt` from valid JSON numbers.

[...]

Issue #46883, opened 2021-08-11
Json decoding is not JSON-compliant because of int/double parsing


Hixie:
The following three JSON files are semantically identical according to the JSON standards:

1. `{ "list": [ "a", "b", "c" ], "index": 1 }`
2. `{ "list": [ "a", "b", "c" ], "index": 1.0 }`
3. `{ "list": [ "a", "b", "c" ], "index": 1e0 }`

However, in the Dart VM they are treated differently. For example, consider the following code:

```dart
import 'dart:convert';

// `input` is expected to be a JSON string in the following form:
//
// ```
// {
//   'list': [ 'a', 'b', 'c' ],
//   'index': 1,
// }
// ```
//
// That is, an object with two keys, one of which, `list`, contains an array of
// values, and the other, `index`, contains a reference to one of the values by
// identifiying its index in the array. In the example above, 'b' is identified.
Object? test(String input) {
  final dynamic model = json.decode(input);
  final dynamic index = model['index'];
  return model['list'][index];
}

void main() {
  print(test('{ "list": [ "a", "b", "c" ], "index": 1 }'));
  print(test('{ "list": [ "a", "b", "c" ], "index": 1.0 }'));
  print(test('{ "list": [ "a", "b", "c" ], "index": 1e0 }'));
}
```

I'm not really sure what the right answer is here. It's probably "wontfix", and the documentation requested in https://github.com/dart-lang/sdk/issues/46682 should just say that developers should not rely on the returned values being `int` or `double` and should instead always say `toInt()` or `toDouble()` as desired. I could also see an argument that we should always use `double`, since that would be less bug-prone.

lrhn:
It is indeed a wontfix.

The [JSON](https://www.json.org/json-en.html) [specifications](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) do not define semantics, only syntax.
They do not say that `1` and `1.0` represents the "same number", or even which number they represent, although the ECMA standard does mention that:
> Numeric values that cannot be represented as sequences of digits (such as Infinity and NaN) are not permitted.

(Pro-tip, anything above 1.7976931348623159e+308 overflows to infinity when parsed as a double, I'd probably use 1e400, it's easier to remember and recognize, and the `00` looks like &infin;).

It even goes as far as stating:
> JSON is agnostic about the semantics of numbers. In any programming language, there can be a variety of
[...]

PR #62777, opened 2026-02-26
Add JsonNumericMode parameter to jsonDecode for opt-in all-double parsing

fabricio-costa:
## Summary

- Adds `JsonNumericMode` enum to `dart:convert` with two values: `preserveType` (default) and `allDouble`
- Adds optional `numericMode` parameter to `jsonDecode()`, `JsonCodec.decode()`, and `JsonDecoder`
- When `allDouble` is used, all JSON numbers are parsed as `double`, aligning with ECMA-404's single "number" type

This is 100% backwards compatible — the default behavior is unchanged.

## Motivation

`jsonDecode('{"price": 5}')` returns `int` for `price`, but `jsonDecode('{"price": 5.0}')` returns `double`. Since JSON makes no distinction between the two, this causes `TypeError` in Flutter apps when a backend returns a number without a decimal point for a field expected as `double`.

The current workaround (`(json['x'] as num).toDouble()`) must be applied manually to every double field in every model class. This proposal provides an opt-in solution at the parser level.

**Related issue:** #62776
**See also:** #46883, #42622, #55499, #56913

## Implementation

The conversion happens at the **listener level**, not in `parseNumber`:

| Platform | Change |
|----------|--------|
| VM | `_JsonListener.handleNumber` converts `int` → `double` when `allDouble` |
| JS | Wraps `JSON.parse` post-processing with int→double synthetic reviver |
| WASM | `_JsonListener.handleIntegerNumber` short-circuits to `double` |

This keeps the performance-critical parsing code untouched. Zero overhead when using the default `preserveType` mode.

## Usage

```dart
// Opt-in: all numbers become double
final data = jsonDecode(
  '{"price": 29, "quantity": 3}',
  numericMode: JsonNumericMode.allDouble,
);
print(data['price'] is double); // true

// Default: unchanged behavior
final data2 = jsonDecode('{"price": 29}');
print(data2['price'] is int); // true
```

## Files changed

- `sdk/lib/convert/json.dart` — public API (enum + parameters)
- `sdk/lib/_internal/vm/lib/convert_patch.dart` — VM backend
- `sdk/lib/_internal/js_runtime/lib/convert_patch.dart` — JS backend
- `sdk/lib/_internal/wasm/lib/convert_patch.dart` — WASM backend
- `tests/lib/convert/json_numeric_mode_test.dart` — test coverage

## Test plan

- [ ] Existing `tests/lib/convert/json_*` tests pass (backwards compat)
- [ ] New `json_numeric_mode_test.dart` covers:
  - Default `preserveType` behavior unchanged
  - `allDouble` converts integers to double (basic, zero, negative, large)
  - Scientific notation handled correctly
  - Non-numeric values (null, bool, string) unaffected
  - Nested structures (objects, arrays, deeply nested)
  - Reviver integration (receives converted doubles)
  - `JsonCodec.decode()` and `JsonDecoder` constructor paths

### DIFF
--- a/sdk/lib/_internal/js_runtime/lib/convert_patch.dart
+++ b/sdk/lib/_internal/js_runtime/lib/convert_patch.dart
@@ -25,7 +25,7 @@ import 'dart:collection' show LinkedHashMap, MapBase;
 ///
 /// Throws [FormatException] if the input is not valid JSON text.
 @patch
-_parseJson(String source, reviver(key, value)?) {
+_parseJson(String source, reviver(key, value)?, JsonNumericMode numericMode) {
   if (source is! String) throw argumentErrorValue(source);
 
   var parsed;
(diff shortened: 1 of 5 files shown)


mraleph:
Closing this as the linked proposal is withdrawn.

PR #62790, opened 2026-02-28
analyzer: handle glob patterns in workspace field validator

Barbirosha:
## Problem

Since Dart 3.11, pub supports glob patterns in the `workspace` field of `pubspec.yaml` (e.g. `packages/*`). The analyzer's workspace validator was never updated to handle this, so it incorrectly reported `path_does_not_exist` for any glob pattern — because it tried to check the literal path `packages/*` on disk, which doesn't exist.

`dart pub get` worked fine; only the analyzer showed a false error.

Fixes #62661

## Solution

For glob patterns (detected via `Glob.quote(s) != s`, consistent with pub's own `_looksLikeGlob`), validate only the base directory — the path prefix before the first wildcard segment:

| Pattern | What we check |
|---|---|
| `packages/*` | `packages/` exists |
| `apps/nested/pkg?` | `apps/nested/` exists |
| `*` | nothing (root always exists) |

Non-glob paths continue to be validated as before.

## Tests

Added tests covering:
- Glob patterns with existing base directory → no error
- Glob patterns with missing base directory → `path_does_not_exist`
- Patterns starting with a wildcard → no error

### DIFF
--- a/pkg/analyzer/lib/src/pubspec/validators/workspace_validator.dart
+++ b/pkg/analyzer/lib/src/pubspec/validators/workspace_validator.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/src/diagnostic/diagnostic.dart' as diag;
 import 'package:analyzer/src/pubspec/pubspec_validator.dart';
 import 'package:analyzer/src/util/yaml.dart';
+import 'package:glob/glob.dart';
 import 'package:path/path.dart' as path;
 import 'package:yaml/yaml.dart';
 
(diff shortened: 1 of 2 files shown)


mraleph:
There are failures in the CL. See https://dart-review.googlesource.com/c/sdk/+/484320?tab=checks

mraleph:
There is still a test failure on the test that you have added. Did you try to run it? Please test before uploading by building the SDK and running `python3 tools/test.py -n analyzer-unittest-asserts-release-linux pkg/analyzer/test/src/pubspec/diagnostics/workspace_field_test`

PR #62881, opened 2026-03-12
Fix enum quick-fix insertion for multi-line enums

mayanksharma9:
Fixes #62874.

This addresses a quick-fix insertion bug for multi-line enums that do not yet
have an explicit semicolon after their constants.

When `Create field` or `Create getter` adds the first enhanced enum member, the
shared enum-member insertion logic was using the last enum constant's end
offset. For enums written with a trailing comma, that can place the insertion
before the comma, producing invalid output such as `field;,`.

This change updates the shared insertion helper to insert after the token before
`}`, which correctly handles enums with trailing commas but no semicolon.

It also adds regression coverage for:
- the shared enum field insertion helper
- `Create field` on a multi-line enum without a semicolon
- `Create getter` on a multi-line enum without a semicolon

Tests run locally:
- Not run in this environment

Why tests were not run locally:
- The Dart SDK repo documents that a functional environment requires `fetch` /
  `gclient` rather than a plain `git clone`.
- This environment does not have a `dart` binary available, so the relevant
  analyzer/analysis_server tests could not be executed honestly here.

### DIFF
--- a/pkg/analysis_server/test/src/services/correction/fix/create_field_test.dart
+++ b/pkg/analysis_server/test/src/services/correction/fix/create_field_test.dart
@@ -102,6 +102,35 @@ int f(E e) {
 ''');
   }
 
+  Future<void> test_usedAsGetter_multiline_withoutSemicolon() async {
+    await resolveTestCode('''
+enum E {
+  longEnumName,
+  anotherLongEnumName,
+  yetAnotherLongEnumName,
+  thisOneIsLongToo,
+}
+
+int f(E e) {
+  return e.a;
+}
+''');
+    await assertHasFix('''
+enum E {
+  longEnumName,
+  anotherLongEnumName,
+  yetAnotherLongEnumName,
+  thisOneIsLongToo;
+
+  final int a;
+}
+
+int f(E e) {
+  return e.a;
+}
+''');
+  }
+
   Future<void> test_usedAsGetter_dotShorthand() async {
     await resolveTestCode('''
 enum E {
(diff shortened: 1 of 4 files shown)


mraleph:
> Why tests were not run locally:

The same comment as on the other PR. This is utter disrespect of our time and you need to be ashamed about it. 

Please come back after you have run the tests locally and verified that the code works. 

Also please read and understand the code that your AI clanker produced. If you don't want to read - then please close the PR and don't waste our time.

csells:
I never read the code my agents produce. That just slows everything down. I do have an orchestration that I use to turn vibed code into product ready code, but not because I read it.

bwilkerson:
> I never read the code my agents produce. That just slows everything down.

It slows everything down for you, but it shifts the burden of reviewing the code to our team.

And if we didn't review it, but just committed it (which we can't do for security reasons), then it would only serve to shift the burden from implementation to maintenance, where it's even more expensive to fix the bugs that the agents produced.

So please review all agent produced code before submitting it to our team.

csells:
> So please review all agent produced code before submitting it to our team.

I have the code reviewed 6 ways till Sunday. It's just not me doing it.

mayanksharma9:
Thanks for the direct feedback. You are right that I should not keep asking for review until I have personally reviewed the change, run the relevant tests locally, and made sure I can explain the implementation clearly. I am closing this for now rather than burn more reviewer time. If I bring it back, it will be from a properly verified state.

PR #62882, opened 2026-03-12
Fix create method for named-arguments-anywhere calls

mayanksharma9:
## Summary
Fix the `Create method` quick fix so it handles named-arguments-anywhere call sites correctly.

Closes #62878.

## Issue
For an undefined method call like:

```dart
c.m(arg1: true, 0, arg3: 'string');
```

the quick fix was generating:

```dart
void m({required bool arg1, int i, required String arg3}) {}
```

That is invalid for the intended API shape because the positional argument was being emitted inside the named parameter section.

## What changed
`writeParametersMatchingArguments()` now separates positional and named arguments before writing the generated parameter list. This preserves the original positional/named classification even when the call site uses named arguments before later positional arguments.

I also added focused tests for:
- the shared parameter writer in `analyzer_plugin`
- the analysis server `Create method` fix reproducer from the issue

## Why this fix
The bug was in shared parameter synthesis, not in the method fix itself. Fixing it there keeps the behavior consistent for code that relies on the same helper.

## Testing
I added regression tests, but I could not run them locally in this environment.

This checkout was created with plain `git clone`, while the repo contribution docs say the Dart SDK expects a `gclient`/depot_tools setup for a functional environment. Also, there is no `dart` binary available on PATH here, so I couldn't run:
- `dart test pkg/analyzer_plugin/test/src/utilities/change_builder/change_builder_dart_test.dart`
- `dart test pkg/analysis_server/test/src/services/correction/fix/create_method_test.dart`

### DIFF
--- a/pkg/analysis_server/test/src/services/correction/fix/create_field_test.dart
+++ b/pkg/analysis_server/test/src/services/correction/fix/create_field_test.dart
@@ -102,6 +102,35 @@ int f(E e) {
 ''');
   }
 
+  Future<void> test_usedAsGetter_multiline_withoutSemicolon() async {
+    await resolveTestCode('''
+enum E {
+  longEnumName,
+  anotherLongEnumName,
+  yetAnotherLongEnumName,
+  thisOneIsLongToo,
+}
+
+int f(E e) {
+  return e.a;
+}
+''');
+    await assertHasFix('''
+enum E {
+  longEnumName,
+  anotherLongEnumName,
+  yetAnotherLongEnumName,
+  thisOneIsLongToo;
+
+  final int a;
+}
+
+int f(E e) {
+  return e.a;
+}
+''');
+  }
+
   Future<void> test_usedAsGetter_dotShorthand() async {
     await resolveTestCode('''
 enum E {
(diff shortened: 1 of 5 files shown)


mraleph:
This contains code from another PR you have made. Please use different branches to upload unrelated PRs.

mraleph:
Also what on Earth is this:

> I added regression tests, but I could not run them locally in this environment.

Please make an effort to run/test the code you have uploaded. We are not going to waste time looking at the code which you have produced with LLM assistance and did not even bother to test yourself.

mraleph:
Just FYI: I have landed an update to CONTRIBUTING.md covering AI assisted changes (it is more or less the same policy as Flutter has adopted): 

https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md#ai-contribution-guidelines

Please make sure to read and adhere to it. Thanks.

mayanksharma9:
Thanks. The branch hygiene issue and the testing point are both fair. I should not keep iterating on this in its current state. I am closing it so I can redo it cleanly on an isolated branch after I have personally reviewed the code and run the relevant tests locally.

PR #62924, opened 2026-03-17
Fixes #62864 Improve error message in the analyzer

### LINKED ISSUE #62864: Improve error message in the analyzer in case of `void` in a spread
```dart
main() {
  void e = <void>[];
  <void>[...e]; // Two errors here.
                 // A nullable expression can't be used in a spread. Try checking that the value isn't 'null' before using it in a spread, or use a null-aware spread.
                 // Spread elements in list or set literals must implement 'Iterable'. • not_iterable_spread
  <void>[...?e]; //Spread elements in list or set literals must implement 'Iterable'. • not_iterable_spread
}
```
Analyzer reports two errors for `...e`
```
• A nullable expression can't be used in a spread. Try checking that the value isn't 'null' before using it in a
          spread, or use a null-aware spread. • unchecked_use_of_nullable_value
• Spread elements in list or set literals must implement 'Iterable'. • not_iterable_spread
```
In case `...?` only the last one is reported. It'd better in both cases say that `e` has type `void` and therefore cannot be used.

`Dart SDK version: 3.12.0-221.0.dev (dev) (Thu Mar 5 20:03:30 2026 -0800) on "windows_x64"`

Darkildo:
Fix analyzer diagnostics for spread expressions with `void` type.

### What this PR changes
- Adds an early `useOfVoidResult` check in spread verification for list/set and map literals.
- Prevents `uncheckedUseOfNullableValueInSpread` from being reported for non-null-aware spreads when the spread expression has type `void`.
- Adds regression tests for list/set/map spreads (including null-aware spreads) to ensure only `useOfVoidResult` is reported.

### Why
Using a `void` expression in a spread should report `useOfVoidResult` and stop further spread validation. This avoids cascading or misleading diagnostics in this invalid scenario.


mraleph:
> **Gerrit [CL](https://dart-review.googlesource.com/c/sdk/+/488528) has build or test failures as well as new comments, please review them in Gerrit.**

@Darkildo ⬆️

mraleph:
Note that we tried to land CL few weeks ago but it fails tests in CQ. Please either address the failures or close the PR.

mraleph:
Last ping before I close this PR as stale.

mraleph:
Closing as stale.

PR #62930, opened 2026-03-18
Filter uninteresting file watcher events at context ingress

### LINKED ISSUE #62805: Consider dropping uninteresting watcher events earlier
Splitting this out of https://github.com/dart-lang/sdk/issues/62785 (see https://github.com/dart-lang/sdk/issues/62785#issuecomment-3985158274).

Currently watch events for files like `.txt` go through quite a bit of code before being ignored, but while we only respond to certain kinds of file changes (and [this now includes plugins](https://github.com/dart-lang/sdk/issues/62785#issuecomment-3985221778)), events for uninteresting files could be dropped much earlier (perhaps while being added to the message scheduler?). Similarly, duplicate events could be dropped here.

@bwilkerson although it occurs to me that doing this now will break old plugins that do rely on this. I don't know whether this is an issue or not (or if there is a plan for when the old plugins cease to be supported).

rootvector2:
## Description

This change implements early filtering of file watcher events in the analysis server to reduce unnecessary work and improve performance.

Currently, all watcher events are queued and processed, even for files that are not relevant to analysis. This PR updates the event handling logic to drop uninteresting events at the point of ingress based on analyzer context relevance.

### What this PR does

- Filters file watcher events early using `contextRoot.isAnalyzed(path)`
- Ensures only events for files within analyzed contexts are processed
- Preserves all behavior for relevant files (e.g., Dart sources and analyzed workspace files)
- Avoids introducing custom file-type allowlists to prevent drift from analyzer logic

### Important details

- No scheduler-level deduplication is introduced (to avoid ordering/semantic risks)
- Filtering is based entirely on analyzer context, not file extensions
- `REMOVE` events are still allowed to ensure proper cleanup and state consistency

### Why this change

As discussed in the issue, many watcher events are not relevant to analysis and can be dropped earlier to reduce load on the analysis server. This change aligns with that goal while keeping correctness and existing behavior intact.

---

## Related Issue

Fixes: https://github.com/dart-lang/sdk/issues/62805

---

## Testing

Added/updated tests to verify:

- Events outside analyzed context are ignored
- Events inside analyzed context are processed correctly
- Irrelevant events do not block subsequent relevant events

All updated tests pass locally.

---

## Notes

This change keeps the implementation minimal and aligned with existing analyzer behavior by relying solely on context-based filtering instead of introducing new heuristics.

---

## Checklist


mraleph:
> **Gerrit [CL](https://dart-review.googlesource.com/c/sdk/+/488760) has build or test failures as well as new comments, please review them in Gerrit.**

@rootvector2 please see this ⬆️

mraleph:
There is a question on the CL. Please either answer the question or close the PR as stale. Thanks

mraleph:
Last ping before I close this PR as stale.

mraleph:
Closing as stale.

PR #62941, opened 2026-03-20
FIX #62280 Parser and analyzer AST should be updated to reject augmented enum values

Darkildo:
This PR aligns the parser and analyzer with `dart-lang/language@ebd9e18`, which removed support for augmenting enum values.

It changes the parser so code like:

```dart
enum E {
  foo
}

augment enum E {
  augment foo
}
```
is no longer accepted as a valid augmented enum constant declaration. Instead, the parser reports the leading augment as an extraneous modifier and recovers on the enum value name.

On the analyzer side, this PR also removes the internal use of enum-value augmentation data and deprecates EnumConstantDeclaration.augmentKeyword. That getter is part of the public analyzer API and is not experimental, so deprecating it now preserves semantic versioning and makes its future removal explicit.

The PR also updates parser and summary tests to reflect the new behavior.

Fix  #62280

### DIFF
--- a/pkg/_fe_analyzer_shared/lib/src/parser/parser_impl.dart
+++ b/pkg/_fe_analyzer_shared/lib/src/parser/parser_impl.dart
@@ -2878,10 +2878,14 @@ class Parser {
     Token beginToken = token;
     token = parseMetadataStar(token);
 
-    Token? augmentToken;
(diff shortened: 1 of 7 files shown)


mraleph:
@scheglov please either review or reject this explicitly.

PR #62943, opened 2026-03-20
FIX #55145 Parse failure for broken augmentation library 

### LINKED ISSUE #55145: Parse failure for broken augmentation library
https://dart-review.googlesource.com/c/sdk/+/356401 adds a test case with a broken augmentation library, it trips an assert in the parse test

https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8754039752317012545/+/u/test_results/new_test_failures__logs_

```
'package:_fe_analyzer_shared/src/parser/parser_impl.dart': Failed assertion: line 846 pos 12: 'optional('augment', augmentKeyword)': is not true.

#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      Parser.parseLibraryAugmentation (package:_fe_analyzer_shared/src/parser/parser_impl.dart:846:12)
#3      Parser.parseTopLevelKeywordDeclaration (package:_fe_analyzer_shared/src/parser/parser_impl.dart:777:20)
#4      Parser.parseTopLevelDeclarationImpl (package:_fe_analyzer_shared/src/parser/parser_impl.dart:530:14)
#5      Parser.parseUnit (package:_fe_analyzer_shared/src/parser/parser_impl.dart:410:15)
```

I'll mark it approved since the failure is unimportant for the PR which adds test_runner support for expectations in augmentation libraries.

Darkildo:
- Fixes a parser crash when the parser test suite sees a broken augmentation library and updates augmentation library coverage to stay aligned with the current feature direction.

This PR changes the handling of augmentation library parsing so that malformed augmentation input is reported through normal parser recovery instead of tripping an assertion in `_fe_analyzer_shared`. The issue showed up after https://dart-review.googlesource.com/c/sdk/+/356401 added a test case with a broken augmentation library, which caused parse tests to fail with:

`Failed assertion: 'optional('augment', augmentKeyword)'`

The change keeps the parser resilient when augmentation syntax is invalid and adds regression coverage for that case. It also updates augmentation-related tests to match the current augmentation library specification so the test suite reflects the currently supported behavior instead of older experimental semantics.

Relevant context:
- https://dart-review.googlesource.com/c/sdk/+/356401
- https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8754039752317012545/+/u/test_results/new_test_failures__logs_
- https://github.com/dart-lang/language/blob/ebd9e1882484380edc6bc3c5cccad23ea1ea845b/working/augmentation-libraries/feature-specification.md

---

FIX #55145


davidmorgan:
As mentioned on the linked issue: I don't think this is the right fix as "import augment" is no longer part of the feature. I've reverted the test infra change that prompted the issue, and removed the `augmentation_libraries` test.

PR #62949, opened 2026-03-22
Add TODO for misleading dot shorthand diagnostic message (#62863)

### LINKED ISSUE #62863: Analyzer misleading message for dot-shorthand on non-scope-type
Example:
```dart
import 'dart:async';
void main() {
  foq(.hash); // Works (base type Object)
  // Fails (correct).
  
  fd(.hash); 
  // CFE: The static getter or field 'hash' isn't defined for the type 'dynamic'.
  // Analyzer: A dot shorthand can't be used where there is no context type.
  
  fv(.hash);
  // CFE: The static getter or field 'hash' isn't defined for the type 'void'.
  // Analyzer: A dot shorthand can't be used where there is no context type.
  
  fr(.hash);
  // CFE: The static getter or field 'hash' isn't defined for the type '(Object?)'.
  // Analyzer: A dot shorthand can't be used where there is no context type.
}

void foq(Object? _) {}
void fd(dynamic _) {}
void fv(void _) {}
void fr((Object?,) _) {}
```

The analyzer's message suggests that there is no context type.
There is a context type, but it's a context type that doesn't have a static scope.

The CFE error message is neutral, it just says there is no static field of that name, which can be both because there is no static scope to look in, or because there is one, but there is no `hash` member in it.
The Analyzer uses the same error message as the CFE when the context type denotes a static scope.
(Although the message then goes on to say
> Try correcting the name to the name of an existing static getter or field, or defining a getter or field named 'hash'.

which isn't helpful for a type that cannot have static members. That's a good reason to have distinct error messages for the two cases, the Analyzer's message for the non-static-scope case is just a little misleading.)

console-log-life:
Fixes #62863

The analyzer message for dot shorthand suggests that there is no context type.
However, in cases like `dynamic` or `void`, a context type exists but it does not have a static scope.

This change adds a TODO comment highlighting that the current diagnostic message is misleading.

No functional behavior is changed.

### DIFF
--- a/pkg/analyzer/lib/src/dart/resolver/instance_creation_expression_resolver.dart
+++ b/pkg/analyzer/lib/src/dart/resolver/instance_creation_expression_resolver.dart
@@ -118,7 +118,7 @@ class InstanceCreationExpressionResolver {
         );
       }
     } else {
-    // TODO: Improve this message. Context type exists in some cases (e.g. dynamic, void)
+    // TODO: Improve this message. Context type exists in some cases (e.g. dynamic, void) 
     // but has no static scope. Current message is misleading (see issue #62863).
     _resolver.diagnosticReporter.report(
       diag.dotShorthandMissingContext.at(node),
--- a/pkg/analyzer/lib/src/dart/resolver/instance_creation_expression_resolver.dart
+++ b/pkg/analyzer/lib/src/dart/resolver/instance_creation_expression_resolver.dart
@@ -118,9 +118,11 @@ class InstanceCreationExpressionResolver {
         );
       }
     } else {
-      _resolver.diagnosticReporter.report(
-        diag.dotShorthandMissingContext.at(node),
-      );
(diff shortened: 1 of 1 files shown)


mraleph:
@Apeksha-22 This is a meaningless change. Also take a careful look at the diff and note that the code is not formatted correctly after your change - so it is a sloppy change. Please don't send meaningless & sloppy changes.

If you wish contribute - find the issue _you **understand** how to fix_  and fix it. Or if you want a stretch goal: pick an issue you do not understand how to fix, figure it out and fix it. Figuring it out is the key here.

Thanks.

console-log-life:
Hi @mraleph and @julemand101,

I sincerely apologize for the previous PR. As a newcomer, I misunderstood the contribution process and submitted an incomplete, poorly formatted change.

I have taken your feedback seriously. I am now studying the analyzer's diagnostic rules properly to provide a functional fix instead of just a comment. I'll also ensure all future contributions follow the project's formatting standards.

Thank you for the guidance and for pointing me in the right direction.

PR #63012, opened 2026-03-30
[vm, simarm64] Extract CalleeRegisterVerifier class

eseidel:
Extract the callee-saved register save/verify/restore logic from `Simulator::Call` into a RAII `CalleeRegisterVerifier` class.

- On construction, saves callee-saved register values and overwrites them with a known value
- On destruction, asserts the known value is preserved and restores originals
- Reduces ~25 lines of inline code in `Simulator::Call` to 3 lines
- No behavioral change

### DIFF
--- a/runtime/vm/simulator_arm64.cc
+++ b/runtime/vm/simulator_arm64.cc
@@ -3996,6 +3996,54 @@ void Simulator::ExecuteTrace() {
   }
 }
 
+// Verifies that callee-saved registers are preserved across Dart execution.
+// On construction, saves and overwrites callee-saved registers with a known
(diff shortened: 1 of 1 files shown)


eseidel:
This is a small clean up that we did in our fork of Dart that I figured I should upstream.

mraleph:
> **Gerrit [CL](https://dart-review.googlesource.com/c/sdk/+/491682) has build or test failures, please review them in Gerrit and fix them before requesting another review.**

@eseidel please see this ⬆️

eseidel:
Addressed.

PR #63058, opened 2026-03-31
[ linter] avoid bool literals in conditional expressions

IldySilva:
Contributes to https://github.com/dart-lang/site-www/issues/7210

### DIFF
--- a/pkg/linter/messages.yaml
+++ b/pkg/linter/messages.yaml
@@ -774,10 +774,9 @@ LinterLintCode:
     documentation: |-
       #### Description
 
-      The analyzer produces this diagnostic when a conditional expression has
-      `true` or `false` as either the then or else expression, when both
-      branches have a type of `bool`. Such expressions can always be simplified
-      using `&&` or `||`.
+      The analyzer produces this diagnostic when a conditional expression uses a
+      boolean literal, and both branches are booleans. You can simplify these
+      expressions using `&&` or `||`.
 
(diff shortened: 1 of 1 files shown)


pkg/linter/messages.yaml
```diff
@@ -771,6 +771,45 @@ LinterLintCode:
...
+
+      Replace the conditional expression with the equivalent logical expression.
+      If the then expression is a literal, replace it with one of the following:
+
+      ```dart
+      bool f(bool a, bool b) => a || b;  // replaces: a ? true : b
+      bool g(bool a, bool b) => !a && b; // replaces: a ? false : b
+      ```
+
+      If the else expression is a literal, replace it with one of the following:
```
IldySilva:
```suggestion
      If the else expression is a boolean literal, replace it with one of the following:
```

pkg/linter/messages.yaml
```diff
@@ -771,6 +771,45 @@ LinterLintCode:
...
+      is the `bool` literal `false`:
+
+      ```dart
+      bool f(bool a, bool b) => [!a ? b : false!];
+      ```
+
+      #### Common fixes
+
+      Replace the conditional expression with the equivalent logical expression.
+      If the then expression is a literal, replace it with one of the following:
```
IldySilva:
```suggestion
      If the then expression is a boolean literal, replace it with one of the following:
```

PR #63127, opened 2026-04-07
[analyzer] Fix cache size counter drift in EvictingFileByteStore._cleanUpFolder

### LINKED ISSUE #63118: Analyzer cache size accounting drift in EvictingFileByteStore
I noticed a small bug in how the analyzer tracks its file cache size in `pkg/analyzer/lib/src/dart/analysis/file_byte_store.dart`.

In `EvictingFileByteStore._cleanUpFolder`, the code tries to delete old files to stay under the limit. It uses a try-catch around `file.deleteSync()`, but it decrements the `currentSizeBytes` counter regardless of whether the deletion actually worked:

```dart
      try {
        file.deleteSync();
      } catch (_) {}
      currentSizeBytes -= fileStatMap[file]!.size;

### PR DESCRIPTION
**Pull request by vedant21-oss**
## Problem

`EvictingFileByteStore._cleanUpFolder` was decrementing `currentSizeBytes` unconditionally,
even when `file.deleteSync()` threw an exception:

```dart
try {
  file.deleteSync();
} catch (_) {}
currentSizeBytes -= fileStatMap[file]!.size; // <- always runs, even on failure
```

If a file cannot be deleted (e.g. due to a permissions error or a race with another process),
the counter is decremented as though the deletion succeeded.  This causes the cache to
**underestimate** its own size, which can prevent the eviction loop from running again even
though the cache is still over its `maxSizeBytes` limit.

## Fix

Move the decrement inside the `try` block so it only executes when the deletion actually succeeds:

```dart
try {
  file.deleteSync();
  currentSizeBytes -= fileStatMap[file]!.size; // <- only runs on success
} catch (_) {}
```

## Tests

Two regression tests are added to a new `EvictingFileByteStoreCleanUpTest` class:

* `test_cleanUpFolder_successfulDeletion_decrementsSizeCounter` - basic happy-path eviction.
* * `test_cleanUpFolder_failedDeletion_doesNotDecrementSizeCounter` - simulates a failed
*   deletion by revoking `chmod 000` on the parent directory and verifies that the files
*   remain on disk (i.e. the loop is not fooled by a corrupted size counter). Skipped on Windows.
Fixes #63118

### DIFF
--- a/pkg/analyzer/lib/src/dart/analysis/file_byte_store.dart
+++ b/pkg/analyzer/lib/src/dart/analysis/file_byte_store.dart
@@ -139,8 +139,8 @@ class EvictingFileByteStore implements ByteStore {
       }
       try {
         file.deleteSync();
+        currentSizeBytes -= fileStatMap[file]!.size;
       } catch (_) {}
-      currentSizeBytes -= fileStatMap[file]!.size;
     }
   }
 }


vedant21-oss:
@google-cla check

vedant21-oss:
@google-cla check

vedant21-oss:
@google-cla check

vedant21-oss:
Submitted a fix in #63127 — moves the `currentSizeBytes` decrement inside the `try` block so it's only decremented when deletion actually succeeds.

keertip:
@vedant21-oss , are you able to see the Gerrit code review? Did you see the comment on the CL?

vedant21-oss:
Yes, I saw Konstantin's comment on Gerrit regarding the regression test. I've just updated the PR by removing the test as suggested, and the new patchset should be pushed to Gerrit now.

PR #63141, opened 2026-04-08
Add a `JSUnsafeObject` type

nex3:
As discussed offline and in design docs, this type acts as a "pressure release" for users like Sass who need to squeeze more power and performance out of compiled-to-JS interop. It makes it possible to define APIs that operate over all types that can appear in JS, including Dart objects represented as JS types.

---


mraleph:
@nex3 if you are not planning to pick this up, consider closing the PR.

nex3:
I do very much intend to work on this, but per @srujzs doesn't want to move forward until https://github.com/dart-lang/sdk/issues/63582 is resolved.

PR #63151, opened 2026-04-10
[dart2wasm] Avoid redundant typed_data buffer allocations in setRange

### LINKED ISSUE #56663: [dart2wasm] Typed data `setRange` allocates redundantly in the fast path
The fast path in typed data `setRange` mixed-in method runs a few checks and then calls `array.copy`: https://github.com/dart-lang/sdk/blob/7e7a1ab32087fb89e34612fa93d05b7a0a5ab793/sdk/lib/_internal/wasm/lib/typed_data.dart#L1780-L1808

This code doesn't know the buffer type of `this`, so it gets the buffer as `ByteBuffer` and then checks the `ByteBuffer`s type.

However getting the `ByteBuffer` of a fast class like `I8List` allocates: https://github.com/dart-lang/sdk/blob/7e7a1ab32087fb89e34612fa93d05b7a0a5ab793/sdk/lib/_internal/wasm/lib/typed_data.dart#L2306

I can see in some benchmark programs that this allocation is not optimized away, even though it's done with a direct call to `buffer`:

```
  (func $_I32List&_WasmI32ArrayBase&_IntListMixin&_TypedIntListMixin.setRange (;586;) (param $var0 (ref $I32List_223)) (param $var1 i64) (param $var2 i64) (param $var3 (ref null $#Top)) (param $var4 (ref null $BoxedInt))
    ...
    local.get $var5
    call $<obj> is TypedData
    if
      local.get $var0
      call $_WasmI32ArrayBase.buffer    ; <-- direct call instead of virtual, but isn't inlined and allocates
      ref.cast $_I32ByteData_224
```

If we manually copy (instead of mixing it in) this method then we know the buffer type so the allocation can be avoided.

kyungilcho:
Fixes #56663.

## What

This changes the same-type Wasm typed array `setRange` fast path to copy
directly through the underlying Wasm array instead of going through `buffer`
first.

That avoids materializing `ByteBuffer` wrappers for the common same-type cases:

- `I8`
- `I16`
- `I32`
- `I64`
- `F32`
- `F64`

It also adds wasm regression coverage for same-type view-to-view `setRange`.

## Why

The issue here is that the current fast path still ends up doing extra wrapper
work even when the source and destination already have matching Wasm typed
array representations.

This keeps the existing `setRange` behavior, but removes that redundant work
from the Wasm fast path.

After rebuilding `dart2wasm_benchmark`, a small local benchmark with repeated
same-type view-to-view copies showed clear improvement for
`Uint16`/`Uint32`/`Float32`/`Float64`.

## Tests

- `python3 tools/test.py -m release -c dart2wasm -r d8 -j4 -p compact tests/web/wasm/typed_data_setrange_test`
- `python3 tools/test.py -m release -c dart2wasm -r d8 -j4 -p compact tests/lib/typed_data/setRange_`
- `python3 tools/test.py -m release -c dart2wasm -r d8 -j4 -p compact tests/lib/typed_data`

I also ran `tests/web/wasm` more broadly. The unrelated failures there
reproduced the same way with and without this patch in the local environment.

### DIFF
--- a/sdk/lib/_internal/wasm/lib/typed_data.dart
+++ b/sdk/lib/_internal/wasm/lib/typed_data.dart
@@ -1996,12 +1996,6 @@ mixin _TypedIntListMixin<SpawnedType extends TypedDataList<int>>
       final destDartElementSizeInBytes = destTypedData.elementSizeInBytes;
       final fromDartElementSizeInBytes = fromTypedData.elementSizeInBytes;
 
-      final fromBufferByteOffset =
-          fromTypedData.offsetInBytes +
-          (skipCount * fromDartElementSizeInBytes);
-      final destBufferByteOffset =
-          destTypedData.offsetInBytes + (start * destDartElementSizeInBytes);
-
       // Use `array.copy` when:
       //
       // 1. Dart array element types are the same.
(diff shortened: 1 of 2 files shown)


mkustermann:
@kyungilcho Thank you for the contribution. I've left a comment on the CL on how we could make this a bit more maintainable and also potentially reducing the code size for it.

kyungilcho:
@mkustermann Thanks, this makes sense. I had the same concern while working on this, but I kept the patch narrower at first so it stayed closer to the scope of #56663. Since you’re calling out the same tradeoff here, I think reworking this around per-type specializations may be the better shape after all. I’m happy to take it that way if you’d prefer.

I also wanted to reply on the Gerrit CL directly, but Gerrit account creation is currently failing with a 500 for me, so I’m leaving the response here for now.

kevmoo:
@kyungilcho – how did you find this? What are you building with Dart & Wasm?

kyungilcho:
@kevmoo We have a Flutter-based vector canvas package that we use on mobile, and we also build that same canvas for the web with dart2wasm and embed it into a React app. While working on image and shape handling there, I got curious how dart2wasm was actually handling those typed-data paths under the hood, so I started poking through the source. While googling one of the questions that came up, I found #56663, and it was pretty much the same thing I had been wondering about.

athomas:
> I also wanted to reply on the Gerrit CL directly, but Gerrit account creation is currently failing with a 500 for me, so I’m leaving the response here for now.

@kyungilcho would you mind trying again? There was an issue and is supposed to be resolved now. If you're still experiencing this issue, I can follow up internally.

mkustermann:
The change landed now. Thank you for the contribution, @kyungilcho!

PR #63358, opened 2026-05-10
[vm/io] Range check SynchronousSocket_WriteList arguments

LemonTeatw1:
The Dart-side wrapper passed `end - (start - bufferAndStart.start)` as the third argument to `_nativeWrite`, expecting it to be the byte count. For the common full-Uint8List path of `_ensureFastAndSerializableByteData` this expression simplifies to `end` (the END index), so the native handler sent `end` bytes from `buffer + start` instead of `end - start`. With `start > 0` and `start + end > buffer.length` -- both values that pass `RangeError.checkValidRange` -- the handler read `start + end - buffer.length` bytes past the typed-data allocation and shipped that heap memory to the socket peer.

Pass the correct length on the Dart side, and replace the debug-only ASSERT on the C side with a runtime range check that returns an ArgumentError, matching the precedent set by [vm/io] Range check Filter_Process arguments.

TEST=tests/standalone/io/regress_writefromsync_overread_test.dart

- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.

Note that this repository uses Gerrit for code reviews. Your pull request will be automatically converted into a Gerrit CL and a link to the CL written into this PR. The review will happen on Gerrit but you can also push additional commits to this PR to update the code review.
</details>

PR #63365, opened 2026-05-12
[vm/io] Range check Socket_SendMessage arguments

peeefour:
Follow-up to #63358, same fix pattern applied to `Socket_SendMessage`.

`_NativeSocket.send` and `_NativeSocket.sendMessage` (`sdk/lib/_internal/vm/bin/socket_patch.dart`) pass `bytes` (the byte count) as the third argument to `_ensureFastAndSerializableByteData(_, _, end)`, whose third parameter is the exclusive `end` index. In the helper's copy path (when `buffer` is not a full-size `Uint8List`) this means it allocates `Uint8List(bytes - offset)` instead of `Uint8List(bytes)`; the native is then told to send `bytes` bytes from the shorter buffer, reading `offset` bytes of adjacent Dart heap memory and forwarding them to the peer via `sendmsg(2)`.

Changes:
- Pass `offset + bytes` (i.e. `end`) to the helper in both `_NativeSocket.send` and `_NativeSocket.sendMessage`.
- Replace the release-no-op `ASSERT((offset + length) <= data.size_in_bytes())` in `Socket_SendMessage` (`runtime/bin/socket.cc`) with a runtime range check, mirroring the post-fix style in #63358.
- Add `tests/standalone/io/regress_sendmessage_overread_test.dart` covering the publicly-reachable `sendMessage` codepath.

The `_NativeSocket.send` callsite is not directly reachable from the public API (`RawDatagramSocket.send` always passes `offset=0`), but it shares the same off-by-one and is corrected for defense-in-depth.

### DIFF
--- a/runtime/bin/socket.cc
+++ b/runtime/bin/socket.cc
@@ -828,7 +828,12 @@ void FUNCTION_NAME(Socket_SendMessage)(Dart_NativeArguments args) {
     Dart_Handle buffer_dart = Dart_GetNativeArgument(args, 1);
     TypedDataScope data(buffer_dart);
 
-    ASSERT((offset + length) <= data.size_in_bytes());
+    const intptr_t end = offset + length;
+    if (!(0 <= offset && offset <= end && end <= data.size_in_bytes())) {
+      delete os_error;
+      Dart_SetReturnValue(args, Dart_NewApiError("Invalid range"));
+      return;
+    }
     uint8_t* buffer_at_offset =
         reinterpret_cast<uint8_t*>(data.data()) + offset;
     bytes_written = SocketBase::SendMessage(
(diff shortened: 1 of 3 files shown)


peeefour:
@googlebot I signed it!

peeefour:
Thanks for the review. Addressed all four comments in the new patch set:

- Verbose top-of-file comment removed; left only the standard copyright header.
- Added the `Platform.isLinux || Platform.isMacOS` early-return guard before any sendMessage work.
- Switched the data buffer to `List<int>.generate(16, (i) => i)` so individual byte positions are distinguishable in the assertion.
- Added `testDatagramSend()` exercising `RawDatagramSocket.send(data, …)` with a `List<int>` buffer to cover the `_NativeSocket.send` fix.

PTAL.

mraleph:
In future please resolve addressed comments in Gerrit.

peeefour:
Okay thank you, working on that now

peeefour:
Dropped the datagram-send test per your follow-up; only the sendMessage test remains. PTAL.

peeefour:
Ran `dart format` on the test file; re-pushed.

peeefour:
https://issu.ee/512376006

PR #63407, opened 2026-05-19
Add Loong64 VM backend support

guanzi008:
## Summary

This adds LoongArch64/Loong64 support across the Dart VM backend. The implementation follows the existing RISC-V and ARM64 backend layout where possible, with the Loong64-specific pieces kept behind the target architecture plumbing.

Included in this change:

- target and platform plumbing for Loong64
- assembler, disassembler, instruction decoding, and code patching support
- compiler backend and stub compiler support for JIT/AOT code generation
- runtime entry and AOT ELF/tooling support
- signed 32-bit handling for the runtime type intrinsic when reading `Class::host_type_arguments_field_offset_in_words`

## Testing

- Built `gen_snapshot` for linux-loong64
- Built a Flutter Linux release AOT bundle with Loong64 artifacts
- Ran the generated Loong64 AOT application on a LoongArch64 Linux host


mraleph:
Thank you for the contribution! Currently we only add architectures to the SDK which we _ourselves_ are ready to support, which unfortunately means we will not be able to take this PR as we do not have plans to maintain LoongArch  in-tree. 

These policies might change in the future - but I would at the very least expect that adding new in-tree architectures will require contributors to commit to ongoing maintenance of all related code, maintain CI/CD infrastructure, etc.

PR #63498, opened 2026-06-02
[analyzer] Restrict analyzer plugin sources to the default pub registry

adilburaksen:
The top-level `plugins:` section of `analysis_options.yaml` accepts a plugin source as a version constraint (optionally with a `hosted:` host), a `git:` URL, or a `path:`. The `git:` source and the `hosted:` source with a non-default host let project configuration resolve a plugin from an arbitrary URL/host, which is not a documented or intended way to specify an analyzer plugin.

This validates the plugin source resolved from project configuration in `AnalysisOptionsBuilder._getSource`:

- `git:` sources are no longer accepted (the plugin is skipped).
- `hosted:` sources are only accepted when the host is the default pub registry (`pub.dev` / `pub.dartlang.org`).

`path:`, plain version constraints, and default-pub `hosted:` sources are unchanged.

Tests in `pkg/analyzer/test/src/options/analysis_options_test.dart` are updated to assert the restricted behavior, and new cases cover both the accepted default-host `hosted:` source and the now-skipped non-default-host `hosted:` / `git:` sources.

### DIFF
--- a/pkg/analyzer/lib/src/dart/analysis/analysis_options.dart
+++ b/pkg/analyzer/lib/src/dart/analysis/analysis_options.dart
@@ -328,26 +328,22 @@ final class AnalysisOptionsBuilder {
       YamlScalar(value: String version),
       YamlScalar(value: String hostedUrl),
     )) {
+      // Only the default pub registry is accepted as a plugin source from
+      // project configuration. A plugin hosted on an arbitrary, non-default
+      // host is skipped.
+      if (!_isDefaultPluginHost(hostedUrl)) {
+        return null;
+      }
       return VersionedPluginSource(constraint: version, hostedUrl: hostedUrl);
     } else if (versionSource case YamlScalar(value: String version)) {
       return VersionedPluginSource(constraint: version);
     }
 
-    var gitSource = pluginNode.valueAt(AnalysisOptionsFileKeys.git);
-    if (gitSource case YamlScalar(:String value)) {
-      return GitPluginSource(url: value);
-    } else if (gitSource is YamlMap) {
-      var urlSource = gitSource.valueAt(AnalysisOptionsFileKeys.url);
-      if (urlSource case YamlScalar(:String value)) {
-        return GitPluginSource(
-          url: value,
-          path: gitSource.valueAt(AnalysisOptionsFileKeys.path).stringValue,
-          ref: gitSource.valueAt(AnalysisOptionsFileKeys.ref).stringValue,
-          tagPattern: gitSource
-              .valueAt(AnalysisOptionsFileKeys.tagPattern)
-              .stringValue,
-        );
-      }
+    // A 'git' source points a plugin at an arbitrary URL, which is not a
+    // supported way to specify an analyzer plugin in project configuration; it
+    // is skipped.
+    if (pluginNode.valueAt(AnalysisOptionsFileKeys.git) != null) {
+      return null;
     }
 
     var pathSource = pluginNode.valueAt(AnalysisOptionsFileKeys.path);
@@ -372,6 +368,12 @@ final class AnalysisOptionsBuilder {
       return PathPluginSource(path: pathValue);
     }
     return null;
+  }
+
+  /// Whether [hostedUrl] refers to the default pub package registry.
+  static bool _isDefaultPluginHost(String hostedUrl) {
+    var host = Uri.tryParse(hostedUrl)?.host;
+    return host == 'pub.dev' || host == 'pub.dartlang.org';
   }
 }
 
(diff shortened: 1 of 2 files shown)


mraleph:
I would imagine that this would not necessarily work for teams which work on proprietary code-bases, but I let @srawlins to review this. 

I don't think this meaningfully improves security (if that's the angle you are coming from).

srawlins:
I agree with @mraleph ; if the main issue is that the code doesn't agree with the documented supported uses, we just need to update the documentation.

adilburaksen:
Thanks both — that is fair, and I think the PR as written (restricting sources) was the wrong lever: you are right that it would break private/proprietary plugins and that source restriction is not really the security boundary. Let me reframe what I think the actual issue is, because I do not think documentation alone addresses it.

The concern is not *where* a plugin comes from — it is that `dart analyze` / `flutter analyze` execute project-specified plugin code with **no consent**, on commands that are otherwise pure static analysis. Unlike `dart run` / `dart test` (which run project code by design), `analyze` is the one command developers and CI run specifically *because* it is understood not to execute the analyzed project. A developer who `git clone`s an untrusted repo and runs `dart analyze`, or — more importantly — a CI job that runs `dart analyze` / `flutter analyze` on an untrusted pull request, executes the PR authors plugin code on their machine / CI runner (with its secrets), with no prompt. Analyze-on-PR is an extremely common CI gate precisely because it is assumed safe.

The reason I framed it as security is that the editor already treats this exact execution as trust-sensitive: the Dart-Code extension sets `capabilities.untrustedWorkspaces.supported: false`, so opening an untrusted folder defers the analysis server (and its plugins) until the user clicks "Trust". The CLI has no equivalent gate — so the same execution that requires explicit trust in the editor happens silently from the terminal and in CI.

So docs do not really cover the CI case (CI does not read docs), and I agree source-restriction is the wrong fix. A better fix would be a consent/trust gate on the CLI mirroring the editor — e.g. do not run project-configured plugins during `analyze` unless the workspace is trusted or an explicit opt-in flag is passed — which preserves private/proprietary plugin workflows for trusted projects while not silently executing untrusted ones. I am happy to redo this PR along those lines if that direction is acceptable. And I completely understand if the teams position is that `analyze` is in the same trust bucket as `run`/`build`; I just wanted to make sure the CI-on-untrusted-PR scenario and the editor asymmetry were on the table before that call.

mraleph:
> And I completely understand if the teams position is that analyze is in the same trust bucket as `run`/`build`

I think that's the current stance. 

@srawlins do you know if we ever documented our security model and decisions for analyzer plugins?

> I just wanted to make sure the CI-on-untrusted-PR scenario and the editor asymmetry were on the table before that call.

CI on PRs usually runs tests right after it run the analyzer - so gating analyzer plugins does not significantly change attack surface.

(FWIW I think IDE's "Do you trust this folder?" button is mostly a security theater - most people click it without thinking twice)

adilburaksen:
Makes sense — I'll defer to the team's threat model on this. You're right that source-restriction was the wrong lever (it breaks private/proprietary plugins and isn't the real boundary), and I take the point that a CI job running `analyze` typically runs `test` right afterward, so gating analyzer plugins doesn't meaningfully change the attack surface in the common case.

If a short note in the analyzer-plugin docs stating that `analyze` / `flutter analyze` execute project-configured plugin code (same trust bucket as `run` / `build`) would be useful for the CI-on-untrusted-PR audience, I'm happy to send that as a separate docs PR. Otherwise I'll close this. Thanks both for the thoughtful discussion.

srawlins:
> @srawlins do you know if we ever documented our security model and decisions for analyzer plugins?

No I don't think we have. I can ask a few folks where we might document this.

But yes it's sort of the same question as whether we document that running `pub get` and then `dart run` or `flutter run` means executing more-or-less arbitrary code on your development machine. 🤷  But expectations being different, it's probably not bad to add a sentence somewhere.

mraleph:
@srawlins I also noticed that git-dependencies (or in general dependency overrides for plugins) are not documented, so people don't realize they can actually write:

```
plugins:
  plugin_one: ^1.2.3
  dependency_overrides:
    some_package:
      git:
        url: https://github.com/dart-lang/some_package.git
        ref: main
```

srawlins:
Thanks for noticing, I've uploaded https://dart-review.googlesource.com/c/sdk/+/509544 to improve the docs.

mraleph:
@srawlins I have seen the reports that `git` dependencies are not actually working. Though I have not tested it myself. Do we have end-to-end integration tests covering `git` dependencies? I see that we do have some unit tests, but unclear if we have anything else.

Issue #42571, opened 2020-07-03
Process.run incorrect arg quoting on Windows


kjk:
```
> dart --version
Dart VM version: 2.9.0-19.0.dev.flutter-63cf56d925 (be) (Fri Jun 26 01:56:03 2020 +0000) on "windows_x64"
```

Run the following from Windows cmd-line (`cmd.exe` or `powershell`):
```
> start "https://blog.kowalczyk.info/?foo=bar&al=me"
```

This opens default web browser with url `https://blog.kowalczyk.info/?foo=bar&al=me`

This Dart program should do the same:
```dart
import 'dart:io';

final url = "https://blog.kowalczyk.info/?foo=bar&me=him";

Future<ProcessResult> launchURL(String url) {
  stdout.writeln("launchURL: " + url);
  return Process.run("start", [url], runInShell: true);
}

void main() {
  launchURL(url);
}
```

(name it `main.dart` and run with: `dart run main.dart`)

This opens a web browser with url:
* `https://blog.kowalczyk.info/?foo=bar` 
instead of:
* `https://blog.kowalczyk.info/?foo=bar&me=him`

My guess is that `Process.run` incorrectly escapes/quotes the cmd-line argument. If the argument has `&` in it, it should be quoted with `"`.

I found it while trying to do `github_graphql_client` Flutter code lab on Windows. 
It uses `package:url_launcher/url_launcher.dart` which doesn't support Windows so I tried to use `Process.run`. Unfortunately, this bug breaks GitHub OAuth callback (because the url has `&` in it and everything after `&` is stripped`)

kjk:
It looks like the code is https://github.com/dart-lang/sdk/blob/master/runtime/bin/process_win.cc#L419 and it doesn't seem to do any escaping.

For reference, you can copy the implementation from Go: https://github.com/golang/go/blob/50bd1c4d4eb4fac8ddeb5f063c099daccfb71b26/src/syscall/exec_windows.go#L86

zichangg:
@kjk Thanks for posting.
The problem is indeed what you said. The url should be enclosed within double quotes.
I'm not familiar with Go but [one thing](https://github.com/golang/go/blob/50bd1c4d4eb4fac8ddeb5f063c099daccfb71b26/src/syscall/exec_windows.go#L90) looks suspicious is that it appends a space to all non-empty arg. It will make sure urls get surrounded by double quotes.

[...]

PR #63531, opened 2026-06-05
Escape cmd.exe metacharacters in runInShell arguments on Windows

Sengtocxoen:
Escape cmd.exe metacharacters in runInShell arguments on Windows

When Process.run/Process.start is invoked with runInShell: true on Windows, _getShellArguments built the cmd.exe command line by appending the executable and each argument verbatim. cmd.exe parses its own metacharacters (& | < > ^ % ( ) !) before the C runtime argv parsing performed by _windowsArgumentEscape, so an argument such as "a & calc" caused cmd.exe to execute "calc" as a separate command.

This lets any application that passes user-controlled data as arguments to a shelled-out command on Windows be coerced into running arbitrary commands (CWE-78). The POSIX branch is unaffected because it already single-quotes the executable and every argument, making them inert.

Neutralize the Windows path symmetrically by prefixing each cmd.exe metacharacter with the '^' escape character. The escaping is done in a single pass so an embedded '^' becomes '^^' rather than being re-escaped.

### DIFF
--- a/sdk/lib/_internal/vm/bin/process_patch.dart
+++ b/sdk/lib/_internal/vm/bin/process_patch.dart
@@ -330,9 +330,9 @@ base class _ProcessImpl extends _ProcessImplNativeWrapper implements _Process {
     List<String> shellArguments = [];
     if (Platform.isWindows) {
       shellArguments.add('/c');
-      shellArguments.add(executable);
+      shellArguments.add(_escapeWindowsShellMetacharacters(executable));
       for (var arg in arguments) {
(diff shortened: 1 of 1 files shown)


mraleph:
Related issue https://github.com/dart-lang/sdk/issues/42571

I am concerned that by now changing this becomes a breaking change, which needs to go through breaking change process where we assess whether it is worth fixing it or not.

Sengtocxoen:
I understand that, please take your time about this

riyaz-raz (approved):

PR #63588, opened 2026-06-12
Offer convertToDeclaringParameter on the `this.` prefix of an initializing formal

### LINKED ISSUE #63565: convertToDeclaringParameter is only offered on the name of the parameter
The convertToDeclaringParameter assist is only offered on the name of the parameter. There's some value in expanding the range over which it's offered. The concrete ask was to include the `this.` for an initializing formal. We should decide whether the type, any modifiers, etc. should also be included.

I'm fairly opposed to offering it in the default value.

Varadraj75:
Previously the `convertToDeclaringParameter` assist was only offered when the cursor was inside the parameter's name token. This widens the trigger so it is also offered when the cursor falls anywhere within the `this.` prefix (the `this` keyword or the following period) of a `FieldFormalParameter`.

The applicability range is computed from the node's own `thisKeyword` and `period` tokens, so no offsets or strings are hardcoded. Behavior for super parameters, type annotations, modifiers (`required`, `covariant`, `final`), and default values is unchanged - these were explicitly discussed and excluded from scope.

New tests cover the cursor on `this`, on the `.`, and on the parameter name (regression), and confirm the assist is correctly *not* offered on the type annotation, on a `required` modifier, or in a default value expression.

This follows the scope agreed with @bwilkerson in #63565.

Fixes #63565

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/dart/convert_to_declaring_parameter.dart
+++ b/pkg/analysis_server/lib/src/services/correction/dart/convert_to_declaring_parameter.dart
@@ -50,8 +50,15 @@ class ConvertToDeclaringParameter extends ResolvedCorrectionProducer {
       return;
     }
 
-    if (!parameterName.sourceRange.contains(selectionOffset)) {
-      // The assist only applies if the name of the parameter is selected.
+    var inName = parameterName.sourceRange.contains(selectionOffset);
+    var inThisPrefix =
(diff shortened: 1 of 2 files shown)


riyaz-raz (approved):

bwilkerson:
Thank you for the review, but reviews done on Github have no effect on the status of the PR. All of the reviews are done on Gerrit.

Issue #63564, opened 2026-06-09
Terminology for non-primary constructors


lrhn:
I've seen non-primary constructors being referred to as "secondary constructor"s.

I don't think that's a very good name to use, and would recommend consistently using "non-primary constructor".

Being "secondary" is a relational property, a thing can only be secondary _to_ something else.
A plain non-primary constructor in a class with no primary constructor is not secondary to anything.

By itself "secondary constructor" is not informative.
Even if someone knows what a "primary constructor" is, because they've heard about the "primary constructors" feature, they can't be expected to guess that "secondary constructor" means constructor that is not a primary constructor.

In some of the uses I've seen, "secondary constructor" is actually used to refer to only "non-primary non-redirecting generative constructor", and not just any non-primary constructor. I'm not sure the term is being used consistently.

The Primary Constructors feature specification uses the word "secondary" exactly once:
> This can be seen as a motivation for the word primary because it makes all other generative constructors secondary in the sense that they depend on the primary one.

That's a commentary on how a class with a primary constructor cannot have any other non-redirecting generative constructor. 
It's only saying that the other generative constructors (necessarily redirecting unless it's an extension type) are "secondary" in relation to a _concrete_ primary constructor of the same class, that they have to defer to. They are not "secondary constructors" by themselves.

bwilkerson:
@conooi

lrhn:
We discussed it in the language meeting, and we recommend using "in-body constructor" to denote a constructor which is not a primary constructor. It's more direct than "non-primary constructor", and all constructor declarations are either primary constructors, or are declared in the body of the type-declaration.

kallentu:
We've updated the docs and the [analysis server](https://dart-review.googlesource.com/c/sdk/+/513340) to use the new terminology. Closing.

PR #63620, opened 2026-06-18
Use code style lints to choose constructor insertion point in ConvertToSecondaryConstructor

### LINKED ISSUE #63558: convertToInBodyConstructor doesn't use enabled lints to choose where to insert the constructor
This should take into account the enablement of the `sort_constructors_first` and `sort_unnamed_constructors_first` lint rules.

Varadraj75:
`ConvertToSecondaryConstructor` now takes `sort_constructors_first` and `sort_unnamed_constructors_first` into account when deciding where to insert the new constructor, instead of always inserting before the first existing constructor.

`DartFileEditBuilder.insertConstructor` gains an optional `isUnnamed` parameter so the general utility can also place unnamed constructors before named ones when the lint is enabled. The parameter defaults to `false`, so existing callers are unaffected. `CodeStyleOptions` gains a `sortUnnamedConstructorsFirst` getter following the same `_isLintEnabled` pattern as the existing `sortConstructorsFirst`.

The assist's own `_offsetForConstructor` couldn't be fully replaced by `insertConstructor` because it also handles field insertion in the same pass, sometimes combining the field and constructor edits into a single insertion when their offsets coincide. Instead it now performs the same `CodeStyleOptions` checks and the same last-matching-member lookup, so
behavior stays consistent with the general utility.

Includes a regression test (`test_lint_sortConstructorsFirst_unnamedPrimary_noConflict`) specifically checking the case @bwilkerson asked about, whether the field and constructor insertions can target the same offset and cause a silent failure to compute an edit. They don't: when the offsets coincide, the existing code already merges them into a single `addInsertion` call rather than issuing two separate ones at the same location.

Fixes #63558

### DIFF
--- a/pkg/analysis_server/lib/src/cider/rename.dart
+++ b/pkg/analysis_server/lib/src/cider/rename.dart
@@ -403,6 +403,7 @@ class CheckNameResponse {
     }
     var edit = await buildEditForInsertedConstructor(
       node,
+      isNamed: newName.isNotEmpty,
       resolvedUnit: resolvedUnit,
       session: fileResolver.contextObjects!.analysisSession,
       (builder) => builder.writeConstructorDeclaration(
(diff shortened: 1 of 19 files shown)


eernstg:
About the terminology: Please note that the phrase 'secondary constructor' is misleading. A constructor which is not primary can be described as an 'in-body' constructor. See also https://github.com/dart-lang/sdk/issues/63564.

bwilkerson:
> About the terminology ...

Yes. And all of the references to 'secondary constructor' in either the code or the docs have been replaced with 'in-body constructor'. I think we've even fixed the Github issues.

PR #63653, opened 2026-06-24
[dart:io] set close-on-exec on all received socket message fds

ubeddulla:
1. a SCM_RIGHTS control message can carry several descriptors (`SocketControlMessage.fromHandles` takes a list), but the receive loop in `SocketBase::ReceiveMessage` only reads the first one and sets close-on-exec on it.
2. on platforms without `MSG_CMSG_CLOEXEC` (macOS/BSD) the remaining descriptors keep their default state, so they leak into child processes started after the message is received.

Loop over every descriptor in the message and set close-on-exec on each, which is what `MSG_CMSG_CLOEXEC` already does for us on Linux.

### DIFF
--- a/runtime/bin/socket_base_posix.cc
+++ b/runtime/bin/socket_base_posix.cc
@@ -163,14 +163,17 @@ intptr_t SocketBase::ReceiveMessage(intptr_t fd,
     new (control_message) SocketControlMessage(
         cmsg->cmsg_level, cmsg->cmsg_type, copied_data, data_length);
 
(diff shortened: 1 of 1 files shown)


ubeddulla:
Done — switched to reading the fds through `const int* fds = reinterpret_cast<int*>(copied_data)` and indexing, so the per-fd memmove is gone.

Left the drop-the-rest behavior alone as you suggested, since that's pre-existing and this change doesn't make it any worse.

PR #63663, opened 2026-06-24
[analysis_server] Use null activeParameter when supported

### LINKED ISSUE #63651: Update LSP SignatureHelp activeParameterIndex to use `null` when supported by the client
The original version of LSP required an active parameter index that we don't use, resulted in this hack:

https://github.com/dart-lang/sdk/blob/28d7327f71bdf326476e9bad44abcd0b23944002/pkg/analysis_server/lib/src/computer/computer_type_arguments_signature.dart#L106-L110

But we now have a solution to this in https://github.com/microsoft/language-server-protocol/pull/1597, so if the client supports `null` we should use that.

Apple-beep:
Fixes #63651.

This updates LSP signature help so that when a client advertises `textDocument.signatureHelp.signatureInformation.noActiveParameterSupport`, the server can return an explicit `activeParameter: null` instead of using an out-of-bounds parameter index.

Changes included:

* Reads `noActiveParameterSupport` from client capabilities.
* Applies the behavior to both regular signature help and type-argument signature help.
* Preserves the existing out-of-bounds fallback for clients that do not support null active parameters.
* Adds LSP tests that assert the raw JSON response includes `activeParameter: null`.

Testing:

* `git diff --check`
* `python3 tools/test.py -n analyzer-unittest-asserts-release-mac pkg/analysis_server/test/lsp/signature_help_test`

  * Could not run locally because `xcodebuild/ReleaseARM64/dart-sdk/bin/dart` is missing in this checkout.

### DIFF
--- a/pkg/analysis_server/lib/src/cider/signature_help.dart
+++ b/pkg/analysis_server/lib/src/cider/signature_help.dart
@@ -31,6 +31,7 @@ class CiderSignatureHelpComputer {
       resolvedUnit.unit,
       offset,
       formats,
+      noActiveParameterSupport: false,
     );
     if (typeArgsComputer.offsetIsValid) {
       var typeSignature = typeArgsComputer.compute();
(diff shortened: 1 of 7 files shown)


DanTup:
@Apple-beep thanks for looking at this, but there's an existing change open for it at https://dart-review.googlesource.com/c/sdk/+/516923

bwilkerson:
The convention the Dart SDK repository uses is that issues are left unassigned until someone is working on it. When work starts the 'assigned to' field is updated. I don't know whether the issue was already assigned when you started working on it, or whether this was an unfortunate overlap in efforts.

If you can't assign an issue to yourself, then you can add a comment letting us know that you're planning on working on it. Either way we'll know and hopefully this won't happen to you in the future.

Issue #63489, opened 2026-06-01
[hooks] Support code assets in `dart test -c cli --target-sanitizer`


dcharkes:
Follow up of https://github.com/dart-lang/sdk/issues/63372.

> * We'll add a `dart test -c cli` that will use `dart build cli` under the hood. It will not properly support sanitizers for now.

We'll use sanitized dylibs from build hooks, and a sanitized aotruntime.

However, since the AOT snapshot is appended, none of the compiled Dart symbols will show up in sanitizer error messages.

To support sanitizers and build hooks together, we'll need to:

* https://github.com/dart-lang/sdk/issues/63435
* And then wire up the `dart test -c cli` to accept the `--target-sanitizer` flag.

Use case:

* https://github.com/google/webcrypto.dart/issues/278

HamdaanAliQuatil:
@dcharkes #63435 is now in, so I’d like to take this next.

I checked the current `package:test` implementation. `_compileToCli` already invokes dart build cli and maps vm-asan, vm-msan,  and vm-tsan to the corresponding --target-sanitizer argument. On Linux, the SDK now includes the matching dartcliruntime_* artifacts.

The remaining work looks like:
- allow Compiler.cli for vm-asan, vm-msan, and vm-tsan, while keeping exe as the default;
- pass the existing sanitizer environment defaults when launching a CLI bundle;
- update the compiler/runtime tests to check for the corresponding dartcliruntime_* artifact;
- add -c cli coverage for the sanitizer runtimes.

I think the user-facing command should remain `dart test -c cli -p vm-asan` (and similarly for MSan/TSan), with `package:test` translating that internally to `dart build cli --target-sanitizer=....`

I’m planning to keep the compiler/runtime coverage in dart-lang/test. Once that revision rolls into the SDK, I’ll add a full hook-built end-to-end case to the SDK’s native-assets tests, since that suite already has the relevant fixtures and dependency overrides. Let me know if you see a reason to keep that coverage in dart-lang/test instead.

dcharkes:
> * pass the existing sanitizer environment defaults when launching a CLI bundle;

What does this mean?

Yeah we need to set up the plumbing and then add then adding a test in the SDK sgtm.

You might want to to set up the plumbing in `package:test` and then write the test in the SDK immediately and use a git override in the package config (or a DEPS test_rev update + a path dependency to the pulled in package test) to ensure everything works before landing `package:test` support.

[...]

PR #63690, opened 2026-06-28
[dartdev] Add shared runtime for Linux CLI bundles

### LINKED ISSUE #63435: [dartdev] `dart build cli` support a separate aotruntime mode?
We want to introduce build hooks into `dart test -c exe` (https://github.com/dart-lang/sdk/issues/63372).

However, if we build it on top of `dart build cli`, we would use `dart compile exe` under the hood.

This would lose us native profilers and symbolizers support

```dart
  /// Compiles [path] to a native shared library using
  /// `dart compile aot-snapshot`.
  ///
  /// Preferable to `dart compile exe` because embedded snapshots are invisible
  /// to native profilers and symbolizers. This should eventually be replaced
  /// with something that supports build hooks.
```

https://github.com/dart-lang/test/blob/fe9e65866919d28364f2c1bdd806a0a3d5cb101a/pkgs/test_core/lib/src/runner/vm/platform.dart#L248-L253

So, we should consider adding a "mode" to `dart build cli` that keeps the aot-snapshot separate from the dartaotruntime and generates a shim executable or shell script that makes the `bin/my_app` run `../bin_private/dartaotruntime ../bin_private/my_aot_snapshot`.

Having such a setup would also help with keeping app-bundles with multiple executables smaller, because we'd only include a dartaotruntime in the bundle once.

(N.b. for `dart test -c exe` it is unnecessary to copy the dartaotruntime, because we have access to a dart sdk, so we can just use that. However, I'd also like to avoid exposing a new `dart build` subcommand or new `dart build cli` flags that are only used in `package:test`. If `package:test` works as an external package to the SDK it should only rely on well-tested public Dart CLI features. cc @jakemac53 @bkonyi)

@rmacnak-google If you have any other bright ideas, let me know!

HamdaanAliQuatil:
Fixes #63435

`dart build cli` currently produces a separate executable with an appended AOT snapshot for each entrypoint. On Linux, this keeps the snapshot opaque to native symbolizers and duplicates the runtime in bundles with multiple executables.

This implements the BusyBox-style layout discussed in the issue:

- adds a Linux `dartcliruntime` that resolves `../lib/dartaotsnapshot<executable-name>` from `argv[0]`
- falls back to the resolved executable name so invoking a bundle executable through an external symlink still works
- distributes normal and ASan/MSan/TSan variants of `dartcliruntime` in the SDK
- changes Linux `dart build cli` bundles to contain one runtime, symlink additional entrypoints to it, and keep each AOT snapshot under `bundle/lib/`
- leaves the existing bundle behavior unchanged on other platforms

The native-assets tests cover the new layout, execution through symlinks, custom entrypoints, and sanitizer runtimes.

Tested with:

```text
python3 tools/test.py -n unittest-asserts-release-linux-x64 pkg/dartdev/test/native_assets/build_test.dart
```

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -51,6 +51,12 @@ group("runtime") {
       "utils/dtd:dtd_aot",
       "utils/kernel-service:frontend_server_aot_product",
     ]
+    if (is_linux) {
+      deps += [
+        "runtime/bin:dartcliruntime",
+        "runtime/bin:dartcliruntime_product",
+      ]
+    }
   } else {
     deps += [
       "utils/dart_runtime_service_vm",
@@ -111,6 +117,9 @@ group("runtime_precompiled") {
     "runtime/bin:process_test",
     "runtime/vm:kernel_platform_files($host_toolchain)",
   ]
+  if (is_linux) {
+    deps += [ "runtime/bin:dartcliruntime" ]
+  }
   if (is_linux || is_android) {
     deps += [ "runtime/bin:abstract_socket_test" ]
   }
--- a/pkg/dart2native/lib/sdk.dart
+++ b/pkg/dart2native/lib/sdk.dart
@@ -52,6 +52,18 @@ class Sdk {
 
   String get dartAotRuntime => dartAotRuntimeFor();
 
+  String dartCliRuntimeFor({
+    String? sanitizer,
+  }) {
+    final name = sanitizer != null && sanitizer != 'none'
+        ? 'dartcliruntime_$sanitizer'
+        : 'dartcliruntime';
+    return _executablePathFor(
+      name,
+      forceProductInBuildRoot: true,
+    );
+  }
+
   String get genSnapshot => _executablePathFor(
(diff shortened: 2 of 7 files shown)


dcharkes:
@HamdaanAliQuatil Could you please rebase your CL. Then I can trigger the CI again.

HamdaanAliQuatil:
Rebased and pushed a new patchset. I also updated the snapshot filename to `libdartaot$executableName.so` to match the shared-library naming discussion.

HamdaanAliQuatil:
@dcharkes All 27 tryjobs are green now and all review threads are resolved. Is there anything else you’d like changed, or is this ready for approval?

If this lands, I’d be happy to take #63489 next. _compileToCli already forwards the sanitizer to dart build cli; the remaining package:test gap looks like enabling Compiler.cli for vm-asan, vm-msan, and vm-tsan, then updating the runtime matrix and SDK artifact checks. Is that the follow-up you had in mind?

PR #63706, opened 2026-06-30
bound netlink address copy to destination size in getifaddrs

ubeddulla:
SetAddresses in the Android getifaddrs fallback copies RTA_PAYLOAD bytes from a netlink IFA_ADDRESS attribute straight into the 4-byte sin_addr or 16-byte sin6_addr, and RTA_OK only bounds the attribute against the read buffer, so a malformed RTM_NEWADDR with an oversized address attribute overflows the freshly allocated sockaddr on the heap. Cap the copy at the destination field size in both branches, matching the length check GetSockAddr already does for the same address bytes.

### DIFF
--- a/runtime/bin/ifaddrs.cc
+++ b/runtime/bin/ifaddrs.cc
@@ -42,32 +42,33 @@ static void SetFlags(struct ifaddrs* ifaddr, int flag) {
   ifaddr->ifa_flags = flag;
 }
 
-static void SetAddresses(struct ifaddrs* ifaddr,
-                         int family,
-                         int index,
-                         void* data,
-                         size_t len) {
+DART_WARN_UNUSED_RESULT static bool SetAddresses(struct ifaddrs* ifaddr,
+                                                 int family,
+                                                 int index,
+                                                 void* data,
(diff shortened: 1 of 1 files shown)


ubeddulla:
Good point, rejecting the message is cleaner than silently truncating. I've switched SetAddresses to DART_WARN_UNUSED_RESULT bool and made it return false when the attribute length doesn't match sin_addr/sin6_addr, then propagated that through SetIfAddrsFromAddrMsg and SetIfAddrsFromInfoMsg. The getifaddrs loop already bails out to FailAndExit when those return false, so a malformed entry now aborts the parse instead of being copied. Pushed.

ubeddulla:
The reject-not-cap change from Slava's comment is in the current patch set: SetAddresses is now DART_WARN_UNUSED_RESULT bool and returns false when the attribute length doesn't match sin_addr/sin6_addr, and both SetIfAddrsFromAddrMsg and SetIfAddrsFromInfoMsg propagate that up so a malformed entry aborts the parse instead of being copied. Should be ready for another look.

PR #63708, opened 2026-06-30
 Linter rule cascade_invocations Quick Fix error #63354

### LINKED ISSUE #63354: Linter rule cascade_invocations Quick Fix error
With the following code and the `cascade_invocations` rule enabled, running the Quick Fix changes the program result (prints `{1: C(1)}` instead of `{1: C(3)}`.

```dart
void main(List<String> args) {
  var a = {1: C(1)};
  var b = a[1] ??= C(2);
  b.i = 3;
  print(a); // {1: C(3)}
}

class C {
  C(this.i);
  int i = 0;
  @override
  String toString() => 'C($i)';
}
```

The conversion result is `var b = a[1] ??= C(2)..i = 3;`
But it should be `var b = (a[1] ??= C(2))..i = 3;`

ayushgupta9906:
Fix `cascade_invocations` quick fix generating semantics-breaking code for assignment initializers

Fixes #63354

### Problem

When the `cascade_invocations` linter rule triggered on code like:

```dart
var b = a[1] ??= C(2);
b.i = 3;
```

The **Convert to cascade** quick fix produced:

```dart
var b = a[1] ??= C(2)..i = 3;  // ❌ Wrong
```

This silently changes program behavior. The cascade operator `..` has **higher precedence** than assignment operators (`??=`, `=`, `+=`), so Dart actually parses this as:

```dart
var b = a[1] ??= (C(2)..i = 3);
```

The mutation `..i = 3` only applies to the temporary `C(2)` object and never updates the value stored in `a[1]`, printing `{1: C(1)}` instead of the expected `{1: C(3)}`.

### Fix

When the variable's initializer is an `AssignmentExpression`, wrap it in parentheses before appending the cascade:

```dart
var b = (a[1] ??= C(2))..i = 3;  // ✅ Correct
```

### Changes

- `pkg/analysis_server/lib/src/services/correction/dart/convert_to_cascade.dart` — Detect `AssignmentExpression` initializers in variable declarations and insert surrounding parentheses in the generated edit.
- `pkg/analysis_server/test/src/services/correction/fix/convert_to_cascade_test.dart` — Add regression test `test_declaration_assignmentInitializer_property` for the exact scenario from the bug report.


ayushgupta9906 (commented):
update the one small block of code (formatting)

ayushgupta9906 (commented):
I’ve added the missing AssignmentExpression? initializerToParenthesize declaration in the appropriate scope.

mraleph:
@ayushgupta9906 please address build failures or close this PR.

PR #63734, opened 2026-07-02
Feature/native api lint ffigen

### LINKED ISSUE #63618: [analyzer,interop] Custom lint for OS-level API compatibility for generated bindings to native APIs
iOS/MacOS APIs have os-versions of from when to when an API is available.

We could make FFIgen generate some kind of annotation on the Dart methods that translates this info and then the interop bindings can then be published as a package.

This package can then be used in a Flutter app.
This Flutter app targets a specific version of iOS.

We'd want lint errors on calling methods that are not available on that iOS version.

We could do this via a custom analyzer plugin.

But that analyzer plugin would need access to the file system to see what the iOS version is that the app is targeting.

Or alternatively, if analyzer plugins could have parameters, we'd want Flutter to somehow configure that parameter.

@srawlins do plugins have access to the file system? can plugins have configurable parameters?

cc @liamappelbe @bparrishmines @stuartmorgan-g

ayushgupta9906:
# Custom lint for OS-level API compatibility

Fixes #63618 

## Description
This PR introduces a custom analyzer plugin and annotation package to provide warnings and errors when generated `ffigen` bindings are incompatible with the user's project deployment targets. 

iOS/macOS APIs have OS-versions defining when an API was introduced, deprecated, or obsoleted. When `ffigen` generates bindings for these APIs, developers need analyzer warnings if they call an API that requires a newer OS version than what their app currently targets.

### Changes Included:
* **`package:native_interop_annotation`**: Adds the `@ExternalVersions` annotation and `ExternalVersion` class. This allows `ffigen` to emit standard annotations specifying `min`, `max`, and `deprecationMessage` onto Dart elements.
* **`package:native_api_lint`**: A standalone analyzer plugin hooked into the Dart analysis server that:
  * Dynamically reads the active workspace's deployment target (`analysis_options.yaml`, `ios/Podfile`, or `ios/Runner.xcodeproj/project.pbxproj`).
  * Emits `api_not_available_on_min_target` (WARNING) if the API requires a newer OS.
  * Emits `api_obsoleted_on_min_target` (ERROR) if the API was removed before the project's minimum target.
  * Emits `api_deprecated_on_target` (INFO) if the API is deprecated on the user's targeted OS version.
* Added comprehensive unit testing with 100% pass rates across `DeploymentTargetResolver`, `AnnotationReader`, and `VersionUtils`.

## Testing
- [x] Tested locally against the internal Dart SDK analyzer models.
- [x] Validated `Podfile` and `pbxproj` parsing with single quotes, double quotes, and leading spaces.
- [x] Verified graceful fallbacks when Xcode build configurations are missing or malformed.


mraleph:
As indicated by @dcharkes this is the wrong place for this lint.

PR #63789, opened 2026-07-09
fix unsigned readlink return in File::LinkTarget on macos

ubeddulla:
File::LinkTarget on macOS stores the readlink() result in a size_t, so the `target_size <= 0` check only rejects an empty read and lets a -1 error return slip through as SIZE_MAX. If the symlink is removed or replaced between the lstat/S_ISLNK check and readlink (a race on a link created by another process, reached through Link.target()), the following ScopedCString(target_size + 1) wraps to a zero-length allocation and memmove copies SIZE_MAX bytes. The Linux and Fuchsia versions store it in an int, so their identical check catches the failure; this changes the macOS variable to match.


mraleph:
Note that there are comments on the CL. Please address them.

ubeddulla:
Good point, switched it to ssize_t so it matches readlink's actual return type. That signed type is the real fix: with size_t the `target_size <= 0` check couldn't catch the -1. The only reason I also touched the second comparison is that the old code cast dest_size to size_t there, which would turn into a signed/unsigned compare once target_size is signed, so I dropped the now-redundant cast (dest_size is already int). Pushed.

mraleph:
You should reopen and push again. Closed PRs don't get synced in Gerrit. The same applies to few other PRs you have made and then accidentally closed. (I assume it's your agentic setup that does this for whatever reason)

ubeddulla:
Tried to reopen this but GitHub won't let me: the original push landed as an orphan commit with no shared history against main, and once I recreated the branch on top of current main it counted as a force-push, so reopen is blocked either way. I opened #64069 with the same fix (now cleanly based on main) to get it back into Gerrit. Sorry for the churn.

PR #63859, opened 2026-07-21
read the zlib dictionary length once in CopyDictionary

ubeddulla:
`CopyDictionary` in `runtime/bin/filter.cc` sizes the dictionary heap buffer from one `Dart_ListLength` call, and `Filter_CreateZLibInflate` / `Filter_CreateZLibDeflate` then call `Dart_ListLength` on the same object a second time to get the length they hand to the filter. For a user-defined class implementing `List<int>` that call dispatches to the Dart `length` getter, so the two reads can disagree, and a getter that returns a bigger value the second time leaves `dictionary_length_` pointing past the end of the allocation that `inflateSetDictionary` / `deflateSetDictionary` later read from. `ZLibCodec`, `ZLibEncoder` and `ZLibDecoder` all pass the caller's `List<int>` straight through to these natives, so no copy happens on the Dart side first.

Having `CopyDictionary` report the number of bytes it actually copied keeps the buffer and the length in sync and drops the second lookup at both call sites.

---

<details>

</details>

### DIFF
--- a/runtime/bin/filter.cc
+++ b/runtime/bin/filter.cc
@@ -33,8 +33,10 @@ static Dart_Handle GetFilter(Dart_Handle filter_obj, Filter** filter) {
 }
 
 static Dart_Handle CopyDictionary(Dart_Handle dictionary_obj,
-                                  uint8_t** dictionary) {
+                                  uint8_t** dictionary,
+                                  intptr_t* dictionary_length) {
   ASSERT(dictionary != nullptr);
+  ASSERT(dictionary_length != nullptr);
   uint8_t* src = nullptr;
   intptr_t size;
   Dart_TypedData_Type type;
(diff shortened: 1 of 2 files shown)


ubeddulla:
any update?

ubeddulla:
Pushed an update for both comments.

The dictionary is now a `Filter::Dictionary` struct (a `std::unique_ptr<uint8_t[]>` plus its length) that gets moved into the filter, so the buffer and length are always in sync and the manual `delete[]` calls are gone. I put the struct on the base `Filter` so both the inflate and deflate filters can share it.

Also dropped the hand-rolled `Dart_TypedDataAcquireData` fast path in `CopyDictionary` and just call `Dart_ListGetAsBytes`, since it already handles the typed-data case internally.

mraleph:
@ubeddulla  Please address comments on the CL or close the PR.

PR #63500, opened 2026-06-02
Fix HTTP response splitting via unvalidated reason phrase in dart:io

Sengtocxoen:
## Problem

`HttpResponse.reasonPhrase` in `dart:io` accepts CRLF characters without validation. Every other HTTP wire-output API in the SDK validates against CRLF via `_isValueChar()` / `_validateValue()`. The `reasonPhrase` setter is the single gap.

An attacker who controls the value passed to `reasonPhrase` can inject CRLF sequences to split the HTTP response, injecting arbitrary headers and response bodies.

### API consistency

```dart
// Headers: VALIDATED — throws FormatException on CRLF
response.headers.set('X-Test', 'value\r\nInjected'); // FormatException

// Reason phrase: NOT VALIDATED before this fix
response.reasonPhrase = 'OK\r\nInjected: header\r\n\r\n<script>alert(1)</script>'; // silently accepted
```

## Fix

Adds CRLF validation to the `reasonPhrase` setter using the existing `_HttpHeaders._isValidValueString()` function — the same validation already applied to **all** header values in `http_headers.dart`.

```diff
 void set reasonPhrase(String reasonPhrase) {
   if (_outgoing.headersWritten) throw StateError("Header already sent");
+  var errorAt = _HttpHeaders._isValidValueString(reasonPhrase);
+  if (errorAt >= 0) {
+    throw FormatException(
+      "Invalid HTTP reason phrase",
+      reasonPhrase,
+      errorAt,
+    );
+  }
   _reasonPhrase = reasonPhrase;
 }
```

Zero new validation infrastructure — reuses the existing `_isValidValueString()` that rejects bytes <= 31 (CR, LF, NUL) except HT.

## Impact

- **Header injection:** Set-Cookie, Access-Control-Allow-Origin, CSP override
- **Response body injection / XSS:** CRLF + CRLF terminates headers, body controlled by attacker
- **Cache poisoning:** Inject a complete second HTTP response into the stream

## Validation

`headers.set('X', 'a\r\nb')` → `FormatException` (existing, correct behavior)  
`reasonPhrase = 'OK\r\n...'` → now also throws `FormatException` (new, consistent)

PR #63907, opened 2026-07-29
reject CR, LF and NUL in HttpResponse.reasonPhrase

ubeddulla:
1. every user-supplied field written to a response gets validated except the reason phrase: header names go through _validateField, header values through _validateValue, a redirect Location through headers.set, and cookie name/value at construction.
2. _writeHeader emits reasonPhrase.codeUnits straight into the status line before the CRLF, so a phrase like "OK\r\nX-Injected: 1" adds an extra header line (response splitting).

The setter now rejects CR, LF and NUL with an ArgumentError, matching the same CR/LF/NUL rule the parser already applies to header values. Regression test added under tests/standalone/io.


mraleph:
This is duplicate of https://github.com/dart-lang/sdk/pull/63500

Issue #63635, opened 2026-06-22
Override inference issue for non-final variable


sgrekhov:
The code below produce an error in VM but no issues in the analyzer.
```dart
class A {
  void set m(int _) {}
  num get m => 0;
}

class C extends A {
  var m = 0; // Analyzer shows that `m` is inferred as `int`
}
```
VM output `Error: Can't infer a type for 'm' as the overridden members don't have a combined signature.
Try adding an explicit type.`

According to the [type inference specification](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md#instance-field-getter-and-setter-override-inference) 

> The type of a non-final field which overrides/implements both a setter and a getter is inferred to be the parameter type of the combined member signature of said setter in the direct superinterfaces, if this type is the same as the return type of the combined member signature of said getter in the direct superinterfaces. If the types are not the same then inference fails with an error. 

That means that the VM is right here. But, probably, this part of the spec is obsolete.

cc @eernstg to confirm the expected result.

eernstg:
I think the specification is still OK, even after https://github.com/dart-lang/sdk/issues/58578. We rely on the ability to compute a combined member signature for the getter and for the setter, separately, and then we verify that the return type of the combined getter signature and the parameter type of the combined setter signature is the same type, and only then do we override-infer the type of the variable. If the types are different then we just report an error.

I don't see a reason to change this at this time. We've relaxed the checks on the declarations such that those two types don't have to be subtype related. However, this just adds a bunch of additional error cases, and I don't think any of those can reasonably be non-errors given that they are "more off track" than the ones that we already reject.

In any case, the VM follows the current specification, and the analyzer should report an error as well. Also, the error message probably shouldn't say that the "overridden members don't have a combined signature"&mdash;a getter and a setter never has a combined member signature, that's only a defined concept for the getter separately and the setter separately. It could perhaps say that the "combined member signature of the getter does not agree with the combined member signature of the setter".

johnniwinther:
cc @scheglov

Yusufihsangorgel:
If the fix lands on the analyzer side per the discussion, I can take it: report the missing error when the combined getter and setter signatures disagree, with diagnostics tests. Adjacent to the ffi_verifier change I have in review (525581), same area.

scheglov:
Should we use structural, or semantic equality?

Is this code OK, or error?
CFE says that it is an error.
```dart
import 'dart:async';

abstract class A {
  Object get foo;
[...]

PR #63945, opened 2026-08-01
[analyzer] Normalize inherited getter and setter types before comparing

Yusufihsangorgel:
The check compared the combined getter and setter types with `==`, which
is structural. The notion of "same type" used elsewhere in the language
normalizes `FutureOr`, so `NORM(FutureOr<Object>)` is `Object` and a
getter typed `Object` agrees with a setter typed `FutureOr<Object>`.
`dynamic` and `Object?` normalize to themselves and keep reporting.

This is the analyzer half of what `@eernstg` described in the issue, on
the example `@scheglov` asked about there.

Two tests. `Object` against `FutureOr<Object>` was red before this change
and is green after; `dynamic` against `Object?` pins the behavior that
must not move.

The CFE half is still open. `inferFieldType` in `members_node.dart`
compares the same two types with kernel `==`, so it reports on the example
today and the two disagree until it changes as well. Happy to follow it
into the front end if that is the order you want.

Bug: https://github.com/dart-lang/sdk/issues/63635
TEST=pkg/analyzer/test/src/diagnostics/different_inherited_getter_and_setter_types_test.dart

---

<details>

</details>

### DIFF
--- a/pkg/analyzer/lib/src/dart/element/type_system.dart
+++ b/pkg/analyzer/lib/src/dart/element/type_system.dart
@@ -137,6 +137,18 @@ class TypeSystemImpl implements TypeSystem {
     return t is FunctionType || t.isDartCoreFunction;
   }
 
+  /// Whether the normal forms of [T1] and [T2] are structurally equal.
+  ///
+  /// Normalization rewrites `FutureOr<Object>` to `Object`, so a getter typed
+  /// `Object` and a setter typed `FutureOr<Object>` compare equal here.
+  /// `dynamic` and `Object?` normalize to themselves.
+  ///
+  /// https://github.com/dart-lang/language
+  /// See `resources/type-system/normalization.md`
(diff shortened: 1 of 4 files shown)


eernstg:
Hi @Yusufihsangorgel, thanks for contributing a substantial amount of work!

However, as I wrote [here](https://github.com/dart-lang/sdk/issues/63635#issuecomment-5201725146), we must approach this topic very cautiously because it changes existing behavior at a rather fundamental level of the language. The language team will have to clarify the situation first, and we might need to refine and clarify the rules in this area. So I'm afraid this PR should not be landed at this time.

I hope you don't think we're wasting your time by putting your work on hold like that, but it wouldn't be good if we make changes to the implementation in this area, and it turns out to be slightly different from what we actually want in the end.

Yusufihsangorgel:
Not at all, and I've withdrawn the front-end offer over on the issue. The branch stays where it is.

mraleph:
I am going to close this PR based on @eernstg comment above.

PR #63987, opened 2026-08-06
[dartdev] Test CLI sanitizers with native assets

### LINKED ISSUE #63489: [hooks] Support code assets in `dart test -c cli --target-sanitizer`

HamdaanAliQuatil:
Fixes #63489.

The CLI sanitizer runtime and `package:test` support have landed, but the SDK did not yet exercise the complete path with a hook-built native asset.

This adds Linux integration coverage using the existing `native_add` fixture. ASan, MSan, and TSan each run `dart test -c cli`, check that the build hook ran, and verify that the native test completed successfully. The ASan coverage also injects a heap buffer overflow into a temporary copy of the native source and verifies that AddressSanitizer reports it.

The copied fixture uses the `test`, `test_api`, and `test_core` packages pinned in the SDK checkout, so the integration test covers the SDK's actual dependency revision. Sanitizer tests are skipped on non-Linux hosts; on supported Linux configurations, a missing CLI runtime is treated as a test failure rather than silently skipped.

Tested with:

```text
python3 tools/test.py -n unittest-asserts-release-linux-x64 pkg/dartdev/test/native_assets/test_test.dart
```

### DIFF
--- a/pkg/dartdev/test/native_assets/helpers.dart
+++ b/pkg/dartdev/test/native_assets/helpers.dart
@@ -300,6 +300,12 @@ Future<void> copyTestProjects(
       'meta',
     ])
       package: {'path': sdkRoot.resolve('pkg/$package/').toFilePath()},
+    for (final package in ['test', 'test_api', 'test_core'])
+      package: {
+        'path': sdkRoot
+            .resolve('third_party/pkg/test/pkgs/$package/')
+            .toFilePath(),
+      },
   };
   final userDefinesWorkspace = {};
   for (final pubspecPath in pubspecPaths) {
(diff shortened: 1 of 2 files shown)


HamdaanAliQuatil:
Continuing this change directly in Gerrit at https://dart-review.googlesource.com/c/sdk/+/533480 so I can update patchsets and trybot metadata without going through Copybara. The direct CL includes the review changes from this PR.

PR #64032, opened 2026-08-12
[analysis_server] Offer "create constructor" fix for non-final uninitialized fields

### LINKED ISSUE #47389: [analyzer] add a fix for a generative constructor for `NOT_INITIALIZED_NON_NULLABLE_INSTANCE_FIELD`
For [NOT_INITIALIZED_NON_NULLABLE_INSTANCE_FIELD](https://github.com/dart-lang/sdk/blob/9836909f3d82ba7534b45dd0fa0d817ed0905ae1/pkg/analyzer/lib/src/error/codes.g.dart#L10711) (and possibly [NOT_INITIALIZED_NON_NULLABLE_INSTANCE_FIELD_CONSTRUCTOR](https://github.com/dart-lang/sdk/blob/9836909f3d82ba7534b45dd0fa0d817ed0905ae1/pkg/analyzer/lib/src/error/codes.g.dart#L10724)).

The correction message is
> Try adding an initializer expression, or a generative constructor that initializes it, or mark it 'late'.

Currently, `NOT_INITIALIZED_NON_NULLABLE_INSTANCE_FIELD` is associated with `AddLate`.

It would be better to also have something like "AddConctructor" or "AddToConstructor".

hkarmoush:
Fixes #47389

If you write `int a;` with no initializer, analyzer only offers to fix it with "Add late". But if you write `final int a;` instead, you also get a "create a constructor for me" quick-fix — that option just never got extended to non-final fields, even though nothing about the underlying logic actually required `final`.

This reuses the same constructor-generating producer for the non-final case. As a bonus, it also handles classes that mix final and non-final uninitialized fields in one shot — previously you'd only get a fix that covered the final ones.

Two things needed guarding since non-final fields behave differently from final ones:
- A nullable non-final field (`int? b;`) is already valid Dart without a constructor, so those are left alone.
- Flutter widget constructors are generated as `const`, but `const` requires every field to be final — that gets dropped automatically when a non-final field is involved.

## Test plan
- Added 7 new test cases to `create_constructor_for_final_fields_test.dart` covering: a single non-final field, mixed final/non-final fields, private fields, late-field exclusion, nullable-non-final fields being left untouched, and the Flutter widget case (confirming `const` is correctly dropped).
- Full `pkg/analysis_server/test/src/services/correction/fix/` suite (4762 tests) passes with no regressions.
- `dart format` / `dart analyze` clean.

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/dart/create_constructor_for_final_fields.dart
+++ b/pkg/analysis_server/lib/src/services/correction/dart/create_constructor_for_final_fields.dart
@@ -60,13 +60,15 @@ class CreateConstructorForFinalFields extends ResolvedCorrectionProducer {
     _FixContext fixContext;
     switch (container) {
       case ClassDeclaration():
-        superType = container.declaredFragment?.element.supertype;
+        var classElement = container.declaredFragment?.element;
+        superType = classElement?.supertype;
         if (superType == null) {
           return;
         }
         fixContext = _FixContext(
           builder: builder,
           containerName: container.namePart.typeName.lexeme,
+          classElement: classElement,
           superType: superType,
           variableLists: container.body.members.interestingVariableLists,
         );
@@ -78,6 +80,7 @@ class CreateConstructorForFinalFields extends ResolvedCorrectionProducer {
         fixContext = _FixContext(
           builder: builder,
           containerName: container.namePart.typeName.lexeme,
+          classElement: null,
           superType: superType,
           variableLists: container.body.members.interestingVariableLists,
         );
@@ -518,12 +521,19 @@ class _Field {
 class _FixContext {
   final ChangeBuilder builder;
   final String containerName;
+
+  /// The element of the class being fixed, or `null` if it's an enum.
+  ///
+  /// Enum instance fields must always be final, so [canBeConst] doesn't need
+  /// this to determine whether an enum's generated constructor can be const.
+  final ClassElement? classElement;
   final InterfaceType superType;
   final Iterable<VariableDeclarationList> variableLists;
 
   new({
     required this.builder,
     required this.containerName,
+    required this.classElement,
     required this.superType,
     required this.variableLists,
   });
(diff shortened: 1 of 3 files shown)


bwilkerson:
It appears that copybara can't copy your latest changes to Gerrit because of a CLA related issue. Let us know if you need help resolving the issue.

mraleph:
I flushed CLA check by forcing a rescan.

bwilkerson:
There are failures on the bots, but it doesn't look like that information is being copied here. Let me know if you need help finding or understanding the problem.

keertip:
@hkarmoush 

Unsorted file /b/s/w/ir/cache/builder/sdk/pkg/analysis_server/lib/src/services/correction/dart/create_constructor_for_final_fields.dart
  Unsorted file /b/s/w/ir/cache/builder/sdk/pkg/analysis_server/test/src/services/correction/fix/create_constructor_for_final_fields_test.dart


Please fix and upload changes.

mraleph:
@hkarmoush please address review comments or close the PR if you no longer interested in landing this.

hkarmoush:
Thanks for the ping @mraleph — still interested in landing this, sorry for the lag.

@keertip Fixed. I ran the SDK's own member sorter over both flagged files:

```
dart pkg/analysis_server/test/verify_sorted_test.dart --update \
  -s pkg/analysis_server/lib/src/services/correction/dart/create_constructor_for_final_fields.dart \
  -s pkg/analysis_server/test/src/services/correction/fix/create_constructor_for_final_fields_test.dart
```

`verify_sorted_test` now passes for both files, and `dart format --set-exit-if-changed` reports no changes.

The commit is pure member reordering — 181 insertions / 181 deletions, and the sorted multiset of lines is byte-identical before and after, so there is no behavior change.

One note for transparency: I wasn't able to run the full `create_constructor_for_final_fields_test` suite locally (no local SDK build available on this machine), so I'm relying on the bots for that — the sort itself is verified as above. @bwilkerson, re: the bot failures not being copied across — the presubmit error on PS5/PS6 was exactly these two `Unsorted file` entries, which is what this patchset addresses.

Ready for another CQ run whenever you get a chance.

PR #64033, opened 2026-08-12
[vm] Support iOS Simulator Mach-O snapshots

FelixMittermeier:
This adds a `--macho-ios-simulator` option to `gen_snapshot`, which marks generated Mach-O snapshots as targeting the iOS Simulator instead of a physical iOS device. It is required to unblock [flutter/flutter#191002](https://github.com/flutter/flutter/pull/191002), which adds profile and release builds for iOS Simulators.

Please let me know whether any kind of test should be added here as well and, if so, how you would recommend testing it. Thanks


mraleph:
I think this looks reasonable. Can you guide this through review @sstrickl ?

sstrickl:
I do have a suggested change, but otherwise looks good to me!

FelixMittermeier:
@mraleph @sstrickl Thanks for the review. Applied the suggested changes and updated the current branch.

PR #64041, opened 2026-08-13
[analysis_server] Fix blank-line placement in sort_constructors_first fix

### LINKED ISSUE #60703: Issues with quick fix for `sort_constructors_first` lint rule
### 1. Spacing
```dart
class Foo {
  final String foo;

  const Foo({required this.foo});
}
```
after applying quick fix:
```dart
class Foo {

  const Foo({required this.foo});
  final String foo;
}
```
expected:
```dart
class Foo {
  const Foo({required this.foo});

  final String foo;
}
```
- no blank space between signature and constructor
- space between constructor and field?

### 2. Enum:
```dart
enum Foo {
  bar('Open'),
  baz('Accepted Provisionally'),

  final String value;

  const Foo(this.value);
}
```
after applying nothing happens despite analyzer properly emits warning. Expected:
```dart
enum Foo {
  bar('Open'),
  baz('Accepted Provisionally'),

  const Foo(this.value);

  final String value;
}
```

hkarmoush:
## Summary
- The `sort_constructors_first` quick-fix moved a constructor's leading whitespace gap (including any blank line separating it from its previous sibling) verbatim to the insertion point, producing a stray leading blank line at the top of the body and no separator between the relocated constructor and the member that now follows it.
- The constructor now always starts on its own line at the insertion point with no leading blank line. If a blank line used to separate it from its previous member, that separation moves to *after* the constructor instead — unless the insertion point already has its own blank line there naturally, to avoid doubling up.

Fixes #60703

## Test plan
- [x] Added `test_noBlankLineAfterOpeningBrace`, `test_enum_noBlankLineAfterSemicolon`, and `test_enum_blankLineMovesFromOldGapToAfterConstructor` to `sort_constructor_first_test.dart`, covering the class and enum scenarios from the issue.
- [x] Updated `SortConstructorFirstBulkTest.test_single_class`'s expected output to match the corrected, consistent spacing behavior.
- [x] `dart test test/src/services/correction/fix/sort_constructor_first_test.dart` — all 17 tests pass.
- [x] Full regression run of `pkg/analysis_server/test/src/services/correction/fix/` — all 4765 tests pass, no regressions.
- [x] `dart format` / `dart analyze` clean on changed files.

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/dart/sort_constructor_first.dart
+++ b/pkg/analysis_server/lib/src/services/correction/dart/sort_constructor_first.dart
@@ -7,6 +7,7 @@ import 'package:analysis_server/src/utilities/extensions/ast.dart';
 import 'package:analysis_server/src/utilities/extensions/range_factory.dart';
 import 'package:analysis_server_plugin/edit/dart/correction_producer.dart';
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/source/line_info.dart';
 import 'package:analyzer/source/source_range.dart';
 import 'package:analyzer_plugin/utilities/change_builder/change_builder_core.dart';
 import 'package:analyzer_plugin/utilities/fixes/fixes.dart';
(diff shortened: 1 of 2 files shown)


keertip:
@hkarmoush , the changes are good, but there are test failures. Can you run the `Sort members` action (if using VS Code) and upload a new patch?

bwilkerson:
In particular, it appears that there's a merge conflict.

PR #64046, opened 2026-08-13
[analysis_server] Infer nullable type in 'Add type annotation' when assignment isn't unconditional

### LINKED ISSUE #64015: `prefer_typing_uninitialized_variables` fix `Add type annotation` incorrect type - loops/ifs/catches
Repro:

```dart
int? f() {
  var result;
  for (var i in [1, 2, 3]) {
    result = i;
  }
  return result;
}

int? g() {
  var result;
  if (1 == 1) {
    result = 0;
  }
  return result;
}

Object? h() {
  var result;
  try {
    throw '';
  } catch (e) {
    result = e;
  }
  return result;
}
```

The code above has a fix that replaces `var` with `int` (or, in the last case, `Object`), but it should be `int?` (or `Object?`) since, for flow analysis, we can't be certain the loop will ever run (or whatever better phrasing) — same logic for the condition, but even more so. Probably more cases that I couldn't think of.

CC @bwilkerson @stereotype441

hkarmoush:
## Summary

The `Add type annotation` fix/assist (triggered e.g. by `prefer_typing_uninitialized_variables`) infers an uninitialized variable's type from assignments found anywhere in the rest of the enclosing block, without regard to whether those assignments are guaranteed to run. For a variable only assigned inside a loop body, a single-branch `if`, or a `try` block, that code might not execute, so the variable could still hold its implicit initial value, `null`, at the point of use — but the fix inferred a non-nullable type anyway.

```dart
int? f() {
  var result;
  for (var i in [1, 2, 3]) {
    result = i;
  }
  return result; // fix incorrectly inferred `int`, should be `int?`
}
```

This adds a conservative check for whether at least one of the remaining statements is guaranteed to assign the variable — accounting for `if`/`else` (both branches), `try`/`finally`, `do`-`while` (body always runs once), labeled statements, and immediately-invoked closures — and makes the inferred type nullable when it isn't.

Fixes #64015

## Test plan
- [x] Added regression tests for the issue's for-each loop, single-branch `if`, and `try`/`catch` cases, plus counterexamples (`if`/`else`, `do`-`while`, `try`/`finally`) that must remain non-nullable
- [x] `pkg/analysis_server/test/src/services/correction/fix/add_type_annotation_test.dart` — 43/43 pass
- [x] `pkg/analysis_server/test/src/services/correction/assist/add_type_annotation_test.dart` — 75/75 pass
- [x] `pkg/analysis_server/test/src/services/correction/test_all.dart` (full correction suite) — 6385/6385 pass, no regressions
- [x] `dart format` / `dart analyze` clean on changed files

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/dart/add_type_annotation.dart
+++ b/pkg/analysis_server/lib/src/services/correction/dart/add_type_annotation.dart
@@ -9,8 +9,10 @@ import 'package:analysis_server_plugin/edit/dart/correction_producer.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/nullability_suffix.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/dart/element/type_system.dart';
+import 'package:analyzer/src/dart/element/type.dart';
 import 'package:analyzer/src/utilities/extensions/ast.dart';
 import 'package:analyzer_plugin/utilities/assist/assist.dart';
 import 'package:analyzer_plugin/utilities/change_builder/change_builder_core.dart';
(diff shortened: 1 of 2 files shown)


hkarmoush:
@keertip The sort-members fix (`Sort members in add_type_annotation_test.dart`) was pushed as patch set 3, so that should be resolved.

@bwilkerson Re: the `do`/`if`/`break` counterexample on `_isDefinitelyAssigned` (`add_type_annotation.dart:269`) — you're right, that's a real gap: `Block`'s `.any()` check doesn't account for a preceding statement (e.g. an `if` with a `break`/`continue`/`return`/throw and no matching `else`) exiting the block before an assigning statement is reached, so it can report a variable as definitely assigned when it isn't on all paths.

A fully sound fix means tracking abrupt completion (break/continue/return/throw) through `Block`'s sequential statements, which is more scope than this CL should take on as a heuristic quick-fix producer, so per your suggestion I've documented the limitation instead — added a doc comment on `_isDefinitelyAssigned` describing exactly this gap. Just pushed as a new patch set.

PR #64053, opened 2026-08-14
[vm/io] Reap exit-handler wakeup process

bsutton:
`ExitCodeHandler::TerminateExitCodeThread()` forks a short-lived child to
wake the exit-handler thread from `wait()`. If that thread observes
`running_ == false` before entering `wait()`, it exits without reaping the
wakeup child. In a container, that child can be adopted by a Dart process
running as PID 1 and remain as a zombie.

Retain the wakeup PID and call `waitpid()` after the exit-handler thread has
terminated. `ECHILD` is accepted because it means the handler thread already
reaped the process. Apply the equivalent change on Linux and macOS.

Add a Linux regression test that:

- makes the test process a child subreaper;
- invokes an activated executable through `dart pub global run`;
- verifies that no zombie child is adopted after the intermediate Dart VM
  exits.

Validation:

- Before the patch, the regression test reproduced the zombie on its first
  invocation using a source-built Dart 3.12.2 SDK.
- Before the patch, the container harness reproduced
  `dart:dartdev_aot <defunct>` under container PID 1.
- After the patch, the official standalone regression test passes.
- After the patch, the exact fatal-dartdoc container harness completed 30
  iterations without a zombie on Dart 3.12.2 and a source build from `main`.
- The focused standalone process suite passed 33 tests. The existing
  `process_child_test` could not create its process stream in the local
  restricted sandbox and failed independently of this change.

Fixes #61624.

---

<details>

</details>

### DIFF
--- a/runtime/bin/process_linux.cc
+++ b/runtime/bin/process_linux.cc
@@ -164,15 +164,27 @@ class ExitCodeHandler {
 
     // Wake up the [ExitCodeHandler] thread which is blocked on `wait()` (see
     // [ExitCodeHandlerEntry]).
-    if (TEMP_FAILURE_RETRY(fork()) == 0) {
+    pid_t wakeup_pid = TEMP_FAILURE_RETRY(fork());
+    if (wakeup_pid == 0) {
       // Avoid calling any atexit callbacks to prevent deadlocks.
       _exit(0);
+    }
+    if (wakeup_pid < 0) {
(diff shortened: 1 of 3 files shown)


mraleph:
There is a comment on the CL that needs addressing. If you feel like you don't want to work on this - let us know (e.g. close the PR), we will then take over landing the fix.

PR #64071, opened 2026-08-18
clamp netlink prefixlen before building netmask in getifaddrs

ubeddulla:
SetNetmask in the Android getifaddrs fallback builds a netmask from ifa_prefixlen, a u8 copied straight from the netlink message, without bounding it to the address family. A prefixlen above 128 runs the 0xFF-fill loop past the 16-byte sin6_addr and overflows the heap sockaddr_in6, and above 32 the AF_INET branch shifts by a negative amount. Clamp prefixlen to the family maximum before use, matching the length check SetAddresses already does for the address bytes in the same file.

### DIFF
--- a/runtime/bin/ifaddrs.cc
+++ b/runtime/bin/ifaddrs.cc
@@ -72,6 +72,14 @@ DART_WARN_UNUSED_RESULT static bool SetAddresses(struct ifaddrs* ifaddr,
 }
 
 static void SetNetmask(struct ifaddrs* ifaddr, int family, int prefixlen) {
+  // prefixlen comes from the netlink message (ifa_prefixlen is a u8) and is not
+  // bounded to the address family, so clamp it before using it to index into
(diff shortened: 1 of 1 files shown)


mraleph:
Don't clamp it - it is meaningless. If the length does not match - reject it, similar to how `SetAddresses` is implemented.

ubeddulla:
Makes sense, switched it to reject instead of clamp. SetNetmask now returns bool and bails on a prefixlen that doesn't fit the family (>128 for v6, >32 for v4), same as SetAddresses does for the address length, and both callers propagate the failure.

ubeddulla:
Makes sense, clamping does produce a bogus mask. Switched it to reject the out-of-range prefixlen instead, the same way SetAddresses rejects a length that doesn't match the family. Pushed.

mraleph:
There is a comment on https://dart-review.googlesource.com/c/sdk/+/538260 which needs addressing.

PR #64087, opened 2026-08-19
[io] Make SecureSocket buffer size configurable

### LINKED ISSUE #64086: SecureSocket's fixed TLS buffers limit sustained transfer throughput
`dart:io` SecureSocket uses fixed 8 KiB plaintext buffers and 10 KiB encrypted/BIO buffers. Since each SSL filter pass is dispatched through IOService, large writes are split into repeated dispatch and copy cycles. `RawSecureSocket.write` can initially accept only 8191 bytes, and there is no supported way to tune the throughput/memory tradeoff.

On Dart 3.14.0-edge (`35d2cf87af7`), Windows x64 release, a 256 MiB loopback TLS transfer measured:

| Buffer size | Round 1 | Round 2 |
|---|---:|---:|
| 8 KiB | 99.6 MiB/s | 104.3 MiB/s |
| 64 KiB | 231.7 MiB/s | 236.5 MiB/s |

This affects sustained data transfer rather than TLS handshake time.

Suggested direction: expose an opt-in per-context buffer-size setting, preserve the current default, and resize the Dart circular buffers and native BIO together.

DubaiChewyCookie:
Fixes #64086

## What this changes

This adds `SecurityContext.secureSocketBufferSize`, an opt-in per-context setting controlling the plaintext buffers allocated for newly created secure sockets.

- The default remains 8 KiB.
- Valid values range from 8 KiB through 1 MiB.
- Changing the property affects only connections created afterward.
- Existing connections retain their original buffer sizes.
- Encrypted circular buffers retain the existing 2 KiB TLS allowance.
- The native BoringSSL BIO pair uses the corresponding encrypted buffer size.

The Dart circular buffers and native BIO must be resized together. Increasing only the Dart-side buffer would otherwise move the bottleneck into the fixed 10 KiB BIO.

## Background

SecureSocket currently uses:

- 8 KiB plaintext circular buffers;
- 10 KiB encrypted circular buffers;
- a 10 KiB native BIO pair.

Large application writes are therefore divided across repeated asynchronous `IOService.sslProcessFilter` dispatches. This is particularly expensive for long-lived, high-throughput TLS connections.

The new setting allows applications to choose the throughput/memory tradeoff without changing existing defaults.

## Native lifecycle changes

The native finalizer is now registered after filter initialization. This allows its external allocation estimate to use the configured BIO size.

Native buffer pointers are initialized to null so cleanup remains safe if initialization fails before every buffer has been allocated.

## Compatibility

This is an additive API on the final `SecurityContext` class.

The default plaintext, encrypted, and BIO sizes are unchanged, so applications that do not set `secureSocketBufferSize` retain their current allocation and runtime behavior. There is no TLS protocol or wire-format change.

Larger values allocate additional memory per secure connection, which is documented on the property.

## Performance

Measured using a Windows x64 release SDK built from `35d2cf87af7`.

The loopback test transferred 256 MiB in repeated 1 MiB writes after a 16 MiB warm-up. Both endpoints used the same configured buffer size.

| Buffer size | Round 1 | Round 2 |
|---|---:|---:|
| 8 KiB | 99.6 MiB/s | 104.3 MiB/s |
| 64 KiB | 231.7 MiB/s | 236.5 MiB/s |

The 64 KiB configuration was approximately 2.3 times faster in this local measurement. The exact improvement will depend on workload and platform.

## Tests

- Built the complete Windows x64 release SDK:

  `python tools/build.py --mode release create_sdk`

- Added `standalone/io/secure_socket_buffer_size_test.dart`, covering:

  - the unchanged 8 KiB default;
  - lower and upper range validation;
  - property updates;
  - default initial `RawSecureSocket.write` capacity;
  - increased initial write capacity with a 64 KiB configuration.

- Passed the new test and relevant SecureSocket coverage, including:

  - raw secure sockets;
  - socket-to-TLS upgrades;
  - standard secure sockets;
  - secure server sockets under all short-read/write variants;
  - ALPN;
  - minimum TLS protocol version.

---

### DIFF
--- a/runtime/bin/secure_socket_filter.cc
+++ b/runtime/bin/secure_socket_filter.cc
@@ -40,10 +40,6 @@ void SSLFilter::Cleanup() {
   trust_evaluate_reply_port_ = ILLEGAL_PORT;
 }
 
-const intptr_t SSLFilter::kInternalBIOSize = 10 * KB;
-const intptr_t SSLFilter::kApproximateSize =
-    sizeof(SSLFilter) + (2 * SSLFilter::kInternalBIOSize);
-
 static SSLFilter* GetFilter(Dart_NativeArguments args) {
   SSLFilter* filter = nullptr;
   Dart_Handle dart_this = ThrowIfError(Dart_GetNativeArgument(args, 0));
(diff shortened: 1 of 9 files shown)


mraleph:
Closing this one based on discussions in https://github.com/dart-lang/sdk/issues/64115 as there is a better way to address the underlying issue.

Issue #64103, opened 2026-08-21
[vm] macOS regression in 3.13.0: every VM handle allocation now pays a thread-local lookup


LahaLuhem:
## What's wrong

Between 3.12.2 and 3.13.0, `Roots` moved from a plain static global to a `thread_local` pointer ([`e98e6a1198c`](https://github.com/dart-lang/sdk/commit/e98e6a1198ce530cebbf92ffdd3314a99df34188), [CL 493866](https://dart-review.googlesource.com/c/sdk/+/493866)). `Object::null()` reads through it, and the VM sets every internal handle to null as it allocates one. So handle allocation now does a thread-local read where it used to do a plain load from a fixed address.

On macOS that costs more than it should. There is no cheap thread-local model there, so each read is an indirect call through the variable's descriptor instead of one or two instructions. `dart::Type::Handle(Zone*)` goes from a 4-instruction leaf function to a 13-instruction one with a call in the middle.

Anything that allocates a lot of handles pays for it. I measure 5 to 6% on AOT code that ends up in the C++ type checker on every iteration.

## Expected and actual

Expected: changing where the roots live shouldn't change what it costs to allocate a handle.

Actual: 5 to 6% slower on AOT paths that call into the runtime once per operation. Paths that stay in Dart code don't move at all, which is most of them.

## Repro

I don't have a small repro for handle allocation on its own, so I used a path that makes exactly one runtime call per iteration: the record type check from #61970. Two lanes in the same file never enter the runtime, so they act as controls. Full file is attached below.

Same source compiled by both SDKs, the two binaries alternated inside every round, minimum of 5 rounds. Per-lane spread across rounds stayed inside 5%.

| lane | enters the runtime? | 3.12.2 | 3.13.1 | change |
|---|---|---|---|---|
| `concrete-pair` | no | 2.79 ns | 2.80 ns | +0.1% |
| `generic-scalar` | no | 3.22 ns | 3.21 ns | -0.5% |
| `generic-pair` | yes | 339.10 ns | 356.23 ns | **+5.1%** |
| `generic-triple` | yes | 430.54 ns | 455.36 ns | **+5.8%** |
| `generic-quad` | yes | 519.78 ns | 550.59 ns | **+5.9%** |

The two control lanes being flat is what rules out the machine drifting between runs. Splitting the record lanes by field count separates the fixed cost from the per-field cost:

- 3.12.2: 158.4 ns + 90.34 ns per field
[...]

Issue #61970, opened 2025-11-14
[VM/AOT] Using records as Map keys is slow on AOT


mkustermann:
Even though https://github.com/dart-lang/sdk/issues/60419 was closed there's still perf issues with record types in maps.
```dart
void main() {
  final strings = <String>[for (int i = 0; i < 1_000_000; ++i) '$i'];

  final sw = Stopwatch()..start();
  final m = <(int, List<String>), String>{};

  for (int i = 0; i < 1_000_000; i++) {
    m.putIfAbsent((i, strings), () => strings[i]);
  }
  print(sw.elapsed);
}
```
```
%  pkg/vm/tool/precompiler2 test.dart test.aot
% out/ReleaseX64/dartaotruntime test.aot
0:00:01.051103
```
with this profile:
```
+   13.76%  DartWorker       dartaotruntime        [.] dart::Object::HandleImpl(dart::Zone*, dart::ObjectPtr, long)
+    9.61%  DartWorker       test.aot              [.] _LinkedHashMapMixin._findValueOrInsertPoint
+    9.41%  DartWorker       dartaotruntime        [.] dart::Class::IsSubtypeOf(dart::Class const&, dart::TypeArguments const&, dart::Nullability, dart::AbstractType const&, dart::Heap::Space, dart::FunctionTypeMapping
+    7.90%  DartWorker       dartaotruntime        [.] dart::VMHandles::AllocateHandle(dart::Zone*)
+    4.25%  DartWorker       dartaotruntime        [.] dart::Instance::RuntimeTypeIsSubtypeOf(dart::AbstractType const&, dart::TypeArguments const&, dart::TypeArguments const&) const
+    3.45%  DartWorker       test.aot              [.] _Record.==
+    3.40%  DartWorker       dartaotruntime        [.] dart::Type::type_class() const
+    3.21%  DartWorker       dartaotruntime        [.] dart::AbstractType::operator=(dart::AbstractTypePtr)
+    3.00%  DartWorker       dartaotruntime        [.] dart::Type::type_class_id() const
+    2.52%  DartWorker       dartaotruntime        [.] dart::MarkingVisitor::VisitPointers(dart::ObjectPtr*, dart::ObjectPtr*)
+    2.22%  DartWorker       test.aot              [.] _Record.hashCode
+    2.09%  DartWorker       dartaotruntime        [.] dart::TypeArguments::IsUninstantiatedIdentity() const
[...]

PR #64111, opened 2026-08-22
[vm] Cache the null instance on the Zone to avoid a TLS lookup per handle

LahaLuhem:
`Roots` became a `thread_local` in `e98e6a1198c`, so `Object::null()` costs a thread-local read and every VM handle is initialised to null. Paths allocating many handles pay it per handle, and that read is a call on Darwin and in ELF shared libraries.

`Handle(Zone*)` is the largest group of readers, needs only the null, has no `Thread` in hand, and is `DART_NOINLINE`, so there is nothing to share a lookup with. This caches the null on the `Zone`, filled on first use and cleared in `Zone::Reset()`.

On arm64 macOS, against a path entering the runtime type checker once per iteration (#61970's record subtype check): 1.4% to 1.8% faster at three and four record fields, `_tlv_get_addr` down from 5.99% to 4.07% of samples, controls that stay out of the runtime unchanged, snapshot output byte identical.

Not GC-visited: safe only because the null instance never moves, which `Thread::object_null_` already assumes. Slightly unsure, since if that guarantee is weaker than I think then the existing field is wrong too.

Measured and discussed in more detail in #64103, including the cross-platform numbers and two approaches that did not work.

Ran locally on arm64 macOS against the parent commit: `vm` 4166 pass / 14 fail against 4165 / 15, `language` + `corelib` 7120 / 10 on both, no new failure names in either. The one difference is `vm/dart/thread_priority_macos_test`, which fails on the parent under the parallel suite and passes in isolation.

<details>

</details>

### DIFF
--- a/runtime/vm/object.h
+++ b/runtime/vm/object.h
@@ -34,6 +34,7 @@
 #include "vm/json_stream.h"
 #include "vm/os.h"
 #include "vm/raw_object.h"
+#include "vm/zone.h"
 #include "vm/regexp/regexp-flags.h"
 #include "vm/report.h"
 #include "vm/roots.h"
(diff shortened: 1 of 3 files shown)


LahaLuhem:
Closed due to newer better direction/approach

PR #64132, opened 2026-08-25
[tools] Fix Result equality operator

amanmaurya92:
Fixes an issue in `tools/diff_results.dart` where the `Result` class overrode `hashCode` but failed to implement `operator ==`, which violates Dart's equality contract. 

Also removes an unused, buggy `equals()` method that always returned `false`.

---

<details>

</details>

### DIFF
--- a/tools/diff_results.dart
+++ b/tools/diff_results.dart
@@ -302,21 +302,14 @@ class Result {
         result == other.result;
   }
 
-  bool equals(Object other) {
-    if (other is Result) {
-      if (name != other.name) return false;
-      if (builderName != other.builderName) return false;
-    }
-    return false;
-  }
-
   @override
   int get hashCode => name.hashCode ^ builderName.hashCode;
 
   @override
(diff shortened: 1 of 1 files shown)


mraleph:
You can just delete both `hashCode` and `operator==` overloads. Not that anybody really cares about this dead code to begin with. 

Also note that overriding `hashCode` without overriding `operator ==` is actually fine at least in this particular case because we are talking about subclass of `Object`, because the only invariant you really have to care about is that `a == b` implies `a.hashCode == b.hashCode` - and this invariant is still maintained.

amanmaurya92:
I was going through the codes line by line while getting familiar with the project, and I happened to come across these older TODOs/code paths.

I have one more PR around some old code, but these two should be the only ones I pursue from the older parts of the codebase. I don't want to add review overhead if these aren't particularly useful anymore. my bad..

Going forward, I'll focus more on newer or higher-priority issues where I can hopefully make more meaningful contributions.

Issue #64115, opened 2026-08-22
[dart:io] Design discussion: should SecureSocket's SSLFilter staging path be replaced?


DubaiChewyCookie:
## Context

This is a design follow-up to [#64086](https://github.com/dart-lang/sdk/issues/64086) and [#64087](https://github.com/dart-lang/sdk/pull/64087). It also brings together several older reports whose throughput, CPU, latency, backpressure, and lifetime symptoms overlap at the same `dart:io` socket/TLS boundary.

I am opening this before attempting a larger patch. `SecureSocket` is used by Flutter applications, HTTP clients, and Dart services, and a change to socket ownership, TLS scheduling, or backpressure can have a wide effect. I would appreciate guidance on three related questions:

1. whether the current path should receive only a short-term mitigation;
2. whether its fixed-buffer `SSLFilter` pipeline should eventually be replaced; and
3. if it should be replaced, which internal transport boundary and rollout plan are appropriate.

My current preference, subject to maintainer feedback, is to preserve the existing public `SecureSocket` API while evaluating an internal TLS-aware VM socket backend behind a temporary runtime flag. This issue does not propose a public `NativeSecureSocket` type, and it does not couple the backend work to deprecating `RawSecureSocket`.

## The current path stages each direction several times

The current separation keeps TLS processing apart from socket transport. Its practical result is that BoringSSL does not exchange data directly with the VM socket layer. A separate `SSLFilter`, invoked through `IOService`, moves data between BoringSSL and `RawSocket` through several fixed circular buffers.

Those circular buffers still serve a useful purpose by amortizing allocation and scheduling costs for small I/O. The concern is that the same fixed-capacity staging path remains mandatory when more data is already available.

The explicit transport-staging chain contains six fixed-capacity buffers per `SecureSocket`:

- two plaintext external buffers;
- two encrypted external buffers; and
- two internal buffers created by `BIO_new_bio_pair`.

[...]

Issue #61690, opened 2025-10-06
TSAN: data race in secure sockets


rmacnak-google:
```
DART_CONFIGURATION=ReleaseTSANARM64 TEST_COMPILATION_DIR=/Volumes/Work/s/w/ir/xcodebuild/ReleaseTSANARM64/generated_compilations/vm-aot-tsan-mac-release-arm64/tests_standalone_io_https_client_certificate_test xcodebuild/ReleaseTSANARM64/dartaotruntime -Dtest_runner.configuration=vm-aot-tsan-mac-release-arm64 --ignore-unrecognized-flags --packages=/Volumes/Work/s/w/ir/.dart_tool/package_config.json /Volumes/Work/s/w/ir/xcodebuild/ReleaseTSANARM64/generated_compilations/vm-aot-tsan-mac-release-arm64/tests_standalone_io_https_client_certificate_test/out.aotsnapshot

exit code:
-6

stdout:
unittest-suite-wait-for-done

stderr:
==================
WARNING: ThreadSanitizer: data race (pid=24703)
  Write of size 4 at 0x000109a0f0cc by thread T7:
    #0 ssl_read_impl(ssl_st*) <null> (dartaotruntime:arm64+0x1001ece84)
    #1 SSL_read <null> (dartaotruntime:arm64+0x1001ecd60)
    #2 dart::bin::SSLFilter::ProcessAllBuffers(int*, int*, bool) <null> (dartaotruntime:arm64+0x100020224)
    #3 dart::bin::SSLFilter::ProcessFilterRequest(dart::bin::CObjectArray const&) <null> (dartaotruntime:arm64+0x10001fbd4)
    #4 dart::bin::IOServiceCallback(long long, _Dart_CObject*) <null> (dartaotruntime:arm64+0x100017ab4)
    #5 dart::(anonymous namespace)::HandleMessage::Run() <null> (dartaotruntime:arm64+0x1003650f4)
    #6 dart::ThreadPool::WorkerLoop(dart::ThreadPool::Worker*) <null> (dartaotruntime:arm64+0x10047ab38)
    #7 dart::ThreadPool::Worker::Main(unsigned long) <null> (dartaotruntime:arm64+0x10047af5c)
    #8 dart::ThreadStart(void*) <null> (dartaotruntime:arm64+0x100410b78)

  Previous write of size 4 at 0x000109a0f0cc by thread T6:
[...]

Issue #59937, opened 2025-01-18
Memory leak when spawning isolates that use HTTP


davidmartos96:
Dart version: `Dart SDK version: 3.6.0 (stable) (Thu Dec 5 07:46:24 2024 -0800) on "linux_x64"`

On our production shelf server, deployed on Cloud Run, we've discovered a memory leak. The leak happens when using `Isolate.run` and inside the isolate callback we fetch data from an external endpoint using the `http` package.

Here is a reproduction repository: https://github.com/davidmartos96/dart-isolate-http-leak

The RSS memory will rapidly grow with the `cli.dart` and `stress_server.dart` scripts provided. The scripts also provide a way to disable the isolate to see the difference.

# Cloud Run visualization

![Image](https://github.com/user-attachments/assets/128a6023-c8f6-4ba3-8e34-7d0331e69a1a)

mraleph:
Are you sure it is a memory leak and not a higher peak memory usage? Each time you invoke `Isolate.run` you need to create a fresh isolate - fresh isolate means fresh static state, etc. If you run many of these concurrently it takes time for GC to reclaim all that memory. It's only a leak if RSS does not drop back eventually.

davidmartos96:
@mraleph Thank you for the response.

We've been trying to debug a leak on our Dart server the last weeks, where memory was starting to grow slowly but steadily. We've used a 1GB RAM configuration, and it never stopped growing until reaching the max memory and Cloud Run starting a new instance because of OOM.

We will still need to wait a couple of days to see if a recent deploy without `Isolate.run` fully stops the leak. Otherwise it could be something else. The server serves requests every 1.5 s, so it's not a big server where we would expect 1 GB memory.

One thing we've been wondering while debugging this the past weeks, is the implications of setting the VM flag `--old_gen_heap_size=500` in the 1 GB Cloud Run machine. Would that, assuming a "non leaking" long running program like a server, make the Garbage collector try to stay below that memory, and possibly do more aggressive collects if needed?

davidmartos96:
The endpoint on our server where we use Isolate.run is only called on some occasions, the reproduction stress script is not a representation of how our server was processing these requests, but it shows the memory growth without garbage collecting. On my local machine it kept growing past 3 GB, even when constraining the old heap size flag.

davidmartos96:
[...]

PR #64160, opened 2026-08-28
[io] Use custom socket BIO for SecureSocket

DubaiChewyCookie:
Related to #64115

It replaces the previous [two plaintext and two ciphertext fixed-size rings](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/_internal/vm/bin/secure_socket_patch.dart?plain=1#L65-L80) and [buffered BIO pair](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.cc?plain=1#L501-L513) with [two growable plaintext rings](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L1473-L1578) and a [non-buffering custom socket BIO](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L450-L638). The rings have 1 KiB, 4 KiB, or 16 KiB of usable capacity, each backing includes one sentinel byte, and ciphertext no longer passes through Dart-visible staging rings.

### Dart I/O layer

- Previously, each filter turn [exposed a native pointer to Dart](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/_internal/vm/bin/secure_socket_patch.dart?plain=1#L193-L195). The native getter [retained the `SSLFilter` for the IO-service thread, which released it only when `ProcessFilterRequest` ran](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.cc?plain=1#L208-L241), while Dart [awaited IO-service request 43](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L1215-L1240). An isolate exiting with filter work pending could therefore leave the native filter retained. Filter processing now [runs synchronously through a native entry point](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L326-L373), removing that cross-thread ownership and leak path. Dart [serializes handshake attempts](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L1146-L1208) and [filter turns](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L1266-L1335), [treats callback reentry as backpressure](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L958-L1027), schedules filter errors and readiness asynchronously, and [defers native destruction until both are idle](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L885-L900).

- Previously, TLS upgrade [queried already-consumed read-close state through a hard `_RawSocketBase` cast](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L761-L762) and [dispatched that state before replacing subscription callbacks or connecting the filter](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L684-L720). With a supplied subscription, this rejected public delegating sockets at the cast. The custom BIO supports them through callbacks, but treating a missing `_RawSocketBase` as open would lose a consumed `readClosed` event and leave the handshake waiting indefinitely. The final implementation therefore [unwraps delegating sockets when inspecting read-close state](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L812-L834), [captures that state before replacing callbacks, and applies it only after the filter is connected](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L720-L770), so a consumed close produces `HandshakeException` instead of a cast failure or connection hang.

- The previous encrypted-ring transport could [compare each `RawSocket.write` result with the queued ciphertext and enable write events after a short write](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L1406-L1415). Direct BIO writes remove that Dart-side comparison, and a positive short write is not guaranteed to produce another socket event. Handshake processing now [uses the byte progress and pending-write state returned by native code to retry until the BIO actually blocks](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L1146-L1195), while connected filter turns [treat native transport bytes as progress](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L1419-L1446), preventing connections from hanging while waiting for an event that may never arrive.

### Native TLS and socket layer

- Previously, BoringSSL [exchanged ciphertext through a BIO pair](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.cc?plain=1#L769-L817), while Dart [copied that ciphertext between encrypted rings and `RawSocket`](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L1136-L1175). The custom BIO now [reads and writes a retained native socket directly or invokes Dart transport callbacks](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L450-L615) for public `RawSocket` implementations, propagating callback byte counts outside the requested range as catchable `ArgumentError`s. A nested `RawSecureSocket` [deliberately remains a callback transport](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/io/secure_socket.dart?plain=1#L836-L849), so an outer TLS layer cannot bypass the inner one. Because native transport bypasses Dart `RawSocket.read` and `write`, the VM now [mirrors availability, write readiness, and profiling counters explicitly](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/sdk/lib/_internal/vm/bin/socket_patch.dart?plain=1#L2506-L2537).

- Previously, TLS-upgrade `bufferedData` was [sliced in Dart and copied into the encrypted input ring](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L1136-L1165). It is now copied into native memory in [chunks of at most 16 KiB](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.h?plain=1#L71-L74); if a Dart-owned tail remains, [the BIO sets its read-retry flag before reading newer socket bytes](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L411-L480), preserving transport order without an unbounded native copy.

- The BIO pair previously kept nonblocking and close handling outside BoringSSL in Dart's [`RawSocket` staging reads and writes](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/sdk/lib/io/secure_socket.dart?plain=1#L1158-L1174). Since the custom BIO calls `SocketBase` directly, it now [maps nonblocking, reset, and close-related socket results to BIO retry state](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L25-L71), allowing `RawSocket` close events to terminate the connection instead of immediately turning those results into `SSL_ERROR_SYSCALL`.

- The previous native filter [owned a BIO-pair endpoint but no native `Socket` or fd](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.h?plain=1#L121-L137), so descriptor lifetime belonged to the Dart transport. The custom BIO caches the descriptor across TLS calls. The filter [retains the native `Socket`](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L896-L947), also [retains its heap-backed handle on Windows and Fuchsia](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/socket.h?plain=1#L61-L74), and [releases both after `SSL_free` frees the BIO](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L1086-L1098), preventing use-after-free when the original Dart socket is collected.

- The previous `SSLFilter` constructor [initialized several fields but not its backing arrays](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.h?plain=1#L59-L68), leaving [`buffers_` and `dart_buffer_objects_` without initial values](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.h?plain=1#L132-L137). `InitializeBuffers` could [return an error before initializing either array](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.cc?plain=1#L360-L406), after which the failure path [called `Destroy`](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.cc?plain=1#L80-L94), whose cleanup [traversed both ownership arrays](https://github.com/DubaiChewyCookie/sdk/blob/0073832ca7fc20d237c61c8fe2b599cae02f1979/runtime/bin/secure_socket_filter.cc?plain=1#L667-L715); this could access invalid ownership state or attempt invalid cleanup during partial initialization. The new constructor [null-initializes every backing slot](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.h?plain=1#L79-L103), and each backing is [reference-counted and wrapped with an `ExternalUint8List` finalizer](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L100-L134). Initialization [publishes a backing only after its wrapper succeeds](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L762-L832), and growth [releases the filter's old generation only after the replacement wrapper exists](https://github.com/DubaiChewyCookie/sdk/blob/600537dd1e14f4570fa03ab7d681841ba1ef350e/runtime/bin/secure_socket_filter.cc?plain=1#L703-L749), closing the cleanup hole without invalidating an old Dart view.

### DIFF
--- a/runtime/bin/io_impl_sources.gni
+++ b/runtime/bin/io_impl_sources.gni
@@ -98,4 +98,7 @@ io_impl_sources = [
   "typed_data_utils.h",
 ]
 
-io_impl_tests = [ "secure_socket_utils_test.cc" ]
+io_impl_tests = [
+  "secure_socket_filter_test.cc",
+  "secure_socket_utils_test.cc",
+]
--- a/runtime/bin/io_natives.cc
+++ b/runtime/bin/io_natives.cc
@@ -118,15 +118,17 @@ namespace bin {
   V(ProcessInfo_CurrentRSS, 0)                                                 \
   V(ProcessInfo_MaxRSS, 0)                                                     \
   V(RawSocketOption_GetOptionValue, 1)                                         \
-  V(SecureSocket_Connect, 7)                                                   \
+  V(SecureSocket_Connect, 10)                                                  \
   V(SecureSocket_Destroy, 1)                                                   \
-  V(SecureSocket_FilterPointer, 1)                                             \
   V(SecureSocket_GetSelectedProtocol, 1)                                       \
+  V(SecureSocket_GrowBuffer, 6)                                                \
   V(SecureSocket_Handshake, 2)                                                 \
   V(SecureSocket_MarkAsTrusted, 3)                                             \
   V(SecureSocket_NewX509CertificateWrapper, 1)                                 \
   V(SecureSocket_Init, 1)                                                      \
   V(SecureSocket_PeerCertificate, 1)                                           \
+  V(SecureSocket_Process, 2)                                                   \
+  V(SecureSocket_QueuePrefetchedData, 3)                                       \
   V(SecureSocket_RegisterBadCertificateCallback, 2)                            \
   V(SecureSocket_RegisterKeyLogPort, 2)                                        \
   V(SecureSocket_RegisterHandshakeCompleteCallback, 2)                         \
--- a/runtime/bin/io_service.cc
+++ b/runtime/bin/io_service.cc
@@ -10,8 +10,6 @@
 #include "bin/directory.h"
 #include "bin/file.h"
 #include "bin/io_buffer.h"
-#include "bin/secure_socket_filter.h"
-#include "bin/security_context.h"
 #include "bin/socket.h"
 #include "bin/utils.h"
 
--- a/runtime/bin/io_service.h
+++ b/runtime/bin/io_service.h
@@ -61,8 +61,7 @@ namespace bin {
   V(Directory, ListStart, 39)                                                  \
   V(Directory, ListNext, 40)                                                   \
   V(Directory, ListStop, 41)                                                   \
-  V(Directory, Rename, 42)                                                     \
-  V(SSLFilter, ProcessFilter, 43)
+  V(Directory, Rename, 42)
 
 #define DECLARE_REQUEST(type, method, id) k##type##method##Request = id,
 
--- a/runtime/bin/io_service_no_ssl.h
+++ b/runtime/bin/io_service_no_ssl.h
@@ -18,8 +18,8 @@ namespace dart {
 namespace bin {
 
 // This list must be kept in sync with the list in sdk/lib/io/io_service.dart
-// In this modified version, though, the request 43 for SSLFilter::ProcessFilter
-// is removed, for use in contexts in which secure sockets are not enabled.
+// Secure sockets use a synchronous custom BIO and do not dispatch filter work
+// through the IO service.
 #define IO_SERVICE_REQUEST_LIST(V)                                             \
   V(File, Exists, 0)                                                           \
   V(File, Create, 1)                                                           \
--- a/runtime/bin/secure_socket_filter.cc
+++ b/runtime/bin/secure_socket_filter.cc
@@ -6,6 +6,8 @@
 
 #include "bin/secure_socket_filter.h"
 
+#include <errno.h>
+
 #include <openssl/bio.h>
 #include <openssl/ssl.h>
 #include <openssl/x509.h>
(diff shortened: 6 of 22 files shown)


DubaiChewyCookie:
In addition to dart-lang/sdk#64115, this change is also related to dart-lang/sdk#64086, dart-lang/sdk#59937, dart-lang/sdk#61690, dart-lang/sdk#52500, dart-lang/sdk#54009, dart-lang/sdk#54752, dart-lang/http#909, dart-lang/sdk#48210, dart-lang/http#1520, localsend/localsend#384, cfug/dio#2100, flutter/flutter#142042, and flutter/flutter#49392.

mraleph:
If you would like to upload a revision of the CL which addresses comments from https://github.com/dart-lang/sdk/issues/64115 - please let us know. 

Otherwise, if you feel that you don't want to work on this anymore please close the PR.

DubaiChewyCookie:
@mraleph Soon after your last comment, I started working on it. I’ve implemented several different approaches and have been comparing their benchmark results. I’m finding the investigation and benchmarking process very interesting and enjoyable. Ultimately, I think I’ll be able to submit an implementation that removes the direct socket I/O path while following the Node.js-style approach you’re looking for. However, due to some personal circumstances, I’m currently unable to devote time to this work, so it may take a while before I can complete and submit it. Would you mind giving me a little more time to finish and submit the implementation?

mraleph:
@DubaiChewyCookie no worries! I am happy that you are still working on this.

PR #64174, opened 2026-08-31
Range check kernel binary trailing-index reads

shaggyinsomniac:
The kernel loader reads the string-table count, the string-data end offset, and the library offsets from the trailing index of a `.dill` file and feeds them unchecked into `TypedData::New` and `Reader::ViewFromTo`. The bounds that exist inside those helpers are `ASSERT`s, which are stripped in release/product builds — so a malformed kernel binary produces out-of-bounds `TypedDataView`s that are then read past the buffer during string decoding, and unbounded allocations from a corrupted library count.

This adds range checks against `Reader::size()` at the three read sites, in the same style as the existing `library_count` range check (#63366) and the size check at `kernel_loader.cc:352` (`FATAL("Invalid kernel binary: …")`).

This fixes memory-safety issues reported via a private security advisory (out-of-bounds reads / SEGV in `dart run`, `dart compile`, and `gen_snapshot` on crafted kernel binaries). I have single-field-corruption mutants of a real `.dill` plus a generator script ready to contribute as regression tests, and am happy to rework these into unit tests or coordinate landing however the team prefers given the advisory.

### DIFF
--- a/runtime/vm/kernel_loader.cc
+++ b/runtime/vm/kernel_loader.cc
@@ -390,6 +390,9 @@ void KernelLoader::InitializeFields(UriToSourceTable* uri_to_source_table) {
   Reader reader(binary);
   reader.set_offset(program_->string_table_offset());
   intptr_t count = reader.ReadUInt() + 1;
+  if (count < 0 || count > reader.size()) {
+    FATAL("Invalid kernel binary: string table count out of range");
+  }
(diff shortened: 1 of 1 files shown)


mraleph:
It's an intermediate binary format with no integrity guarantees, so this change is meaningless. 

I know LLMs think these checks are somehow useful (it's the third AI slop PR we get with the same content), but they are not.

Issue #63444, opened 2026-05-25
Suggestion: `Create constructor` fix, `Create class` false negative and missing `Create getter`


FMorschel:
Repro:

```dart
class C() {}

void f() {
  C.other();
}
```

Especially now that we have the binding assist, we should be able to create something really similar to a static method and then bind the values later.

<img width="462" height="262" alt="Image" src="https://github.com/user-attachments/assets/12af2522-5c32-4726-8eb0-57f393c415b1" />

CC @bwilkerson @srawlins

bwilkerson:
The ability to define either a constructor or a getter is something we should consider adding at some point, but it isn't urgent.

The ability to define a class is a bug that we should correct by not offering it as a fix in this case.

PR #64193, opened 2026-09-03
[analysis_server] Do not offer CreateClass for method invocations on non-prefix targets

### LINKED ISSUE #64192: Invalid fix offered for undefined member
Given the following file:
```dart
void f() {
  C.m();
}

class C {}
```

The server is currently offering to create a class named `m` to fix the diagnostic on line 2. It shouldn't be offered because it won't fix the problem.

ayushgupta9906:
### Summary
Do not offer the `Create class` quick fix on method invocations where the target is not an import prefix.

In Dart, an invocation of the form `target.name()` can only be a class instantiation if `target` is an import prefix (`PrefixElement`). Previously, when calling an undefined member on a class (e.g. `C.m()`), `target.staticType` evaluated to `null`, causing `CreateClass` to erroneously offer to create a top-level `class m`.

Fixes https://github.com/dart-lang/sdk/issues/64192
Work towards https://github.com/dart-lang/sdk/issues/63444

### Changes
- Updated `CreateClass.producers` in `pkg/analysis_server/lib/src/services/correction/dart/create_class.dart` to verify that when `targetNode == methodName`, `target` must either be `null` (unqualified invocation) or a `PrefixElement`. If `target != null` and not an import prefix, creating a class is rejected.
- Added unit tests in `pkg/analysis_server/test/src/services/correction/fix/create_class_test.dart` covering lowercase and uppercase static method invocations, as well as preserving support for undefined target invocations (e.g. `Foo.bar()`).

### DIFF
--- a/pkg/analysis_server/lib/src/services/correction/dart/create_class.dart
+++ b/pkg/analysis_server/lib/src/services/correction/dart/create_class.dart
@@ -50,11 +50,13 @@ class CreateClass extends MultiCorrectionProducer {
       requiresConstConstructor |= _requiresConstConstructor(targetNode);
     } else if (targetNode case SimpleIdentifier(:var parent)
         when parent is! PropertyAccess && parent is! PrefixedIdentifier) {
-      if (parent case MethodInvocation(:var target)) {
-        if (target case SimpleIdentifier(:PrefixElement element)) {
-          prefixElement = element;
-        } else if (target?.staticType != null) {
-          return const [];
+      if (parent case MethodInvocation(:var target, :var methodName)) {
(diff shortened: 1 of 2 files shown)


keertip:
@ayushgupta9906 , there is a comment on the CL 

Unsorted file /b/s/w/ir/cache/builder/sdk/pkg/analysis_server/test/src/services/correction/fix/create_class_test.dart

Sorry we alphabetically sort members. Please fix and upload.

mraleph:
@ayushgupta9906 ping, or we close PR as stale.

Issue #64157, opened 2026-08-27
standalone/io/client_socket_add_close_no_error_test flaky -> Fail on vm-ubsan-fuchsia-release-x64


alexmarkov:
The tests
```
standalone/io/client_socket_add_close_no_error_test Fail (expected Pass)
```
are failing on configurations
```
vm-ubsan-fuchsia-release-x64
```

```
--- Command "vm" (took 02.000598s):
DART_CONFIGURATION=ReleaseFuchsiaUBSANX64 FFX_ISOLATE_DIR=/b/s/w/ir/x/t/ZQZVQA QEMU_LD_PREFIX=/usr/x86_64-linux-gnu ./build/fuchsia/with_envs.py ./third_party/fuchsia/test_scripts/test/run_executable_test.py --test-name=fuchsia-pkg://fuchsia.com/dart_test#meta/dartvm_test_component.cm --test-realm=/core/testing/system-tests --out-dir=out/ReleaseFuchsiaUBSANX64/gen/dart_test --package-deps=dart_test.far -Dtest_runner.configuration=vm-ubsan-fuchsia-release-x64 --ignore-unrecognized-flags --packages=.dart_tool/package_config.json --disable-dart-dev pkg/data/tests/standalone/io/client_socket_add_close_no_error_test.dart

exit code:
1

stdout:
Running test 'fuchsia-pkg://fuchsia.com/dart_test#meta/dartvm_test_component.cm'
[RUNNING]	main
[stderr - main]
Unhandled exception:
Expect.equals(expected: <1048576>, actual: <460609>) fails.
#0      Expect._fail (package:expect/expect.dart:992:5)
#1      Expect._failNotEqual (package:expect/expect.dart:184:5)
#2      Expect.equals (package:expect/expect.dart:161:5)
#3      clientSocketAddCloseNoErrorTest.<anonymous closure>.<anonymous closure>.<anonymous closure> (file:///b/s/w/ir/cache/builder/sdk/tests/standalone/io/client_socket_add_close_no_error_test.dart:26:18)
#4      Zone.runGuarded (dart:async/zone.dart:474:14)
#5      _BufferingStreamSubscription._sendDone.sendDone (dart:async/stream_impl.dart:434:13)
#6      _BufferingStreamSubscription._sendDone (dart:async/stream_impl.dart:444:7)
#7      _BufferingStreamSubscription._close (dart:async/stream_impl.dart:333:7)
[...]

PR #64194, opened 2026-09-03
[gardening] Skip client_socket_add_close_no_error_test on Fuchsia

ayushgupta9906:
### Summary
Skips `standalone/io/client_socket_add_close_no_error_test` on Fuchsia where it is flaky and failing under QEMU UBSan configurations.

Bug: https://github.com/dart-lang/sdk/issues/64157


mraleph:
Marking it flakey does not do anything useful. If you want to help you should debug to understand why it is failing.

PR #64247, opened 2026-09-09
[dart2wasm] Fix multi-memory operands and optimizer support

nmfisher:
[CL 470020](https://dart-review.googlesource.com/c/sdk/+/470020) introduced the `Memory` class for memory declarations and interoperability with modules that use linear memory. 

In the [WebAssembly binary specification](https://webassembly.github.io/spec/core/binary/instructions.html#binary-memarg), the order of operands for a memory instruction when an explicit memory index is present is `n:u32 x:memidx m:u64`  (i.e. alignment/flags, then memory index, then offset). In the current implementation, `MemoryOffsetAlign` sets bit 6 for nonzero memory indices but then incorrectly swaps the order of memory index and offset:

```
  @override
  void serialize(Serializer s) {
    if (memory.index == 0) {
      s.writeByte(align);
      s.writeUnsigned(offset);
    } else {
      s.writeByte(64 + align);
      s.writeUnsigned(offset); <-- index should be here
      s.writeUnsigned(memory.index); <-- offset should be here
    }
  }

  static MemoryOffsetAlign deserialize(Deserializer d, Memories memories) {
    final alignAndMemory = d.readByte();
    final align = alignAndMemory & 0x3F; <-- offset should be read here
    final offset = d.readUnsigned(); <-- alignment should be read here
    final memoryIndex = (alignAndMemory & 0x40) != 0 ? d.readUnsigned() : 0;
    return MemoryOffsetAlign(
      memories[memoryIndex],
      offset: offset,
      align: align,
    );
  }
```

This PR corrects this (preserving the existing encoding for memory index zero) and adds some related regression tests.

This PR also enables Binaryen’s `--enable-multimemory` flag in both the standard and multi-module optimization configurations (without this, the two-memory regression programs fail validation during optimization).

The use case is accessing independent native Wasm libraries’ memories from Dart. The changes preserve the existing experimental API and its static memory-selection model.

To run tests:
```
# Build SDK and Wasm test dependencies.
python3 third_party/emsdk/emsdk.py install 3.1.3
python3 third_party/emsdk/emsdk.py activate 3.1.3
python3 tools/build.py -m release -a x64 dart2wasm_bot

# Unit-test wasm_builder's binary encoder/decoder (checks Wasm bytes)
python3 tools/test.py -a x64 \
    -n wasm-unittest-asserts-linux \
    pkg/dart2wasm/test/wasm_memory_encoding_test

# Run the new multi-memory tests and existing memory tests in V8,
# both without optimization (O0) and with Binaryen optimization (O1).
python3 tools/test.py -a x64 \
    -n dart2wasm-linux-d8,dart2wasm-linux-optimized-d8 \
    'web/wasm/(deferred_)?(multi_)?memory_test'
```

- The encoding regression fails before the operand fix and passes afterward.
- Optimized multi-memory compilation fails before the Binaryen flag fix and passes afterward.
- Multi-memory execution passes at O0 and O2, including an O2 deferred-loading case.
- Existing memory and deferred-memory tests pass at O2.
- Validation used the patched compiler directly from source; the full SDK test suite and CQ have not been run.

---

<details>

</details>


nmfisher:
Tagging @simolus3 who contributed the original CL - would appreciate any comments/input.

tests/web/wasm/multi_memory_corruption_test.dart
```diff
@@ -0,0 +1,27 @@
...
+import 'package:expect/expect.dart';
+
+@pragma('wasm:memory-type', MemoryType(limits: Limits(1)))
+external Memory get first;
+
+@pragma('wasm:memory-type', MemoryType(limits: Limits(1)))
+external Memory get second;
+
+void main() {
+  first.storeInt32(0, WasmI32.fromInt(0x01020304));
```
simolus3, line 18: Isn't this also covered by `multi_memory_test.dart`?

pkg/dart2wasm/test/wasm_memory_encoding_test.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:typed_data';
+
+import 'package:expect/expect.dart';
+import 'package:wasm_builder/wasm_builder.dart';
+
+void main() {
```
simolus3: From what I understand, how `dart2wasm` generates wasm is typically tested with ir tests in `test/ir_test` (you can update them by running `./tools/sdks/dart-sdk/bin/dart pkg/dart2wasm/test/ir_test.dart -w` in the SDK root).

Obviously it's not great when both the encoder and decoder are wrong in the same way which gives the illusion of stuff working, but these tests go through `wasm-opt` so we have an external tool also validating this. It might be possible to replace this test by adding a second memory to `test/ir_tests/memory_use.dart` and then accessing it with offsets? The compilation should fail in `wasm-opt` without your fix.

simolus3 (commented):
Yikes! I did not run into this because I've never used modules with multiple memories, but looking at the wasm spec this fix is clearly correct.

tests/web/wasm/multi_memory_corruption_test.dart
nmfisher: The idea was to compile two modules, one which will pass module validation but read/write to the wrong location (`multi_memory_corruption_test.dart`) and another which will fail validation because it references a non-existent memory (`multi_memory_test.dart`). But yes, the store/load in the former are duplicated in the latter, so I could remove from `multi_memory_test.dart`.

nmfisher: Removed the duplicate store/load from `multi_memory_test.dart`

pkg/dart2wasm/test/wasm_memory_encoding_test.dart
nmfisher: I'm not overly familiar with the SDK test setup so that makes more sense. I removed the test in favour of adding a second memory to `pkg/dart2wasm/test/ir_tests/memory_use.dart` with a store/load.  I regenerated & committed the expected `memory_use.wat` (I also rebased onto the latest `main` branch which had touched this too).

PR #64330, opened 2026-09-20
[analysis_server] Parenthesize low-precedence targets in 'Replace cascade with dot'

### LINKED ISSUE #63971: `avoid_single_cascade_in_expression_statements` fix handling an `await`ed `Future` target
Repro:

```dart
class C {
  List<int> list = [];
}

Future<C> getC() => Future.syncValue(C());

void f() async {
  await getC()
    ..list.forEach((e) => print(e));
// avoid_single_cascade_in_expression_statements
}
```

Trying to "use the quick-fix" (not available for this case yet):

```dart
class C {
  List<int> list = [];
}

Future<C> getC() => Future.syncValue(C());

void f() async {
  await getC()
    .list.forEach((e) => print(e));
//   ^^^^ undefined_getter (The getter 'list' isn't defined for the type 'Future<C>'.)
}
```

I think we could enable the fix and wrap the target in parentheses.

CC @bwilkerson

hkarmoush:
## Summary
- The fix for `avoid_single_cascade_in_expression_statements` replaced `..` with `.` without looking at the cascade target. When the target binds looser than a selector, the result parses differently: `await getC()..list.forEach(...)` became `await getC().list.forEach(...)`, which applies the `await` to the whole selector chain and reports `undefined_getter` on `Future<C>`.
- The target is now wrapped in parentheses when its precedence is lower than postfix (`await`, prefix operators, `as`, binary operators, etc.), so the fix produces `(await getC()).list.forEach(...)`. This uses the same `precedence < Precedence.postfix` check as `AddExplicitCast` and `AddNullCheck`.

Fixes #63971

## Test plan
- [x] Added `test_method_awaitTarget_normalCascade` and `test_method_awaitTarget_nullAwareCascade` to `replace_cascade_with_dot_test.dart`.
- [x] `dart format` / `dart analyze` clean on the changed files.
- [ ] The fix test suite has not been run locally (no SDK build available here), so the new tests are unverified until CI runs.


bwilkerson:
One of the bots has failed, preventing this from landing. You can see the failure information on Gerrit. Let us know if you need help finding it.

